PR #1, opened 2016-02-26
Updated links to in tree files

IanLee1521:
@cyrush -- Don't know if you know, but you can do links directly to the files in the tree (default branch) on Github with `[TEXT](/PATH/TO/FILE)`.

PR #2, opened 2016-02-27
Fixed / updated links in README

IanLee1521:
Managed to break one of them in #1 and also, with the DNS alias I set up, now we can use http://software.llnl.gov/... instead of http://llnl.github.io/... for project links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Documentation
 
 To get started building and using Conduit, check out the full documentation:
 
-http://llnl.github.io/conduit/
+http://software.llnl.gov/conduit/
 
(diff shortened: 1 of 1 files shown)


cyrush:
thanks for the update -- its nice to be able to use software.llnl.gov

IanLee1521:
Yeah. Wasn't sure if you'd seen that transition I made, so thought I would do it as a PR rather than just changing it.

Trying to see if I might be able to namespace all of the GitHub repos down into something like, `software.llnl.gov/open-source/conduit` but currently I don't think I can.

PR #44, opened 2016-09-14
enh: add vector field to blueprint example meshes

cyrush:
adds a point-centered vector field to the 2d and 3d braid example
meshes, and hdf5 output.


cyrush:
address part of #12

PR #46, opened 2016-09-14
enh: add lines topo examples to mesh bp

cyrush:
adds a 2d and 3d example topologies that use lines.

these examples create meshes with lines elements that simply
represent the outline of the braid quads (2d), or hexs (3d) using
a naive approach.

With this approach there are lines that overlap spatially. This
was done so we could reuse the existing helper functions that
generate the values for the example element centered scalar field.
With more effort, we could create a better sample mesh.


cyrush:
this resolves the final part of #46

Issue #49, opened 2016-09-16
reconsider the names of Node::paths() and NodeIterator::path() methods


cyrush:
We are adding a path() function that returns a string with the full path to the given node

has_path and fetch path variants descend the tree, but, the new path() clashes a bit with a few other uses "path" in method names:

For example, in the Node and Schema classes:

- the two paths() variants list the names of the node's direct children, not full paths to them.

In the NodeIterator class:

- path() returns the name of the child, not the full path to the current child

I propose changing the path variants in the Node and Schema classes to 
- child_names() (or similar: names_of_children() , is closer in style to number_of_children()) 

And path() in NodeIterator to:
- name()

It's an API breaking change, but I think 0.2.0 is the right time to do this.

cyrush:
spoke with Brian Ryujin & @xjrc about this, both concur it make sense to make a change.
going to go with 

child_names() for the paths() methods in Node & Schema

and name() for the path() method NodeIterator

cyrush:
resolved in #71

Issue #38, opened 2016-09-08
generate blueprint index from blueprint tree


cyrush:

cyrush:
this will be included in PR #55

cyrush:
merged in pr #55

Issue #40, opened 2016-09-09
add initial blueprint verify for mesh, mesh::topology and mesh::field protocols


cyrush:

cyrush:
changed title, mcarray::verify already exists

cyrush:
merged in PR #55, still need more unit tests

cyrush:
resolved with @xjrc's changes in PR #72

PR #55, opened 2016-09-20
enh: add shell namespaces and verify for mesh bp

cyrush:
added implementation of mesh/coordset verify

added implementation of mesh verify that relies on:
 mesh/coordset (implemented)
 mesh/topology (unimplemented)
 mesh/field (unimplemented)

added info output to mcarray verify

wired generic interface for path based protocol identification

 ex:
  blueprint::verify(n,"mesh/coordset")
   maps to
  blueprint::mesh::coordset::verify(n,"mesh/coordset")

still evolving structure of the info output for verify, but
it looks like the following:

{
  valid = "true" or "false"
  (if errors)
      errors = ["error string 1", "error string 2"]

  child
  {
    valid = "true" or "false"
    (if errors)
        errors = ["error string 1", "error string 2"]
  }
  ...
}

this allows us to report errors at any part in a hierarchy, but
assumes protocols won't require the names "valid" or "errors"


src/libs/blueprint/blueprint_mcarray.cpp
```diff
@@ -92,30 +115,75 @@ verify(const Node &n)
 bool verify(const conduit::Node &n,
             Node &info)
 {
+    bool res = true;
+
+    // mcarray needs to be an object or a list
+    if( ! (n.dtype().is_object() || n.dtype().is_list()) )
```
xjrc: This is an incredibly minor complaint, but I feel like this line doesn't match the coding standards used for of the rest of Conduit.  In particular, it doesn't feel like the expanded spacing in this statement fits; something like `if(!(n.dtype().is_object() || n.dtype().is_list()))` seems more appropriate.

src/libs/blueprint/blueprint_mesh.cpp
```diff
@@ -94,8 +133,504 @@ bool
...
+            {
+                info["errors"].append().set(chld_name 
+                                + " is not a valid blueprint::mesh::coordset");
+                res = false;
+            }
+        }
+    }
+    
+    // required: "topologies",  with at least one child
+    // each child must conform to "mesh::topology"
```
xjrc: The functionality below for verifying the `topologies` path is nearly identical to that used above to verify the `coordsets` path.  The commonalities between these two segments should be factored out if possible.

src/libs/blueprint/blueprint_mesh.cpp
```diff
@@ -94,8 +133,504 @@ bool
...
+                    << "\"" << chld_name  << "\" "
+                    << "references a invalid coordset "
+                    << "\"" << coords_name  << "\" ";
+                info["errors"].append().set(oss.str());
+                res = false;
+            }
+        }
+    }
+    
+    // optional: "fields", each child must conform to "mesh::field"
```
xjrc: The functionality used below to verify the `fields` path is also nearly identical to the `coordsets` path verification functionality and should be considered during factoring.

xjrc: It may be worth noting in the `info` node that this section is optional as is done with the extra dimensions for nodes in the `coordset::verify_coordset_uniform` function below.

xjrc:
One quick addendum to my previous comment: using the schema validation functionality in `rapidjson` will require upgrading to version 1.1.0 (up from the current version of 1.0.2).  If this isn't feasible, then just disregard my previous suggestion.

src/libs/blueprint/blueprint_mcarray.cpp
cyrush: sounds good

src/libs/blueprint/blueprint_mesh.cpp
```diff
@@ -94,8 +133,504 @@ bool
...
+    {
+        const Node &dims = coordset["dims"];
+        
+        if(!dims.has_child("i"))
+        {
+            std::string msg ="missing coordset child \"dims\"";
+            info["errors"].append().set(msg);
+            res = false;
+        }
+        else if(!dims["i"].dtype().is_number())
```
cyrush: One of my goals is for the the code used to verify the protocol to reflect how the data will be consumed. 

Given this, I don't think we should use rapidjson here b/c that would require transforming the data (or its schema) into json.  At this point the data is not json  -- it's in conduit's in memory format. 

Even with an exemplar schema, or access to standard set of schema validation primitives I suspect we will still have to have some custom code for cases where schema validation doesn't get us all the way there. In that case we have two styles of validation which will be harder to maintain. 

We could try something similar in conduit (compare to exemplar nodes), but like schema validation I don't think that will cover all of what is admissible.

src/libs/blueprint/blueprint_mesh.cpp
cyrush: thats a good observation, if we can condense these checks into reusable functions that sounds like a good improvment

src/libs/blueprint/blueprint_mesh.cpp
cyrush: yes, we should add an info entry to optional indicating that fields exist.

src/libs/blueprint/blueprint_mesh.cpp
xjrc: You certainly bring up a lot of valid points, but I still think that leveraging `rapidjson` here will help us a lot more than it will hinder us.  If you don't mind, could we discuss this further during our meeting tomorrow afternoon?  If you're not convinced after I try to elaborate more on why I think this is a good idea, I won't pursue this any further; I just want to advocate a bit more for this idea because I think it could save us a lot of work.

cyrush: sure, feel free to explore it more. I don't know if we will have too much time to get into the details tomorrow, but if you have a prototype of how it would look that would help frame the discussion.

cyrush:
index gen commit will resolve #38

cyrush:
@xjrc -- this is coming along. The last piece is mesh:topology::verify. I may take a pass at it, but I will still need lots of help with testing.

Also wondering if how I am doing the nested namespaces in the .cpp file is confusing. 
Since most of the functions are "verify", it may be better to drop the namespace statements and simply write something like: 

```
bool
blueprint::mesh::topology::verify(const Node &n, Node &info)
{
}
```

instead of 

```
namespace mesh
{
namespace toplogy
{
bool
verify(const Node &n, Node &info)
{
}
..
}
}
```

Thoughts?

xjrc:
@cyrush: Even though the first version is more verbose, I like it better because I can easily just read the function title and know exactly which `verify` function I'm reading.  It doesn't seem like eliminating the namespace scopes will be too detrimental anyway since each namespace is relatively sparsely populated.

I'll try to take a look at this sometime later this week.  I'm pretty busy preparing for my presentation in a few weeks, but I'm going to try to be done by Friday.  After I'm finished, I'll start contributing some test cases.

cyrush:
thanks, I agree -- I'll make changes to the how namespaces are used in the cpp file.

cyrush:
@xjrc this should include all of the additions for mesh bp verify  (#40)  and index gen ( #38 )

#49 will require a good amount of changes to the new verify code -- I am wondering if we should merge this PR, and start a new branch / PR focused on unit tests to improve coverage of the new code (and any fixes we find) (https://coveralls.io/builds/8274435/source?filename=src%2Flibs%2Fblueprint%2Fblueprint_mesh.cpp). That way I can easily move in parallel to resolve #49. What do you think?

xjrc:
@cyrush: I agree; I think that merging this branch will make it much easier to move forward with testing.  I'm going to start writing some tests today, but I'll hold off on pushing them to a public branch until this is merged to make future merges simpler.

cyrush:
@xjrc  merged this pr, feel free to create a new branch and open a new pr

src/libs/blueprint/blueprint_mesh.cpp
```diff
@@ -74,34 +125,1720 @@ namespace conduit
...
+check_sph_coord_sys_axis_name(std::string &name)
+{
+    return ( name == "r" || name == "theta" || name == "phi");
+}
+
+//-----------------------------------------------------------------------------
+bool
+check_cyln_coord_sys_axis_name(std::string &name)
+{
+    return ( name == "r" || name == "z");
```
xjrc: @cyrush: Shouldn't cylindrical coordinates be expressed using 3 coordinates (i.e. r, z, and theta)?  If we're only planning to use cylindrical to describe cylindrical profiles then there isn't an issue, but I think that the third coordinate will be necessary for fully describing 3D geometries.

cyrush: rz is for 2D cylindrical, the "r" component collapses 2 comps from 3d. 
 
all the sim codes I know about only support 2D-cylindrical -- for the 3D case the r would be broken out into two components.

Issue #16, opened 2016-07-30
resolve unnecessary cmake transitive link dep for civet and b64


cyrush:
related to #14, we should try to resolve the underlying issue with PRIVATE causing a transitive link dependency for imported targets

cyrush:
the best solution for this is to simply use a cmake object library:

https://cmake.org/Wiki/CMake/Tutorials/Object_Library

PR #63, opened 2016-09-23
Bugfix/2016 09 windows type mapping

cyrush:


cyrush:
This will resolve #63, #16, and #5

PR #71, opened 2016-10-13
enh: use name instead of paths in several methods

cyrush:
change Node::paths() to Node::child_names()
change Schema::paths() to Schema::child_names()

remove copy out variants of Node::paths and Schema::paths

return const ref to empty string vec for case where Node
is not in the object role (avoid exception)

change NodeIterator::path() to NodeIterator::name()


cyrush:
this will resolve #49

cyrush:
resolves #49

PR #72, opened 2016-10-13
tests: add test cases for mcarray/mesh verify functions

xjrc:
The changes in this pull request will add test cases to verify the correctness of the `conduit::blueprint::mcarray::verify` and `conduit::blueprint::mesh::verify` functions in simple cases not covered by the existing example tests (see  #55 for more information).

### DIFF
--- a/src/libs/blueprint/blueprint_mesh.cpp
+++ b/src/libs/blueprint/blueprint_mesh.cpp
@@ -689,29 +689,26 @@ mesh::coordset::uniform::verify(const Node &coordset,
     }
     else
     {
-        const Node &dims = coordset["dims"];
-        
-        if(!mesh::logical_dims::verify(dims,info["dims"]))
-        {
-            res= false;
-        }
-    }
-    
-    if(!coordset.has_child("origin"))
-    {
-        log_optional(info,proto_name, "has origin");
-        
-        if(!mesh::coordset::uniform::origin::verify(coordset["dims"],
-                                                    info["dims"]))
+        if(!mesh::logical_dims::verify(coordset["dims"],info["dims"]))
         {
             res= false;
         }
     }
 
-    if(!coordset.has_child("spacing"))
+    if(coordset.has_child("origin"))
+    {
+        log_optional(info,proto_name, "has origin");
+
+        if(!mesh::coordset::uniform::origin::verify(coordset["origin"],
+                                                    info["origin"]))
+        {
+            res= false;
+        }
+    }
+
+    if(coordset.has_child("spacing"))
     {
         log_optional(info,proto_name, "has spacing");
-        
 
         if(!mesh::coordset::uniform::spacing::verify(coordset["spacing"],
                                                      info["spacing"]))
--- a/src/libs/blueprint/blueprint_mesh.cpp
+++ b/src/libs/blueprint/blueprint_mesh.cpp
@@ -1161,7 +1161,7 @@ mesh::topology::structured::verify(const Node &topo,
     else
     {
         const Node &topo_elements = topo["elements"];
-        
+
         if(!topo_elements.has_child("dims"))
         {
             log_error(info["elements"],proto_name, "missing child \"dims\"");
@@ -1176,6 +1176,9 @@ mesh::topology::structured::verify(const Node &topo,
             }
         }
     }
+
+    // FIXME: Add some verification code here for the optional origin in the
+    // structured topology.
 
     log_verify_result(info,res);
 
@@ -1200,7 +1203,7 @@ mesh::topology::unstructured::verify(const Node &topo,
     else
     {
         const Node &topo_elements = topo["elements"];
-        
+
         if(topo_elements.has_child("shape"))
         {
             // single shape case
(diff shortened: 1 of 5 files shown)


src/tests/blueprint/t_blueprint_mesh_verify.cpp
```diff
@@ -52,20 +52,292 @@
...
+    return is_valid;
+}
+
+/// Testing Constants ///
+
+const std::vector<std::string> LOGICAL_BASIS = create_basis("i","j","k");
+const std::vector<std::string> CARTESIAN_BASIS = create_basis("x","y","z");
+const std::vector<std::string> SPHERICAL_BASIS = create_basis("r","theta","phi");
+const std::vector<std::string> CYLINDRICAL_BASIS = create_basis("r","z");
+
```
cyrush: I know this is only for tests, but we actually use basis as one of the entires for the field protocol, so seeing it here though me off a bit

cyrush:
@xjrc looking good! chipping away at the coverage

src/libs/blueprint/blueprint_mesh.cpp
```diff
@@ -689,29 +689,26 @@ mesh::coordset::uniform::verify(const Node &coordset,
...
+    if(coordset.has_child("origin"))
     {
         log_optional(info,proto_name, "has origin");
-        
-        if(!mesh::coordset::uniform::origin::verify(coordset["dims"],
-                                                    info["dims"]))
+
+        if(!mesh::coordset::uniform::origin::verify(coordset["origin"],
+                                                    info["origin"]))
         {
```
cyrush: thanks for finding this bug!

src/tests/blueprint/t_blueprint_mesh_verify.cpp
cyrush: I think coordinate system is good (the tuples in a coord sets are anchored to some coord system)

cyrush:
awesome, thanks os much -- I'll take a look soon!

src/tests/blueprint/t_blueprint_mesh_verify.cpp
```diff
@@ -0,0 +1,1163 @@
...
+        n[coordsys_dim].set("test");
+        is_valid &= !coordsys_valid_fun(n,info);
+
+        n[coordsys_dim].set(10);
+        is_valid &= coordsys_valid_fun(n,info);
+
+        // FIXME: The coordinate system checking functions shouldn't accept
+        // systems such as (y) or (x, z); all successive dimensions should
+        // require the existence of previous coordsys dimensions.
+        /*
```
xjrc: That makes sense to me!  I'll just leave this code in until we make a final decision on whether or not these types of coordinate spaces will be supported.

src/tests/blueprint/t_blueprint_mesh_verify.cpp
```diff
@@ -0,0 +1,1163 @@
...
+    EXPECT_FALSE(blueprint::mesh::field::association::verify(n,info));
+}
+
+
+//-----------------------------------------------------------------------------
+TEST(conduit_blueprint_mesh_verify, field_basis)
+{
+    // FIXME: Does this have to be verified against anything else?  What does
+    // this basis refer to if it isn't a different path in the mesh structure?
+    Node n, info;
```
xjrc: Thanks for the clarification on this.  I guess it isn't too important that we verify it for now, but I think that keeping the 'fixme' here will be helpful to remind us to expand on this test if/when we decide to expand the `blueprint::mesh::field::basis::verify` function.

src/tests/blueprint/t_blueprint_mesh_verify.cpp
```diff
@@ -0,0 +1,1163 @@
...
+
+        { // Type Field Tests //
+            n.remove("type");
+            EXPECT_FALSE(verify_topology(n,info));
+            n["type"].set("explicit");
+            EXPECT_FALSE(verify_topology(n,info));
+
+            // FIXME: Remove the comments from the following line once the verify functions
+            // for uniform and rectilinear topologies have been implemented.
+            const std::string topology_types[] = {/*"uniform", "rectilinear", */"structured"};
```
xjrc: These functions do exist and are currently implement as no-ops, but wiring them up here wouldn't be too helpful.  The purpose of this part of the test case is to verify that a valid "explicit" topology isn't accepted as any other form of topology when fed through the `blueprint::mesh::topology::verify` function (essentially just ensuring that the if clauses in this function point to the correct subroutines).  Since the "uniform" and "rectilinear" topology tests are no-ops that return true, calling them here would cause this test to fail (this is actually good because this indicates that these verify functions aren't returning values we'd expect).  Ultimately, I think that it's best to keep this commented out for now, especially considering that we already have no-op coverage for the "uniform" and "rectilinear" cases in the "topology_uniform" and "topology_rectilinear" test cases above.

cyrush: got it, thanks!

src/tests/blueprint/t_blueprint_mesh_verify.cpp
cyrush: agreed.

cyrush:
this all looks good! Thanks so much for getting this in.

For some reason the appveyor (windows) checks didn't run -- I'll resolve anything after I merge. 

In the future, feel free to create a branch off of the main repo instead of forking. It will make it easier for use to work closer together and make sure all the CI stuff works.

Issue #81, opened 2016-11-02
bp: change field assoc "point" to "vertex"


cyrush:
reserve point as a shape, vertex is a better description for how the field is associated with the mesh (matches mfem nomenclature)

cyrush:
merged in PR #83

PR #83, opened 2016-11-02
bugfix: change point to vertex for field assoc

cyrush:


cyrush:
will resolve issue #81

PR #95, opened 2016-11-22
Reverted software.llnl.gov URLs to llnl.github.io

IanLee1521:
This is a "workaround" to the break that happened last week when we enabled HTTPS on the software.llnl.gov site.


cyrush:
since the redirect is back in working order -- I am going to close this pr.

Issue #96, opened 2016-11-29
rectilinear coords values are not necessarily mcarrays


cyrush:
since we are doing a cartesian product, the length of the coords arrays can be different.

cyrush:
resolved in #110

PR #110, opened 2016-12-08
bugfix: fix bug in mesh bp recti coord set verify

cyrush:
The values for rectilinear coordsets aren't necessarily
mcarrays, since they use an implied cross product of
separate arrays to define the actual coordset.

This update changes verify for the rectilinear coordset
case  to check that "values" has numeric children, instead
of checking that "values" is an mcarray.


cyrush:
this will resolve #96

Issue #97, opened 2016-11-29
support set_external(const Node &data)?


cyrush:
set_external(Node &) effectively allows you to create a shallow copy the passed node. 

Since it does not modify the structure of the passed Node, I think we should also support a variant that allows the passed node to be const.

ajkunen:
I like this idea…  I’m not sure how you would implement the const version…  If it’s const, does that mean that the entire node’s subtree is const?  how is the constness ‘stored’ in the node?

cyrush:
we already do have a const interface that prevents any changes from the structure of a node. 
This allows you to iterate over the values (using const iterators, etc) but disables any dynamic changes to the node tree. 

That is what we would be abiding by in this case for the input data.  

There are limits to the power of constness here,  depending on what aspects of the interface are used we can actually guard against data changes. Pointers can only be accessed as const type * pointers, etc.


In this case, the new node could still modify values if it is used in a non const manner, but there is a contract not to modify the passed tree.

ajkunen:
Then that sounds awesome!

-Adam

cyrush:
resolved in pr #111

PR #111, opened 2016-12-08
enh: set_external now takes const Node ref

cyrush:
the typical use case of this is a shallow copy
the node structure is duplicated, but e data is zero-copied.
For this use case you want a const guarantee for the node
structure of the passed ref.


cyrush:
this will resolve #97

Issue #99, opened 2016-11-30
add filesystem utils related to dirs


cyrush:
conduit::utils::create_directory(const std::string &path,  bool recursive=true);
conduit::utils::remove_directory(const std::string &path, bool recursive=true);

cyrush:
also add access to proper file sep char & add rsplit_path()

cyrush:
going to stick with providing non-recursive variants:

create_directory(path)
remove_file_path(path) -- works for empty dirs or files

cyrush:
resolved in #119

PR #119, opened 2016-12-19
enh: add file system path split helpers

cyrush:
also expose the proper platform specific path separator


cyrush:
this will resolve #99

Issue #139, opened 2017-02-02
bug with locating web_client_root for installed relay node viewer


cyrush:
Error launching Conduit Relay Node Viewer Server:

```
{
  "file": "/private/var/folders/wj/jjhjt_0d3kg1vmk56d5bq_38001kr8/T/harrison37/spack-stage/spack-stage-tQRXse/conduit-0.2.1/src/libs/conduit/conduit_node.cpp",
  "line": 11780,
  "message": "Node::as_string() const -- DataType empty at path web_client_root does not equal expected DataType char8_str"
}

```

It seems web_client_root  root is empty,  it may not be set -- or we may be calling the wrong "about"

cyrush:
resolved in #142

PR #142, opened 2017-02-04
bugfix: fix web root detection for installs

cyrush:
fixed problem with function that locates the web root
(it was using the wrong about() routine)


cyrush:
this resolves #139 #142

Issue #107, opened 2016-12-08
add multi-domain verification support to blueprint verify


xjrc:
The `conduit::blueprint::mesh::verify` function needs to be adjusted so that it's capable of verifying the contents of multi-domain input meshes.  For reference, a multi-domain input mesh is specified in the following way:

```
mesh: {
  domain0: {
    coordsets: { /* ... */ },
    topologies: { /* ... */ },
    fields: { /* ... */ }
  },
  /* ... */
  domainN: {
    coordsets: { /* ... */ },
    topologies: { /* ... */ },
    fields: { /* ... */ }
  }
}
```

PR #148, opened 2017-02-22
enh: add support for multidomain mesh verification

xjrc:
The changes in this pull request add support for multidomain mesh verification to Conduit (fixing issue #107).  These changes alter the Conduit interface in the following ways:

- Update `conduit::blueprint::mesh::verify`: This function is now capable of verifying standard meshes and multidomain meshes.
- Add `conduit::blueprint::mesh::is_multidomain`: This function returns true iff the given Blueprint mesh is a multidomain mesh.
- Add `conduit::blueprint::mesh::to_multidomain`: This function transforms the first single domain mesh argument to a multidomain mesh and replaces the second parameter node with this new mesh.

Please let me know if I should update these changes in any way to fix bugs, interface inconsistencies, or any other issues!


src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -364,6 +358,23 @@ bool mesh::is_multidomain(const Node &n,
 }
 
 
+//-----------------------------------------------------------------------------
+bool
+mesh::verify(const Node &n,
+             Node &info)
+{
+    return mesh::verify_multidomain(n, info) || mesh::verify_unidomain(n, info);
```
xjrc: One thing I wanted to highlight in this PR was the implementation of this function.  While this function works properly insofar as it returns the correct value, it doesn't always put the right information into the `info` node.  Currently, this function always returns error information from the `blueprint::mesh::verify_unidomain` (i.e. the old `blueprint::mesh::verify` function), which can be incorrect if the user's mesh intends perform a multidomain mesh check.  Going forward, we need to decide what in the structure of the given mesh node classifies it as unidomain or multidomain so that the error output of this function is more helpful for multidomain verifications.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -307,6 +311,83 @@ mesh::verify(const Node &n,
...
+                chld_name = oss.str();
+            }
+
+            if(!mesh::verify_unidomain(chld, info[chld_name]))
+            {
+                log_error(info,proto_name,
+                          "child " + chld_name + " is not a valid mesh");
+                res = false;
+            }
+        }
```
cyrush: This looks good, I think we should add a top level info entry that says we have a compliant multi domain  mesh.

src/libs/blueprint/conduit_blueprint_mesh.cpp
cyrush: I think we can check single domain, if that fails check multi. 
We can clear the info for the second case.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -59,6 +59,10 @@ using namespace conduit;
 // access verify logging helpers
 using namespace conduit::blueprint::utils;
 
+namespace conduit { namespace blueprint { namespace mesh {
+    bool verify_unidomain(const conduit::Node &n, conduit::Node &info);
```
cyrush: Minor style thingy:

instead of uni can we use verify_single_domain?
and verify_multi_domain?

src/libs/blueprint/conduit_blueprint_mesh.cpp
cyrush: Along the lines of:

`log_info(info,proto_name,"is a multi domain mesh");`

src/libs/blueprint/conduit_blueprint_mesh.hpp
```diff
@@ -99,6 +99,20 @@ bool CONDUIT_BLUEPRINT_API verify(const std::string &protocol,
...
+
+//-----------------------------------------------------------------------------
+/// blueprint mesh property and transform methods
+/// 
+/// These methods can be called on any verified blueprint mesh.
+//-----------------------------------------------------------------------------
+
+//-------------------------------------------------------------------------
+bool CONDUIT_BLUEPRINT_API is_multidomain(const conduit::Node &n);
+
```
cyrush: is_multi_domain and to_multi_domain

src/tests/blueprint/t_blueprint_mesh_verify.cpp
```diff
@@ -148,6 +148,11 @@ bool verify_field_index_protocol(const Node &n, Node &info)
     return blueprint::mesh::verify("field/index",n,info);
 }
 
+bool verify_mesh_multidomain_protocol(const Node &n, Node &info)
+{
```
cyrush: multi_domain

xjrc:
@cyrush: I've revised the changes in this pull request based on your feedback.  The only deviation I made from your suggestions was in the implementation of the `blueprint::mesh::verify` function, which I just changed to return the `info` node of the corresponding successful verification (and the single verify `info` initialization in the event of a double failure).  Please let me know if you'd like me to make any other changes and thanks for your help!

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -307,6 +311,98 @@ mesh::verify(const Node &n,
...
+    return mesh::verify_multi_domain(n, info);
+}
+
+
+//-------------------------------------------------------------------------
+bool mesh::to_multi_domain(const conduit::Node &n,
+                           conduit::Node &dest)
+{
+    dest.reset();
+
```
cyrush: one more thing: the transforms should work for all cases where mesh verify is true. 

For this, we need to also handle the case where the input node is already a multi domain mesh.

xjrc: Good catch!  I've fixed this in my latest commit.

cyrush:
Looks good, thanks for adding this!

xjrc:
@cyrush: Would it be okay for me to merge this into master?  I'd like to get this in so that I can use it in other code, but there's no big rush if you'd prefer to wait.

cyrush:
yes, sorry -- feel free to squash merge.

PR #149, opened 2017-03-07
enhancments to relay mpi

cyrush:


cyrush:
@bryujin  -- added some docs, with the hope that they capture the overall strategy.

Still need to work on names for the "generic" and "known schema" cases for both the API and the docs.  If we could sit down and get something we are both happy with that would help me move forward.

cyrush:
@bryujin can we converge on the naming issue (plural vs singular for schema):

Here are the cases:

singular:
- send_using_schema
- recv_using_schema
- broadcast_using_schema

plural:
- gather_using_schemas
- allgather_using_schemas


In the singular cases, there is only one schema sent. 

In the case of gather, each nodes schema is sent, so that is why I was leaning towards plural

src/libs/relay/conduit_relay_mpi.cpp
```diff
@@ -113,171 +113,618 @@ rank(MPI_Comm mpi_comm)
...
+    }
+
+    return res;
+}
+
 
 //---------------------------------------------------------------------------//
 int 
-send(Node &node, int dest, int tag, MPI_Comm comm)
+send_using_schema(const Node &node, int dest, int tag, MPI_Comm comm)
```
bryujin: I'll vote for everything to be using_schemas. Technically, you have a schema on both sides, even if they're the same one.

src/libs/relay/conduit_relay_mpi.cpp
```diff
@@ -113,171 +113,618 @@ rank(MPI_Comm mpi_comm)
...
+    else
+    {
+        node.compact_to(snd_compact);
+        snd_schema = snd_compact.schema().to_json();
+        snd_ptr = snd_compact.data_ptr();
+    }
+    
+    int snd_schema_size = snd_schema.length() + 1;
 
-    int intArray[2] = { schema_len, data_len };
```
cyrush: On thing that constrains 2 vs 3 messages -- if we use one big-buffer,  we can directly pass the existing buffer to MPI (in the case where that is allowed). Its a chicken and egg issue as well, we can't identify this case until the schema is parsed. 

Can we avoid the first message (with counts) totally by sending the schema, and using mpi probe?

src/libs/relay/conduit_relay_mpi.cpp
```diff
@@ -113,171 +113,618 @@ rank(MPI_Comm mpi_comm)
...
+                             1,
+                             MPI_INT,
+                             src,
+                             tag,
+                             comm,
+                             &status);
+                             
+    CONDUIT_CHECK_MPI_ERROR(mpi_error);
+
+    Node rcv_buffers;
```
cyrush: I see your point -- schema payloads will always be char *s.  I will change this case. 

My main reason for doing this was to prevent any leaks. We will still need an optional Node buffer for the recv result (one which allocs for the proper type) so we definitely need to use a node there. 



For the more complex cases (ex: gather) I do feel using a node to manage the temp allocs and pointers needed makes things much nicer.

src/libs/relay/conduit_relay_mpi.cpp
cyrush: Ok, we want to settle on singular vs plural, instead of mixed usage. 

I keep going back and forth on this. 

Here is an argument for using all singular instead of all plural, trying to convince myself things are clearer value-wise vs instance wise, and the singular supports that:

For all current cases (send,recv,broadcast,gather) the input is a single node that has a 
schema. 

In all cases other than gather, the output schema is the effectively (value-wise) the transmitted schema.


Even with gather, the input values of the sent schema are preserved. It's iffy-er,  but gather_using_schema() doesn't seem awful. 


When transmitting we aren't doing anything like splitting the input's children (and as a result the input schema into "schemas") to provide any of these ops.  In fact, considering doing that is what we felt would made simple extensions of our current model to scatter and all-to-all awkward.

  
The fact that there is a copy (new instance) over the network of course happens, but I don't think most folks would infer that detail from the plural form.

cyrush:
@bryujin When you get a chance, can you look at the send changes to see if I am on the right track?

src/libs/relay/conduit_relay_mpi.cpp
```diff
@@ -113,32 +113,235 @@ rank(MPI_Comm mpi_comm)
...
+        n_msg["schema"].set_external_char8_str(const_cast<char*>(snd_schema.c_str()));
+    }
+    
 
-    Schema schema_c;
-    node.schema().compact_to(schema_c);
-    std::string schema = schema_c.to_json();
-    int schema_len = schema.length() + 1;
+    Node n_msg_compact;
+    n_msg.compact_to(n_msg_compact);
```
bryujin: Is it possible to start with this compacted schema and then push things into it in the if/else above? It's a pretty simple schema and you would just have to preallocate the data. Otherwise, looks good.

cyrush: I'll look into this, as long as Node::compact_to has enough smarts to avoid realloc for the non-contig case (which I believe it does, but I want to scrutinize) this could work

src/libs/relay/conduit_relay_mpi.cpp
```diff
@@ -113,32 +113,231 @@ rank(MPI_Comm mpi_comm)
...
+    return res;
+}
+
 
 //---------------------------------------------------------------------------//
 int 
-send(Node &node, int dest, int tag, MPI_Comm comm)
-{ 
+send_using_schema(const Node &node, int dest, int tag, MPI_Comm comm)
+{
```
cyrush: @bryujin here is the new `send_using_schema`,  following your suggestions I think it looks much better.

What do you think?

cyrush:
resolved a few corner case memory access issues with node, memcheck now only reports issues with in python (our current steady ground)

cyrush:
Resolved the random crash issue with `t_relay_mpi_test`

The crash was happening in mpi finalize, one of the test cases had a mismatch between the number of i-msgs and the count passed to waitall. 

This was hard to track down, this post provided the final clue: 
 
https://stackoverflow.com/questions/8418716/why-wont-mpi-finalize-gracefully

cyrush:
@bryujin  I think this is ready to merge, do you want to to look at it any more -- or can I proceed?

cyrush:
thanks @bryujin !

Issue #160, opened 2017-04-20
json handling of inf and nan


cyrush:
ran into a case where badly formatted inf cased a json parsing issue. 
'inf.0',

Note: json doesn't have suppport for inf and nan, however rapidjson may have extra support for it.

cyrush:
newer versions of rapidjson provide kParseNanAndInfFlag, which handles inf, -inf and nan. 
Again, json itself doesn't handle these, so for the web browser we may need to send as strings.

If we present as strings, we could look for "-inf", "inf", and "nan" strings as part of json sanitize and fix them before handing to rapidjson. However need need to make sure we only transform json values (and not ids), which complicates the json santize logic.

cyrush:
resovled in #162

PR #162, opened 2017-04-27
fix malformed nan and inf json output

cyrush:


cyrush:
This PR includes fixes to avoid creating malformed json and creates string-enclosed values for -inf, inf, and nan as described in #160 

We aren't supporting round trip parsing yet. When parsed, these become char8_str nodes, not float64 nodes set to the proper values. Newer rapidjson has better support, but transparently supporting from string-enclosed variants of these will require complex preprocessing logic in json_sanitize -- not sure if the benefit is strong enough.

Issue #169, opened 2017-05-26
use ints in c-api instead of bools


cyrush:
instead of #167, the more portable solution for C is to used ints a return values instead of using stdbool.

cyrush:
resolved in #171

PR #171, opened 2017-05-30
use int instead of bool in c-api for wider compat

cyrush:
changes all instances of bool to in in the c-api
for conduit + blueprint.

bool is in C99, but requires include of stdbool.h
using int instead of bool will make c-api easier
to use on a wider set of compilers.


cyrush:
this will resolve #169

cyrush:
resolves #169

PR #179, opened 2017-06-13
refactor travis logic into bash scripts

cyrush:


cyrush:
Closing: It's much simpler to test docker on circle ci and keep our existing travis logic.

Trying to keep sudo: false builds for all current cases, while opting in to it just for the docker test was painful.  Also, I fear our docker builds on travis will face timeout issues.  Circle allows us to easily increase the timeout for no output.

PR #180, opened 2017-06-14
enh: update blueprint documentation for materials/multi-domain

xjrc:
The changes in this pull request update the documentation for Blueprint so that it includes information about materials and multi-domain meshes.  The following is a short summary of the changed/added sections:

- Added the `Material Sets` and `Domain Adjacencies` sections.
- Updated the `Protocol` section to include information about multi-domain Blueprint meshes.
- Updated the `Fields` section to explain integration with `Material Sets`.

PR #185, opened 2017-06-16
enh: update blueprint::verify for materials/multi-domain

xjrc:
The changes in this pull request will update the `blueprint::verify` functions so that they support the all of the latest material and multi-domain features (see #180 for more information).  The fully enumerated list of changes is as follows:

- [X] Add `blueprint::matset::verify` to validate Blueprint material information.
- [X] Add `blueprint::matset::index::verify` to validate Blueprint index material information.
- [X] Add `blueprint::domain_adjacency:verify` to validate Blueprint abutment information
- [x] Add `blueprint::domain_adjacency::index::verify` to validate Blueprint index abutment information..
- [X] Update `blueprint::field::verify` to support validation of per-material fields.
- [X] Update `blueprint::field::index::verify` to support validation of per-material field indexes.
- [X] Update `blueprint::verify` to include validation for material sets and domain adjacencies.
- [x] Update the `blueprint::mesh::examples::braid` function to include materials and domain adjacencies.
- [x] Add test cases for all of the new/updated verification functions.


src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -65,6 +65,142 @@ namespace conduit { namespace blueprint { namespace mesh {
 } } }
 
 //-----------------------------------------------------------------------------
+// -- begin internal helper functions --
+//-----------------------------------------------------------------------------
+
+//-----------------------------------------------------------------------------
```
xjrc: @cyrush: When you get a chance, could you take a look at these functions below (i.e. `verify_*_field`) and let me know whether or not they're at a good level of abstraction?  I just want to verify that these functions are close to the abstraction level that we discussed so that I can make changes if necessary before the PR is merged.

cyrush: @xjrc out of cycles for now, but I'll try to get you a review Monday afternoon.

src/libs/conduit/conduit_node_iterator.cpp
```diff
@@ -135,6 +137,24 @@ NodeIterator::index() const
 }
 
 //---------------------------------------------------------------------------//
+std::string
+NodeIterator::id() const
```
xjrc: @cyrush: I added this new `conduit::NodeIterator::id` function in order to simplify the process of tagging the children of a node that can either be an object or a list.  Does this concept seem reasonable (i.e. a string identifier that works for both objects and lists) and, if so, does the name `id` seem adequately descriptive?  If not, please let me know and I'll modify this pull request.

cyrush: @xjrc, for some reason I thought that "name()" already did this logic, but I guess that wasn't the case?

xjrc: @cyrush: I ran a test case just now to determine the exact behavior; attempting to call `name()` on a list iterator causes the following error:
```
{
  "file": /path/to/conduit/src/libs/conduit/conduit_schema.cpp",
  "line": 1191,
  "message": "<Schema::object_hierarchy[OBJECT_ID]>: Schema is not OBJECT_ID"
}
```
It looks like the `object_hierarchy` variable used by `name()` is only valid for objects.

If the `name()` function is supposed to return a valid string for list iterators, then I can just replace the `name()` implementation with the new `id()` function's contents.  If not, then I'm content keeping `id()` or renaming it to something more appropriate.

xjrc:
@cyrush: With the most recent commit, this PR is more or less ready.  There are a few other things that we should discuss before this is merged (e.g. materials/multi-domain examples, status of the Blueprint enumeration verifiers, updated documentation), but I think the content of this PR is ready for review.

src/docs/sphinx/blueprint_mesh.rst
```diff
@@ -282,27 +283,67 @@ specified using a single shape topology for each element shape.
...
+
+
+Adjacency Sets
+++++++++++++++++++++
 
+Adjacency Sets are used to outline the shared geometry between subsets of domains in multi-domain meshes.
 
- * High Order Fields:
+Each entry in the Adjacency Sets section represents an adjacency group, which is meant to encapsulate a set of adjacency information shared between domains.
+Each individual adjacency group contains a source topology, an element association, and list of shared element groups (which themselves contain a subset of adjacent neighbor domains IDs and a list of shared element IDs).
```
cyrush: the example is correct, but I think this text needs to be udpated. 
It says each adj group specifies a topo, now its each adj set (prior bullet)

cyrush:
@kennyweiss Joe & I discussed `adjsets` with Rob to get his blessing and it was also integrated into mfem. 

The long form "domain_adjacencies"  did not fit the mold of the other entires, and b/c he was typing this often this was painfully apparent to Joe. 

We couldn't use `adjgroups` b/c with the evolved design --  each `adjset` is actually a collection of adjacency groups.

kennyweiss:
Looks good @xjrc.

Thanks for the explanation @cyrush.

xjrc:
@cyrush: I made all the revisions that we discussed (e.g. moving test cases, keeping enumeration verify functions), so I think this PR is ready for a final review whenever you have time.  I'll merge these changes 
as soon as we hammer out the final details.

PR #195, opened 2017-07-07
add bounds check to Node::child

cyrush:


cyrush:
note: hdf5 fetch fail in circle docker tests should be resolved on master

PR #197, opened 2017-07-10
add mesh topology transformations to blueprint

xjrc:
The changes in this pull request add coordinate set and topology transformation operations to Blueprint, which will allow client codes to easily "upgrade" any coordinate set/topology to a more explicit representation to suit their needs (e.g. transform a given uniform topology into an equivalent rectilinear topology).  The full list of changes contained in this pull request is as follows:

- [x] Add stubs for every transformation operation.
- [x] Add all coordinate set transformation operations.
  - [x] Add `blueprint::mesh::coordset::uniform::transform`.
  - [x] Add `blueprint::mesh::coordset::rectilinear::transform`.
  - [x] Add `blueprint::mesh::coordset::_explicit::transform`.
  - [x] Add `blueprint::mesh::coordset::transform` (protocol function).
- [x] Enforce consistency of input/output data types in all the coordinate set transformation operations.
- [x] Add all topology transformation operations.
  - [x] Add `blueprint::mesh::topology::points::transform`.
  - [x] Add `blueprint::mesh::topology::uniform::transform`.
  - [x] Add `blueprint::mesh::topology::rectilinear::transform`.
  - [x] Add `blueprint::mesh::topology::structured::transform`.
  - [x] Add `blueprint::mesh::topology::unstructured::transform`.
  - [x] Add `blueprint::mesh::topology::transform` (protocol function).
- [x] Enforce consistency of input/output data types in all the topology transformation operations.
- [x] Add test cases for all coordinate set transformation operations.
- [x] Add test cases for all topology transformation operations.


xjrc:
It turns out that supporting input/output data consistency is trickier than I thought it would be. The primary reason for this is Conduit's lack of data conversion support between equivalent types of different bit widths (e.g. `int{8|16|32|64}`, `uint{8|16|32|64}`). Attempting to perform these conversions naively using one of the `Node::set` operations will cause a data type transform, which is highly undesirable from the perspective of data consistency. It is also difficult to write this functionality inline since there are nontrivial intricacies involved in detecting a node's data type and applying the proper data transform(s). With all of that being said, I think it's best to merge these changes as they are and enhance them later with data consistency once conversion-style functions are implemented in the interface.

cyrush:
@xjrc  I am fine with merging, but not sure why we can't use the to_zzz() methods vs set?

xjrc:
@cyrush: We can't because "zzz" is a dynamic value in the general case (e.g. defined by user input), so we don't know which of the "to_" methods to call. It's not hard to write some code to detect what "zzz" is and properly dispatch the conversion to a "to_" function, but this code will be needed in many different places and thus merits abstraction.

xjrc:
@cyrush: I've finished implementing everything I wanted to get in for this PR. Would you mind giving it a once-over to make sure that all the new functionality is adequately named and organized? With your approval, I'll merge this and work on integrating it into #285.

src/libs/conduit/conduit_node.cpp
```diff
@@ -9513,6 +9513,147 @@ Node::to_double_array(Node &res) const
...
+void
+Node::to_data_type(index_t dtype_id, Node &res) const
+{
+    switch(dtype_id)
+    {
+        /* ints */
+        case DataType::INT8_ID:
+        {
+            if(this->dtype().number_of_elements() == 1)
+            {
```
cyrush: do we need the 1 ele cases? I think to_array will handle both?

src/libs/conduit/conduit_data_type.hpp
```diff
@@ -380,6 +380,7 @@ class CONDUIT_API DataType
 //-----------------------------------------------------------------------------
     index_t     id()    const { return m_id;}
     std::string name()  const { return id_to_name(m_id);}
+    std::string group() const { return id_to_group(m_id); }
```
cyrush: Do we need a new "group" concept to be in the public interface?

We already have:

    bool        is_number()           const;
    bool        is_floating_point()   const;
    bool        is_integer()          const;
    bool        is_signed_integer()   const;
    bool        is_unsigned_integer() const;

It seems the group "strings" kind of capture this, can these be used?

src/libs/conduit/conduit_data_type.cpp
```diff
@@ -552,6 +552,15 @@ DataType::spanned_bytes() const
...
 }
 
+//---------------------------------------------------------------------------//
+bool
+DataType::convertible(const DataType& dtype) const
+{
+    return ( (group() == dtype.group()) &&
+             (m_ele_bytes >= dtype.m_ele_bytes) &&
+             (m_num_ele >= dtype.m_num_ele) );
+}
```
cyrush: I believe any numeric type can be converted to any other numeric.  Seems like a property of just the type id, not the full DataType.

Seems odd we check the bytes and the number of elements?

Does this try to capture if data will be lost? Again the semantics are simply that of a c cast, creating our own semantics gets confusing.

src/libs/blueprint/conduit_blueprint_mesh.hpp
```diff
@@ -157,6 +162,10 @@ namespace coordset
         bool CONDUIT_BLUEPRINT_API verify(const conduit::Node &n,
                                           conduit::Node &info);
 
+        //---------------------------------------------------------------------
+        bool CONDUIT_BLUEPRINT_API transform(const conduit::Node &n,
+                                             conduit::Node &dest);
```
cyrush: while it makes sense for verify -- transform w/o a protocol seems too generic.

for these leaf case transforms, what about `to_uniform()`, etc (I think this mirrors our mcarray transforms)

src/libs/conduit/conduit_node.cpp
xjrc: I just tested this out and you are correct! I'll amend this with my next set of changes.

src/libs/conduit/conduit_data_type.hpp
xjrc: As we discussed earlier, I'm not thrilled with the `group` concept either. I'll remove it for the time being, but I do think some general "these are trying to represent exactly the same data with different bitwidths" concept would be useful.

src/libs/conduit/conduit_data_type.cpp
xjrc: > I believe any numeric type can be converted to any other numeric. Seems like a property of just the type id, not the full DataType.

Yeah, it seems like "convert" really isn't the right verb here. Due to this ambiguity, I've removed this public interface and moved the functionality into the private `find_widest_type` function instead.

> Seems odd we check the bytes and the number of elements?

In order for type A to be "upscalable" to type B (or whatever the right verb is), these types must target the same form of data (e.g. floats, unsigned integers, signed integers, etc.) and type A's data needs to fit into type B. In order to fit, type B needs to be larger in both per-element bytes (or else there's the potential for per-element truncation) **and** total number of elements (or else there's the possibility of array value truncation).

> Does this try to capture if data will be lost? Again the semantics are simply that of a c cast, creating our own semantics gets confusing.

As I envision it, "upscaling" a type comes with no possibility of data loss because the upscaled type will always have at least as many bytes as the type being upscaled.

src/libs/blueprint/conduit_blueprint_mesh.hpp
xjrc: The way this is organized actually confuses me a bit because the style in `mcarray` seems to conflict with that used in `mesh`. We can discuss this later; it's difficult to convey in this context.

xjrc:
@cyrush: I made all of the changes we discussed! The only thing I left out was updating the signatures of the `conduit::blueprint::mcarray::to_{contiguous|interleaved}` functions to be void returns; those functions had Fortran compatibility that I kept breaking by trying to make that change. If that small omission is permissible, then I think this pull request is finally ready to be merged.

src/libs/blueprint/conduit_blueprint_mesh.hpp
```diff
@@ -291,6 +333,11 @@ namespace topology
         //---------------------------------------------------------------------
         bool CONDUIT_BLUEPRINT_API verify(const conduit::Node &n,
                                           conduit::Node &info);
+
+        //---------------------------------------------------------------------
+        bool CONDUIT_BLUEPRINT_API transform(const conduit::Node &n,
+                                             conduit::Node &dest,
+                                             conduit::Node &cdest);
     }
```
cyrush: did we miss one `transform`?

xjrc: Good catch! This extraneous function declaration has now been removed.

PR #200, opened 2017-07-25
Use blt for conduit's core build system

cyrush:


cyrush:
it looks like coverage dropped b/c w/ blt we are now building fortran libs with coverage flags.

cyrush:
resolved all issues, but I want to do a final review before merging.

PR #206, opened 2017-08-30
update blt and add mpi testing via msmpi on appveyor

### LINKED ISSUE #205: Add MSMPI testing to appveyor
MFEM has an example:

https://github.com/mfem/mfem/blob/master/.appveyor.yml

cyrush:
Add recipe for using msmpi in appveyor following @kennyweiss's solution for blt. 

resolves #205

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -42,7 +42,7 @@ before_build:
   #- cmake -Hsrc -Bbuild -G "%CMAKE_GENERATOR%" -DHDF5_DIR="C:\\Program Files\\HDF_Group\\HDF5\\1.8.18" -DENABLE_PYTHON=TRUE
   #
   # enable mpi support and use default python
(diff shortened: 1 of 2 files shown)


cyrush:
Looks like we have a static vs shared libs issue w/ mpsmpi

kennyweiss:
Looks like there is a problem with how ms-mpi is interpreting ``long`` and ``unsigned long``

>   C:\projects\conduit\src\libs\relay\conduit_relay_mpi.cpp(127): error C2196: case value '5' already used [C:\projects\conduit\build\libs\relay\conduit_relay_mpi.vcxproj]
  C:\projects\conduit\src\libs\relay\conduit_relay_mpi.cpp(137): error C2196: case value '9' already used [C:\projects\conduit\build\libs\relay\conduit_relay_mpi.vcxproj]

Lines 127 and 137 are:
>   case CONDUIT_NATIVE_LONG_ID:   res = MPI_LONG;  break;
     case CONDUIT_NATIVE_UNSIGNED_LONG_ID: res = MPI_UNSIGNED_LONG; break;

cyrush:
@kennyweiss, yes -- just caught that, there can't use a switch b/c some of native id values are overloaded on windows

PR #238, opened 2017-11-30
refactor api export headers

cyrush:
* generate headers cmake configure file
* provide proper exports for static builds on windows
* cleanup and fix mistakes


cyrush:
after staring a vs projects: seems object libs (like conduit_b64) are wrongly tied to the Multi-threaded Debug DLL (`/MDd`)  run time instead of /MTd Multi-threaded runtime (for static libs)

cyrush:
overriding the  conduit_b64 in VS target to use `/MTd` resolves the issue, need to find out why cmake is creating a bad project for this object lib.

cyrush:
object libs issue reported for old cmake that in the balk park, but not exactly the same:
https://stackoverflow.com/questions/18410689/cmake-visual-studio-project-created-using-object-library-feature

cyrush:
Ok, seems like a clear cmake bug:

difference between the culprit lib and other object lib I am using seems to be the linker language. 

This lib is pure c, if I spoof cmake by changing the file extensions to cpp files I get the proper static runtime (MTd).

I then have to edit my lib headers to add extern C clauses (since we are now using the C++ compiler), and everything links.

cyrush:
Created a simple example and could not replicate, I think the issue is actually caused by the inclusion of gtest.  Gtest `fixes` the runtime lib flags for CXX, but doesn't for CPP. 

```
macro(fix_default_compiler_settings_)
  if (MSVC)
    # For MSVC, CMake sets certain flags to defaults we want to override.
    # This replacement code is taken from sample in the CMake Wiki at
    # http://www.cmake.org/Wiki/CMake_FAQ#Dynamic_Replace.
    foreach (flag_var
             CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
             CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
      if (NOT BUILD_SHARED_LIBS AND NOT gtest_force_shared_crt)
        # When Google Test is built as a shared library, it should also use
        # shared runtime libraries.  Otherwise, it may end up with multiple
        # copies of runtime library data in different modules, resulting in
        # hard-to-find crashes. When it is built as a static library, it is
        # preferable to use CRT as static libraries, as we don't have to rely
        # on CRT DLLs being available. CMake always defaults to using shared
        # CRT libraries, so we override that default here.
        string(REPLACE "/MD" "-MT" ${flag_var} "${${flag_var}}")
      endif()

      # We prefer more strict warning checking for building Google Test.
      # Replaces /W3 with /W4 in defaults.
      string(REPLACE "/W3" "/W4" ${flag_var} "${${flag_var}}")
    endforeach()
  endif()
endmacro()
```

Issue #136, opened 2017-01-31
conduit::Node::compare() method


cyrush:
Something I (Cyrus) have wanted to do for a bit – also a request from Mark Miller:

Notes from Mark Miller:
Out of curiosity, would you be amenable to having 'diff' support added to Conduit?
Given two conduit nodes, return a bool if they are the same (by value) or different
Given two conduit nodes, return a conduit node that represents their difference
Given two conduit nodes, return a string (or send strings to a caller-specified stream) of the difference
With knobs to control exclusive/inclusive diffs, absolute/relative diff tolerances, numerical diff algorithm (some examples are in silo)?

=====

Node n;
Node other;
Node res;
n.compare(other,res)
res could contain a node that has the following sub nodes:
difference - parts of the tree that are in both "n" and "other", with leaves diffed 
missing - parts of the tree that are in "other" but not in "n"
extra - parts of the tree that are in "n" but not in "other"
Tricky things:
String diffs - how do we present them (maybe just say equal or not, keep it numeric with a lexical distance?)
Lists, with varying elements between n and other. Shouldn't be too bad, but requires some thought.
or we could go simpler, and just compare the things that are both in "n" and "other"

xjrc:
@cyrush: I'd like to tackle implementing this feature next week so that I can use it in another application. Let me know if you have any extra ideas related to this feature and we can discuss a game plan. Thanks!

cyrush:
@xjrc awesome! feel free to ping me with questions and we can iterate in a PR.

cyrush:
resolved in #246

Issue #98, opened 2016-11-30
consider void_array


cyrush:
A void_array (a DataArray<void*> instance) would be useful for the case were you want to generically stride though and access the bytes for any array w/o type checking. 

The void_array will still have access to the full data type spec, so the parts of the DataArray inteface related to set, to_json, etc helpers could still work fine, but we need to look if it has implications on plumbing of the template use.

cyrush:
It seems the DataArray::set methods would need some special logic for the void * case

cyrush:
We can effectively access and iterate over the data directly from a node via element_ptr(). 

The main benefit of this comes in the a case where you have a leaf Node that you don't want to access with a specific bw style type, but you have some data values you want to set:

```
void_array arr = n.value();
arr.set(...) // pass in ptrs to btw style types, any other data array, etc
```

That said, the bracket operator overloads don't really make sense in this context. We could try some sort of magic that would mirror Node::Value, but that seems extreme. 

As such "void_array" doesn't really adhere to the DataArray interface -- so we should think of a different way to provide what is needed.

PR #246, opened 2017-12-20
add node comparison operations

xjrc:
The changes in this pull request add node comparison operations to Conduit as described in #136.

- [x] Implement a functional version of `conduit::Node::compare`
- [x] Refine the informational strings returned in the `info` input parameter to the `conduit::Node::compare` function.
- [x] Add test cases for the `conduit::Node::compare` function.


src/libs/conduit/conduit_node.cpp
```diff
@@ -13501,6 +13502,146 @@ Node::find_first_data_ptr() const
...
+            const conduit::Node m_elem(elem_dtype, (void*)element_ptr(i), true);
+            const conduit::Node n_elem(elem_dtype, (void*)n.element_ptr(i), true);
+            // TODO(JRC): Figure out how to make this work to enable better
+            // element-by-element comparisons.
+            if(m_elem.value() != n_elem.value())
+            {
+                info.set("Data value mismatch.");
+                res = false;
+            }
+        }
```
cyrush: this is the hardest key to the puzzle, `m_elem.value() != n_elem.value() `won't work b/c each `.value()` requires a constraining type to work. 

One solution I thought  for these type of general problems was to have a `void_array`, but for various reasons I am evolving to think this is not a good idea. (https://github.com/LLNL/conduit/issues/98)

src/libs/conduit/conduit_node.cpp
```diff
@@ -13501,6 +13502,146 @@ Node::find_first_data_ptr() const
...
+    {
+        info.set("Data type incompatibility.");
+        res = false;
+    }
+    else
+    {
+        // TODO(JRC): Determine how to customize this function so that users that
+        // want the stride/endianness to be the same can still perform a proper query.
+        DataType elem_dtype(dtype().id(), 1, 0, 0, dtype().element_bytes(), dtype().endianness());
+        for(size_t i = 0; i < (size_t)dtype().number_of_elements(); i++)
```
cyrush: element_ptr takes care of the stride
 
I don't think we need to worry about endianness. b/c if endianness is different any actual value comparison will fail (there isn't magical endianness conversion when accessing values, users must do it explicitly).

Some thoughts:

The most useful outcome would be a list of what is different between the two arrays, we may want a bytes vs numeric option. 


Can we just diff the bytes and then for output drop into special cases for numeric differences (dtype().is_number()?).

Should we store the all differences, or store the differences compactly (element 2 differs by -.003343, element 123123 differs by .333, omitting those that are the same)

One possibility: 
At this point given that `(dtype().compatible(n.dtype())  && n.dtype().compatible(dtype()))` We know they are the same type. We can create logic for all bitwidth types like:
```
if (this->dtype().is_int8()
{
 int8_array a = this->value();
 int8_array b = n->value();
}
```

And then push the difference operation into the the DataArray class. 
a.diff(b,out);

src/libs/conduit/conduit_node.cpp
```diff
@@ -13501,6 +13502,146 @@ Node::find_first_data_ptr() const
...
+        //   nodes between the comparison parents are filled in the list.
+        //   All other nodes in the list are empty.
+
+        index_t m_nchild = number_of_children();
+        index_t n_nchild = n.number_of_children();
+
+        size_t i = 0;
+        for(; i < (size_t)std::min(m_nchild, n_nchild); i++)
+        {
+            const Node &m_child = child(i);
```
cyrush: Small request: I usually limit  using the `m_` prefix to member vars, not local ones.  Would it be ok to change these?

xjrc: Point well taken.  Also, in retrospect, I don't think it's a great idea to differentiate variables using the `m` and `n` characters, so I'll definitely update this.

cyrush:
How about having `compare` give you a general compare info and having `diff` give you numeric results (for the leaves)?

Your current diff becomes:
```
bool DataArray<T>::compare(const DataArray<T> &array, Node &info) const 
```

and we add:
```
T DataArray<T>::diff(const DataArray<T> &array, Node &out) const 
```

For out, the creates an array of type T that holds the actual numeric difference, and returns the sum of all differences?

xjrc:
That seems like a reasonable suggestion to me, but I like the fact that the `Node::diff` and `DataArray<T>::diff` functions are currently consistent in their behavior (in particular, they both currently return a bool that indicates if there are differences and fill the given `out` node with the full digest).  Perhaps we could move your proposed behavior to a new function instead (e.g. `numeric_diff` or `ndiff`)?

src/libs/conduit/conduit_data_array.cpp
```diff
@@ -136,6 +138,96 @@ DataArray<T>::element(index_t idx) const
...
+    for(; i < (size_t)std::min(t_nelems, o_nelems); i++)
+    {
+        T &t_elem = (*this)[i];
+        T &o_elem = array[i];
+
+        T diff_elem = t_elem - o_elem;
+        if(diff_elem < (T)0.0)
+        {
+            diff_elem *= (T)-1.0;
+        }
```
cyrush: do we want abs, or simply the difference?

cyrush: also, what happens here for unsigned types?

src/libs/conduit/conduit_node.cpp
```diff
@@ -13501,6 +13505,200 @@ Node::find_first_data_ptr() const
     return res;
 }
 
+//---------------------------------------------------------------------------//
+bool
+Node::equals(const Node &n) const
+{
+    Node info;
+    return !diff(n, info);
```
cyrush: since the messages created by `diff` can include large strings when there are differences and we won't even to use all that info, we may want to use `ndiff`

cyrush:
still not sure on the names, to clarify my earlier comment:  I was thinking that Node, as well as DataArray would have `compare` and `diff` & they would be consistent. 

For Node: 

```
bool Node::compare()
double Node::diff()
```

compare would be the case that gives you info, diff would be the case that gives you the numeric differences (at the leaves) and the aggregate difference.

cyrush:
@xjrc have pondered a bit more, and consulted with @bryujin -- I'll provide some more input tomorrow.

cyrush:
Here are a few notes: 

Scrap my earlier idea for `double Node::diff()` the return sum won't be very meaningful for complex trees.

For diffs on the leaves, the most common data folks will want is the actual numeric difference, not a textual summary (they can derive a textual summary from the numeric difference)  

So what do you think about one method: `bool Node::diff()`, which provide numeric results for the leaves (modulo something sensible for char8_str case) and an overall bool result?

xjrc:
The single `bool Node::diff` function sounds reasonable; I'll re-evaluate the interface and trim off all of the unnecessary functions.

xjrc:
@cyrush: Assuming that the Windows build works, I believe my latest commit to include a final (or at least near-final) version of the implementation. Whenever you get a chance, please let me know if everything looks okay and if there are any other changes I need to make!

src/libs/conduit/conduit_data_array.cpp
```diff
@@ -160,69 +161,51 @@ template <typename T>
...
 
-    for(size_t i = 0; i < (size_t)std::min(t_nelems, o_nelems); i++)
+    DataType info_dtype(array.dtype().id(), std::max(t_nelems, o_nelems));
+    info.set(info_dtype);
+    T* info_ptr = (T*)info.data_ptr();
+
+    // TODO(JRC): This is a bit sloppy, but it feels like the only decent
+    // way to signal missing values in cases where the lengths are mismatched.
+    T t_fill, o_fill;
+    if(std::numeric_limits<T>::has_infinity)
```
xjrc: @cyrush: This is the biggest/most unusual change I can think of since some of the Travis jobs started failing. Do you think this is the culprit behind the failures, or do you suspect that the problem lies elsewhere? If we can't use the functionality in the "limits" header due to compatibility concerns, what do you think the best way to populate the extra entries in during a `DataArray<T>::diff` operation?

cyrush: The travis issues don't have anything to do with your changes.  Not sure what is going on in appveyor. Not sure about populating with infinity, sound idea, but might create headaches in practice.

src/libs/conduit/conduit_data_array.cpp
```diff
@@ -136,6 +139,87 @@ DataArray<T>::element(index_t idx) const
...
+
+//---------------------------------------------------------------------------//
+template <typename T> 
+bool
+DataArray<T>::diff(const DataArray<T> &array, Node &info, const Node &config) const 
+{ 
+    // TODO(JRC): Consider moving the default machine epsilon to a constant
+    // value somewhere more generally accessible within Conduit.
+    float64 epsilon = config.has_child("eps") ? config["eps"].as_float64() : 1e-12;
+
```
cyrush: I think this is handy, but maybe we can just pass as a float64 and avoid the node?

src/libs/conduit/conduit_data_array.cpp
cyrush: May be safer to pick one path to avoid surprises based on the compiler -- might have to be he non has_infinity path.

cyrush:
travis failure looks like a strange fpic related issue with python, I can try to get the to bottom of it -- I don't think it's related to this branch.

cyrush:
@bryujin 

Q: diffing two DataArrays' with different number of elements -what are your thoughts on values for elements that don't exist in both arrays? Also, any thoughts on how symmetry should apply, etc?

src/libs/conduit/conduit_data_array.cpp
xjrc: That's a good thought. I'd considered doing that originally, but I went with a `Node` to prevent tying the user-specified epsilon to any given type and in case there were more options in the future. I think though that it's best not to so predictive in this case as it's at the expense of usability (and the epsilon gets cast ultimately to a `float64` anyway), so I'm applying your suggestion in my next update.

cyrush:
@xjrc don't fight the windows stuff, I can help you cleanup windows issues soon.

cyrush:
@xjrc  windows failure is a linking issue with utils::epsilon. I'll have to think of how to handle a constant like this so it works on windows. It compiles and links if I turn epsilon into a function, but thats not ideal.

xjrc:
@cyrush: I like `mlarray` and agree that name overloading with `numpy` is generally unadvisable; I've changed the code to adopt this name for the concept instead.

xjrc:
@cyrush: Assuming that the CI tests all pass, I think this PR is very close to being merge-ready. At this stage, the last item that needs to be reviewed is the `info` formatting for `{Node|DataArray}::{diff|equals}`. I've included a simple example of the current formatting scheme below; let me know what you think and we can converge on a strategy whenever you're ready/available.

### Instance Node ###

```
{
  "common": 
  {
    "same": [1, 2, 3],
    "diff_len": [4, 5],
    "diff_val": [4, 5, 6]
  },
  "extra": 
  {
    "a": 1,
    "b": 2
  }
}

```

### Argument Node ###

```
{
  "common": 
  {
    "same": [1, 2, 3],
    "diff_len": [4, 5, 6],
    "diff_val": [6, 5, 4]
  },
  "missing": 
  {
    "a": "one",
    "b": "two"
  }
}
```

### `Node::equals` Result ###

```
[
  
  {
    "errors": 
    [
      "node::equals: arg missing child 'extra'",
      "node::equals: self missing child 'missing'"
    ],
    "valid": "false"
  },
  
  {
    "common": 
    [
      
      {
        "valid": "false"
      },
      
      {
        "same": 
        [
          
          {
            "info": 
            [
              "dataarray::equals: data length match (3)"
            ],
            "valid": "true"
          },
          [0, 0, 0]
        ],
        "diff_len": 
        [
          
          {
            "errors": 
            [
              "dataarray::equals: data length mismatch (2/3)"
            ],
            "valid": "false"
          },
          null
        ],
        "diff_val": 
        [
          
          {
            "info": 
            [
              "dataarray::equals: data length match (3)"
            ],
            "errors": 
            [
              "dataarray::equals: data item(s) mismatch; see diff below"
            ],
            "valid": "false"
          },
          [-2, 0, 2]
        ]
      }
    ]
  }
]

```

cyrush:
@xjrc thanks, lets sync up next week face to face and I think we will be able to quickly converge on the final info node contents

xjrc:
After discussing this with @cyrush, we've converged on a clearer format for digest nodes returned by the comparison functions. An example of this new digest format is given below:

### Instance Node ###

```
{
  "common": 
  {
    "same": [1, 2, 3],
    "diff_len": [4, 5],
    "diff_val": [4, 5, 6]
  },
  "extra": 
  {
    "a": 1,
    "b": 2
  }
}

```

### Argument Node ###

```
{
  "common": 
  {
    "same": [1, 2, 3],
    "diff_len": [4, 5, 6],
    "diff_val": [6, 5, 4]
  },
  "missing": 
  {
    "a": "one",
    "b": "two"
  }
}
```

### `Node::diff` Result ###

```
{
  "valid": "false",
  "children":
  {
    "diff":
    {
      "common":
      {
        "valid": "false",
        "children":
        {
          "diff":
          {
            "same":
            {
              "valid": "true",
              "value": [0, 0, 0],
            },
            "diff_len":
            {
              "valid": "false"
              "errors":
              [
                "dataarray::diff: data length mismatch (2/3)"
              ],
              "value": null,
            },
            "diff_val":
            {
              "valid": "false"
              "errors":
              [
                "dataarray::diff: data item(s) mismatch; see 'value' below"
              ],
              "value": [-2, 0, 2],
            },
          },
        },
      },
    },
    "missing":
    [
      "missing",
    ],
    "extra":
    [
      "extra",
    ],
  },
}
```

src/libs/conduit/conduit_data_array.cpp
```diff
@@ -229,93 +214,92 @@ DataArray<T>::diff(const DataArray<T> &array, Node &info, const float64 epsilon)
...
         if(dtype().is_char8_str())
         {
+            // TODO(JRC): Currently, due to the way that strings are represented
+            // in C/C++ (i.e. null-terminated), this comparison isn't very useful.
+            // Consider comparing "a" with "aa"; the first string has two elements
+            // (i.e. "a" and "\0"), so will come back true for this diff when
+            // compared against "aa" (whose first two elements are "a" and "a").
+            // This behavior should be updated if a reasonable and useful
+            // interpretation is discovered for string diffs.
             uint8 *t_compact_data = new uint8[dtype().bytes_compact()];
```
cyrush: i agree -- i don't think folks would expect this result, can we use `as_string()` to obtain the values and compare?
If it's not compact, we could call `compact_to()` and then us `as_string()`

cyrush:
@xjrc looks great!  I added the c, python and fortran diff methods and made a few small changes.  Will merge after looking at CI results.

PR #263, opened 2018-02-08
Bugfix: Node:load() needs to open its file in binary mode.

kennyweiss:
Otherwise, the file might not load all the requested bytes.

### DIFF
--- a/src/libs/conduit/conduit_node.cpp
+++ b/src/libs/conduit/conduit_node.cpp
@@ -323,7 +323,7 @@ Node::load(const std::string &stream_path,
 
     allocate(dsize);
     std::ifstream ifs;
-    ifs.open(stream_path.c_str());
+    ifs.open(stream_path.c_str(), std::ios_base::binary);
     if(!ifs.is_open())
         CONDUIT_ERROR("<Node::load> failed to open: " << stream_path);
     ifs.read((char *)m_data,dsize);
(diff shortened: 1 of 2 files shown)


kennyweiss:
The bug exhibited in a 32-bit build on windows with visual studio.

I can add a simple regression test, if this is helpful. It was a ``Node`` with and array of 20 ``int64`` values, with values ``arr[i] = 2 * i``.  For some reason, it only loaded the first 12 values.

cyrush:
@kennyweiss, thanks for resolving this! A test would be great.

I resolved something like this before and it was very hard to track down. 

Do you think there are other places an issue like this may pop up?

cyrush:
For context, I solved this for write & should have solved for read as well but missed it (again thanks!)
https://github.com/LLNL/conduit/commit/618e722e63cedb7f5f59c9ddfebd8efe16501821

kennyweiss:
@cyrush -- I added the regression test, and confirmed the bug when ``ifstream`` is initialized without the binary flag in 32-bit windows configurations.

I also changed the flag from ``ifstream::binary`` to ``ios_base::binary`` for consistency with https://github.com/LLNL/conduit/commit/618e722e63cedb7f5f59c9ddfebd8efe16501821

kennyweiss:
Looks like the travis failure was due to the build taking too long:

https://travis-ci.org/LLNL/conduit/jobs/339174077#L2269-L2276

Suggestions?

cyrush:
@kennyweiss since I had to switch to spack for tpls in travis, timeouts are an unfortunate reality. 
I'll keep at running the jobs until we get a pass.

cyrush:
@kennyweiss everything passed -- thanks again!

PR #272, opened 2018-03-07
Bugfix: Node::diff fails for string data with offsets

xjrc:
The changes in this pull request fix a bug in the `Node::diff` implementation that caused it to fail when comparing two nodes that had string data with non-zero offset values. This pull request also amends the `Node::diff` test cases to validate that these fixes work properly.

### DIFF
--- a/src/libs/conduit/conduit_node.cpp
+++ b/src/libs/conduit/conduit_node.cpp
@@ -13680,8 +13680,8 @@ Node::diff(const Node &n, Node &info, const float64 epsilon) const
         {
             // NOTE: Can't use 'value' for characters since type aliasing can
             // confuse the 'char' type on various platforms.
-            char_array t_array((const void*)as_char8_str(), dtype());
-            char_array n_array((const void*)n.as_char8_str(), n.dtype());
+            char_array t_array((const void*)m_data, dtype());
+            char_array n_array((const void*)n.m_data, n.dtype());
             res |= t_array.diff(n_array, info, epsilon);
         }
         else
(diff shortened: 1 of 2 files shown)


src/libs/conduit/conduit_node.cpp
```diff
@@ -13830,8 +13830,8 @@ Node::diff_compatible(const Node &n, Node &info, const float64 epsilon) const
         {
             // NOTE: Can't use 'value' for characters since type aliasing can
             // confuse the 'char' type on various platforms.
-            char_array t_array((const void*)as_char8_str(), dtype());
-            char_array n_array((const void*)n.as_char8_str(), n.dtype());
+            char_array t_array((const void*)m_data, dtype());
+            char_array n_array((const void*)n.m_data, n.dtype());
             res |= t_array.diff_compatible(n_array, info, epsilon);
```
cyrush: got it: 
as_char8_str() already applies the offset, by then using the dtype again ends up double applying.

src/tests/conduit/t_conduit_node_compare.cpp
```diff
@@ -191,31 +194,39 @@ TEST(conduit_node_compare, compare_leaf_numeric)
...
 
         for(index_t ci = 0; ci < 4; ci++)
         {
-            std::string comp_str = compare_strs[ci];
-            std::string diff_str = comp_str;
-            {
-                size_t test_len = comp_str.length();
-                diff_str[test_len-1] += 1;
-            }
+            index_t leaf_tid = DataType::CHAR8_STR_ID;
```
cyrush: To help with future debugging -- can you add a comment describing how the offset is being used. Maybe even show the pattern of the data you are comparing?

"heM" == "heM" (or whatever it may be )

xjrc: I've added this with my latest commit.

xjrc:
I'd really like this fix for a dependent project. Despite being kludgey, would it be out of the question to move the release tag v0.3.1 to include this fix? If not I should be able to get by, but sliding the release would make things a bit easier for me.

cyrush:
Can't change the tag b/c folks are already consuming it, but if we gather enough changes we can do a 0.3.2.

PR #288, opened 2018-05-17
make small improvements to array conversion test cases

xjrc:
The changes in this pull request make a small set of improvements to the `to_[type]_array` conversion test cases.


src/tests/conduit/t_conduit_node_convert.cpp
```diff
@@ -55,84 +55,145 @@
...
-    nconv.print();
-
-    // float native c
-    n.to_float_array(nconv);
-    nconv.print();
-    
-    n.to_double_array(nconv);
-    nconv.print();
-
-
```
cyrush: I actually prefer the old way here.  Its much more to type, but its a lot clearer what is being tested

src/libs/conduit/conduit_node.cpp
```diff
@@ -9513,6 +9513,289 @@ Node::to_double_array(Node &res) const
...
+//---------------------------------------------------------------------------//
+/// zero copy if possible array to other array types
+//---------------------------------------------------------------------------//
+
+//---------------------------------------------------------------------------//
+void
+Node::zcip_int8_array(Node &res) const
+{
+    if(!dtype().is_int8())
+    {
```
cyrush: I think you want the compact check as well for your use case. 

I think that gets us further away from this being able to converge.  A sign we may be trying to do too much with this magic.

src/tests/conduit/t_conduit_node_convert.cpp
cyrush: Also The new form also doesn't save us that much typing b/c of all the c-style functions you had to provide ...

xjrc: I'll concede that the old version was more clear, but it was also only testing the bare minimum in terms of functionality (i.e. that the `to_[type]_array` functions all ran without throwing errors for simple data) and extending it to perform any additional validation tests would make it much longer than the new code.

src/libs/conduit/conduit_node.cpp
xjrc: I think I need to compact the data beforehand regardless in my case, but I see your point. Excluding the compact check here means that there is a small bit of magic insofar as `zcip_[type]_array` will always compact the data if a data type transform occurs, which is far from ideal.

Looking over my client code again, this "zero copy if possible" business is more endemic to the Conduit interface than I'd realized. In particular, I want to have this option in all cases where I need the data to be of a particular type/layout. Somehow, I want to tell Conduit "this is what I want the data to look like (e.g. contiguous integers); please make all the transforms necessary to get it into this layout while using as little extra memory as possible". That said, I don't think this `zcip_[type]_array` business is the answer; I'm going to go back to the drawing board on this one and see if I can come up with something more general.

xjrc:
I'm going to scrap most of this PR, but I'd like to keep the `to_[type]_array` test change because I think it's useful. I'll update this shortly.

src/tests/conduit/t_conduit_node_convert.cpp
```diff
@@ -55,84 +55,72 @@
...
+        Node type_node;
+        convert_fun(data_node, type_node);
+
+        EXPECT_EQ(type_node.dtype().id(), node_tid);
+        EXPECT_EQ(type_node.dtype().number_of_elements(),
+            data_node.dtype().number_of_elements());
+
+        Node temp_node;
+        convert_to_uint8_array(type_node, temp_node);
+        EXPECT_FALSE(memcmp(temp_node.data_ptr(), data_node.data_ptr(),
```
cyrush: I am confused b/c I feel like there should be one case where its basically uint8 to uint8, and this should pass?

PR #289, opened 2018-05-17
mesh blueprint: add implicit point topology docs and example

### LINKED ISSUE #287: mesh bp: allow implicit points topology
make it easy to define a topology that uses all points in a coordinate set.

cyrush:
To resolve #287

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -1285,7 +1285,11 @@ mesh::topology::verify(const Node &topo,
     {
         const std::string topo_type = topo["type"].as_string();
 
-        if(topo_type == "uniform")
+        if(topo_type == "points")
+        {
+            res &= mesh::topology::points::verify(topo,info);
+        }
+        else if(topo_type == "uniform")
         {
             res &= mesh::topology::uniform::verify(topo,info);
         }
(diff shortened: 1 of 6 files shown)


cyrush:
docker build failed due to spack not picking up gfortran, I'll try to jump down the rabbit hole and resolve soon ...

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -77,9 +80,15 @@ namespace conduit { namespace blueprint { namespace mesh {
...
         coord_system_list + sizeof(coord_system_list) / sizeof(coord_system_list[0]));
 
-    static const std::string topo_type_list[4] = {"uniform", "rectilinear", "structured", "unstructured"};
+    static const std::string topo_type_list[5] = {"points",
+                                                  "uniform",
+                                                  "rectilinear",
+                                                  "structured",
+                                                  "unstructured"};
+
     static const std::vector<std::string> topo_types(topo_type_list,
```
cyrush: @xjrc I think I tackled the only change needed for verify. 

Verify is being tested as part of the blueprint mesh examples tests -- if this wasn't working I would expect a failure.

xjrc: This change is certainly a big part of making `conduit::blueprint::verify` work, but there are a few other minor changes that should be included. In particular, the test cases should be extended to test this case explicitly (see [`src/tests/blueprint/t_blueprint_mesh_verify.cpp:637`](https://github.com/LLNL/conduit/blob/master/src/tests/blueprint/t_blueprint_mesh_verify.cpp#L637) for details). Also, it may be worth adding a `conduit::blueprint::mesh::topology::points::verify` function as well to go along with all the other per-type verify functions.

cyrush: perfect, thanks for keeping all of this in mind!

cyrush:
The issue seems to be with gfortran 7, may be related to :
https://github.com/spack/spack/pull/7017

that said, I believe we are using a spack after this was merged. So I may just need to run the command to find the compilers in the docker container or use an older version of ubunutu. 

Regardless -- this error is not a blocker for this PR.

cyrush:
Looks good, thanks @xjrc !

Issue #293, opened 2018-05-30
make conduit's bitwidth style integer types match c++11 standard integer types


cyrush:
or bridge extra types when necessary to support c++1 standard types.

Add tests that exercise conversions, etc

I think the primary difference is the preference of `long  `as `int64`, instead of `int`, but I will have to check a few compilers and platforms to see whats feasible

cyrush:
resolved in #301

PR #301, opened 2018-06-12
ensure mapping c types to bw types is compatible with c++11 std bw types

cyrush:
This changes the bitwidth style type mapping to select `long long` before `long` when necessary to match c++11's std bitwidth mappings. 

This also changes the default C++ language standard used to c++11. 

When on, there are static asserts that check that the conduit integer bitwidth style types match the c++11 bitwidth style types. 

c++98 is still supported - via (`BLT_CXX_STD="c++98"`).


cyrush:
this will resolve #293

cyrush:
after testing with several compilers, it's clear there isn't an implicit precedence ordering  for how c++11 compilers map the types to bit-width style types. So we have to use a c++11 compiler to inform our precedence. I added code for this -- of course this is only possible when c++11 is enabled, so we can't guarantee that conduit_int64 will equal std::int64 unless c++11 is used.

PR #302, opened 2018-06-13
Use PROJECT_SOURCE_DIR for RapidJSON include

davidbeckingsale:
This allows conduit to be included as a submodule in another CMake project.


davidbeckingsale:
@cyrush - any comments/questions?

Issue #297, opened 2018-06-05
make dedicated blueprint examples for new users


xjrc:
There should be a dedicated set of Blueprint examples targeted at acquainting new users to the format. These examples should be tested like all of the other existing examples while facilitating trivial comparison between a Blueprint node and what it represents. This task can be accomplished by enhancing the [`t_conduit_docs_blueprint_examples.cpp`](https://github.com/LLNL/conduit/blob/master/src/tests/docs/t_conduit_docs_blueprint_examples.cpp) file and its associated documentation.

wnissen:
I would love this, I was just looking at the braid example and wishing it was in Python. I think it would easily fit in a single file.

xjrc:
These new examples have been added with the merging of #311.

PR #311, opened 2018-07-05
add dedicated blueprint examples for new users

xjrc:
The changes in this pull request add a suite of new examples that allow new users to easily compare different types of schema to the visualizations of the meshes they represent. This suite of examples is encapsulated by the `conduit::blueprint::mesh::examples::basic` method, which is a simplified version of the `braid` example that has only one field that contains the global ID for each mesh element. This pull request directly addresses issue #297.


src/docs/sphinx/blueprint_mesh.rst
```diff
@@ -392,9 +392,228 @@ To conform, the ``state`` entry must be an *Object* and can have the following o
...
+
+* **Call**
+
+.. literalinclude:: ../../tests/docs/t_conduit_docs_blueprint_demos.cpp
+   :lines: 132-137
+   :language: cpp
+   :dedent: 4
+
+* **Schema**
+
```
cyrush: These aren't really Schemas, it's actually the data tree.  Can we just say `Result` instead of Schema ?

src/docs/sphinx/blueprint_mesh.rst
```diff
@@ -392,9 +392,228 @@ To conform, the ``state`` entry must be an *Object* and can have the following o
...
++---------------+--------------------+-------------------+-------------------+------------------+
+
+The remainder of this section demonstrates each of the different ``gradient()`` mesh types, outlining
+each type with a simple example that (1) presents the generating call, (2) shows the results of the
+call in Blueprint schema form, and (3) displays the corresponding graphical rendering of this schema.
+
+Uniform
+====================================
+
+* **Call**
```
cyrush: `Usage Example` instead of `Call`?

src/docs/sphinx/blueprint_mesh.rst
```diff
@@ -392,9 +392,228 @@ To conform, the ``state`` entry must be an *Object* and can have the following o
...
 ~~~~~~~~~~~~~~~~~~~~~
 
-In C++, the ``conduit::blueprint::mesh::examples`` namespace and in Python the `conduit.blueprint.mesh.examples` module provide
-functions that generate example mesh blueprint data. For details on how to write these data sets to files, see the unit tests that exercise these examples in ``src/tests/blueprint/t_blueprint_mesh_examples.cpp``. This section outlines ``braid``, ``sprial``, and ``julia`` examples.
+The C++ ``conduit::blueprint::mesh::examples`` namespace and the Python `conduit.blueprint.mesh.examples` module provide
+functions that generate example mesh blueprint data. For details on how to write these data sets to files, see the unit
+tests that exercise these examples in ``src/tests/blueprint/t_blueprint_mesh_examples.cpp``. This section outlines the
+examples that demonstrate the most commonly used mesh schemas.
 
+gradient
```
cyrush: Thinking `gradient` might confuse folks (the dataset doesn't really contain a spatial gradient), should we just call it `basic`?

xjrc: Oops; I was considering doing a spatial gradient of the form `x+y+z` originally, but I changed it to IDs without modifying the name. I'll update this now!

src/libs/relay/conduit_relay_io_blueprint.cpp
```diff
@@ -0,0 +1,243 @@
...
+// -- begin conduit::relay --
+//-----------------------------------------------------------------------------
+namespace relay
+{
+
+//-----------------------------------------------------------------------------
+// -- begin conduit::relay::io_blueprint --
+//-----------------------------------------------------------------------------
+namespace io_blueprint
+{
```
cyrush: header seems to say `blueprintio`, and this file `io_blueprint`?

I woudl actually go with a nested name space:
`io::blueprint`

xjrc: @cyrush: I actually had that as the name originally, but I changed the namespace inclusion to an underscore to prevent potential ambiguity when `using namespace` directives get involved in the mix. Do you still prefer the nested namespace idea despite this potential problem? If so, I don't mind changing it to suit your suggestion.

xjrc: @cyrus: After giving the rename a go, I really don't think it's a good idea because the potential ambiguities make writing the Blueprint/Relay code nontrivially more involved. For example, just trying the rename on its own resulted in the following errors:
```
/path/to/conduit/src/libs/relay/conduit_relay_io_blueprint.cpp: In function 'void conduit::relay::io::blueprint::save(const conduit::Node&, const string&, const string&)':
/path/to/conduit/src/libs/relay/conduit_relay_io_blueprint.cpp:148:20: error: 'conduit::relay::io::blueprint::mesh' has not been declared
     if(!blueprint::mesh::verify(mesh, info))
                    ^
/path/to/conduit/src/libs/relay/conduit_relay_io_blueprint.cpp:158:19: error: 'conduit::relay::io::blueprint::mesh' has not been declared
     if(blueprint::mesh::is_multi_domain(mesh))
                   ^
/path/to/conduit/src/libs/relay/conduit_relay_io_blueprint.cpp:184:33: error: 'conduit::relay::io::blueprint::mesh' has not been declared
                     !blueprint::mesh::topology::unstructured::verify(topo, info) ||
                                 ^
/path/to/conduit/src/libs/relay/conduit_relay_io_blueprint.cpp:190:28: error: 'conduit::relay::io::blueprint::mesh' has not been declared
                 blueprint::mesh::generate_index(
```
With this in mind, I'm going to keep it as `io_blueprint` for the time being.

cyrush: gotcha, that sounds like a good plan for now.

I have some thoughts about how to better centralize some of the mesh related relay read and write stuff that 
we can pursue after this PR. 

Working on getting ADIOS support in, and with that will have a mpi enabled relay io library.

That library will be a good place to centralize similar logic in VisIt, Ascent, and Conduit.

src/libs/relay/conduit_relay_io_blueprint.cpp
```diff
@@ -0,0 +1,219 @@
...
+//-----------------------------------------------------------------------------
+// -- begin conduit::relay::io --
+//-----------------------------------------------------------------------------
+namespace io_blueprint
+{
+
+//---------------------------------------------------------------------------//
+std::string
+identify_protocol(const std::string &path)
+{
```
cyrush: ran into the fact that this needs to be pulled for use in the ADIOS stuff as well, so ill solve that in the future so we dont have duplicated code

src/libs/blueprint/conduit_blueprint_mesh_examples.hpp
```diff
@@ -81,17 +81,23 @@ namespace mesh
...
                                      conduit::index_t nz,
                                      conduit::Node &res);
 
-    void CONDUIT_BLUEPRINT_API misc(const std::string &mesh_type,
-                                    conduit::index_t nx,
-                                    conduit::index_t ny,
-                                    conduit::index_t nz,
-                                    conduit::Node &res);
+    /// Generates a braid-like example mesh that covers elements defined in a
+    /// rectilinear grid. The sub-element type (e.g. triangles, quads,
```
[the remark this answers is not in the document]
xjrc: Good call; this has been fixed with the latest change.

cyrush:
@xjrc thanks, feel free to merge!

Issue #344, opened 2018-11-08
this logic isn't quite right


cyrush:
if(mesh::verify_multi_domain(n, info))
    {
        res = true;
    }
    else
    {
        info.reset();
        res = mesh::verify_single_domain(n, info);
    }

cyrush:
resolved in #350

PR #350, opened 2018-12-04
bug fix for mesh blueprint verify 

cyrush:
change mesh bp verify cascade to avoid losing info about 
multi domain failures


cyrush:
Will resolve #344

cyrush:
@xjrc sanity check my changes when you get a chance, thanks!

PR #373, opened 2019-01-18
compiler warnings pass

cyrush:


cyrush:
silo spack package has problem with system zlib

PR #374, opened 2019-01-18
verify for structured options for adjsets

nselliott:


cyrush:
@nselliott we are getting close on this one - I exported this to VisIt and noticed gaps between the domains: 

Screen Shot 2019-05-06 at 10 04 49 AM

Maybe a simple indexing fix is needed?

gzagaris:
I've seen this before. 

Perhaps, it is related to node-based partitioning vs. cell-based partitioning. Some codes may partition nodes (instead of cells) to different domains. Thereby, there exists a cell that spans two different domains, which, creates those gaps when visualized.

cyrush:
@gzagaris yes, I have seen things like this before as well.

this is an example mesh, i think this is just a simple index issue.  Right now the set of domains aren't presenting topology covering those gaps.

nselliott:
@cyrush That should be fixed now--I had the wrong size for the uniform coordsets.

nselliott:
@cyrush This PR is updated to include refinement ratios in structured adjsets, to identify differing resolutions between neighboring domains.

cyrush:
@nselliott took this for a spin, looks good to me.
I am going to merge and add some code to save the example in a follow on PR.

PR #382, opened 2019-02-26
Add ENABLE_EXAMPLES option to make them optional

davidbeckingsale:


cyrush:
With the exception of the staging, which is a special case that mike isn't building -- src/examples aren't actually compiled, only installed (aka copied over to the install dir). I think we want to do that in all cases - so I don't think we need this option yet.  Mike may have been thinking about Ascent, but it does already have an examples guard.

davidbeckingsale:
That's fine with me. Can you shoot Mike an email?

cyrush:
@davidbeckingsale I pinged mike

PR #405, opened 2019-04-19
fix issues w/ non-standard bit widths in blueprint transforms

xjrc:
The changes in this pull request fix a host of issues in the `conduit::blueprint::mesh::{coordset|topology}::to_*` functions associated with processing inputs containing non-standard bit width types. These bugs would cause these functions to crash when users attempted to pass mesh data with anything other than `{u}int32` and `float64` type data. With these fixes in place, mesh data of any viable type can be passed through these transformation interfaces.


xjrc:
@cyrush: The test cases that validate the correctness of these fixes have been added, so this pull request is ready for review. I've verified that these test cases only work with the inclusion of the implementation code in this PR. Thanks and please let me know if you have any comments or questions.

PR #411, opened 2019-05-08
add builtin yaml parsing

cyrush:


cyrush:
for windows static builds we either need the following fix: https://github.com/LLNL/blt/issues/268

Or we need to make libyaml a cpp lib. I will likely make it a cpp lib, it's being consumed as an object lib by a cpp lib.

cyrush:
LLNL/blt#268 is fixed in latest blt

Issue #398, opened 2019-03-08
place holder for potential relay_mpi issue


cyrush:
place holder for issue @mclarsen encountered, where broadcast of a child node went off the rails

cyrush:
resolved in #419

PR #419, opened 2019-06-15
fix crash in relay mpi bcast with non empty recv nodes

cyrush:
Schema::compatible() doesn't give is the green light to zero copy complex types, so we need more checks to avoid crashes.


cyrush:
resolves #419 and I believe #398

PR #430, opened 2019-06-28
add zfp relay features and zfparray blueprint

cyrush:


src/libs/blueprint/CMakeLists.txt
```diff
@@ -112,14 +112,27 @@ if(FORTRAN_FOUND)
...
+
+###################
+# Add zfp interface
+# if zfp was found
+###################
+
+if(ZFP_FOUND)
+  list(APPEND blueprint_headers "conduit_blueprint_zfparray.hpp")
+  list(APPEND blueprint_sources "conduit_blueprint_zfparray.cpp")
+  list(APPEND conduit_blueprint_deps zfp)
```
cyrush: We want to keep blueprint itself dependency free.  
I see zfp is included in the cpp-source, but do we actually need it here? 
I don't think its being used

cyrush:
@salasoom  This is looking great!  I was able to run everything on macOS, see below for a few changes that were necessary. 

After #429 is merged, our CI will start building zfp.

I still want to review a bit more but I think this is in good shape to merge soon!


```
diff --git a/src/cmake/thirdparty/SetupZfp.cmake b/src/cmake/thirdparty/SetupZfp.cmake
index ee40246d..d1b156a7 100644
--- a/src/cmake/thirdparty/SetupZfp.cmake
+++ b/src/cmake/thirdparty/SetupZfp.cmake
@@ -64,7 +64,7 @@ find_path(ZFP_INCLUDE_DIR zfp.h
           NO_CMAKE_SYSTEM_PATH)
 
 find_library(ZFP_LIB NAMES zfp
-             PATHS ${ZFP_DIR}/lib64
+             PATHS ${ZFP_DIR}/lib64 ${ZFP_DIR}/lib
              NO_DEFAULT_PATH
              NO_CMAKE_ENVIRONMENT_PATH
              NO_CMAKE_PATH
diff --git a/src/tests/blueprint/t_blueprint_zfp.cpp b/src/tests/blueprint/t_blueprint_zfp.cpp
index 4eba9e62..5d9923b4 100644
--- a/src/tests/blueprint/t_blueprint_zfp.cpp
+++ b/src/tests/blueprint/t_blueprint_zfp.cpp
@@ -161,7 +161,7 @@ TEST(conduit_blueprint_zfp, zfp_verify_invalid_zfparray_with_compressed_data_dty
         case 32:
         case 16:
             {
-                uint8 data = 3;
+                conduit::uint8 data = 3;
                 result[blueprint::zfparray::ZFP_COMPRESSED_DATA_FIELD_NAME] = data;
                 EXPECT_FALSE(blueprint::zfparray::verify(result, info));
             }
@@ -169,7 +169,7 @@ TEST(conduit_blueprint_zfp, zfp_verify_invalid_zfparray_with_compressed_data_dty
 
         case 8:
             {
-                uint64 data = 3;
+                conduit::uint64 data = 3;
                 result[blueprint::zfparray::ZFP_COMPRESSED_DATA_FIELD_NAME] = data;
                 EXPECT_FALSE(blueprint::zfparray::verify(result, info));
             }
diff --git a/src/tests/relay/t_relay_zfp.cpp b/src/tests/relay/t_relay_zfp.cpp
index 5166145b..c49770c8 100644
--- a/src/tests/relay/t_relay_zfp.cpp
+++ b/src/tests/relay/t_relay_zfp.cpp
@@ -98,8 +98,8 @@ TEST(conduit_relay_zfp, zfp_read_and_verify_compressed_data)
     // create compressed-array
     uint nx = 9;
     uint ny = 12;
-
-    float vals[nx * ny] = {0};
+    uint ntotal =  nx * ny;
+    float * vals = new float[ntotal];
     uint i;
     for (i = 0; i < nx*ny; i++) {
         vals[i] = i*i;
@@ -157,6 +157,8 @@ TEST(conduit_relay_zfp, zfp_read_and_verify_compressed_data)
             break;
     }
 
+    delete [] vals;
+
     EXPECT_TRUE(0 == std::memcmp(arr.compressed_data(), n_data.data_ptr(), arr.compressed_size()));
 }
```

cyrush:
@salasoom I merged in the zfp spack changes to master.  If you can update your branch above with those few suggested changes, I think we are close to ready to merge.

salasoom:
@cyrush I made the changes you suggested, but now there seem to be Spack issues on Travis CI that weren't there yesterday

src/libs/relay/conduit_relay_zfp.hpp
```diff
@@ -0,0 +1,108 @@
...
+//-----------------------------------------------------------------------------
+// -- begin conduit::relay::io --
+//-----------------------------------------------------------------------------
+namespace io
+{
+
+static const std::string ZFP_HEADER_FIELD_NAME = "zfp_header";
+static const std::string ZFP_COMPRESSED_DATA_FIELD_NAME = "zfp_compressed_data";
+
+zfp::array* CONDUIT_RELAY_API zfp_write(const Node &node);
```
cyrush: would `wrap` and `unwrap` be better names for these methods?

cyrush:
@xjrc  we are adding a new blueprint for a `zfparray`.  It all seems pretty straight forward to me, but can you take a quick look at the blueprint logic  (verify, etc) with another set of eyes and see if you have any suggestions?

src/libs/relay/conduit_relay_zfp.hpp
salasoom: We will be adding support for storing data compressed with zfp's C API (zfparray C++ objects currently require fixed-rate mode). So how about `wrap_zfparray` and `unwrap_zfparray`? Those seem a bit more descriptive about what is being passed in/out.

cyrush: Those sound good!

src/libs/blueprint/conduit_blueprint.hpp
```diff
@@ -64,6 +64,7 @@
 #include "conduit_blueprint_mcarray.hpp"
 #include "conduit_blueprint_mcarray_examples.hpp"
 
+#include "conduit_blueprint_zfparray.hpp"
```
salasoom: Looking at the latest CI build failure, this needs to be removed or conditionally included. If users build with ZFP and wish to use the blueprint zfp support, should they be expected to explicitly include this header, or have it automatically included with this file (conduit_blueprint.hpp)?

src/libs/relay/conduit_relay_zfp.hpp
```diff
@@ -0,0 +1,108 @@
...
+namespace relay
+{
+
+//-----------------------------------------------------------------------------
+// -- begin conduit::relay::io --
+//-----------------------------------------------------------------------------
+namespace io
+{
+
+static const std::string ZFP_HEADER_FIELD_NAME = "zfp_header";
```
salasoom: My other concern is that these const strings are defined twice: once here in relay, and once in blueprint's zfp header file. Any way to have one set of definitions shared between the two projects?

src/libs/blueprint/conduit_blueprint.hpp
cyrush: i think its fine to conditionally include it here.

That reminds we that we should also add it to this list:
https://github.com/LLNL/conduit/blob/master/src/libs/blueprint/conduit_blueprint.cpp#L94

Which can help folks query want is enabled at runtime.

src/libs/relay/conduit_relay_zfp.hpp
cyrush: If you want - it's ok for relay to depend on the blueprint lib.

That said, I think its a pretty low risk to have these defined twice.

salasoom: I agree it's low risk - and the tests will fail if they are mismatched.
Will leave as-is

cyrush:
thinking about it a bit more, zfp blueprint can always be there since there is no dep on zfp proper?
There may be rare cases where we want to identify zfp data but not actually touch it. I think you can move to always support the zfp blueprint.

salasoom:
Yeah I just reached that same conclusion. In blueprint, we check for named nodes and underlying dtype, but nothing from zfp itself is needed.

I'll remove conditionals to always have those zfp files included in blueprint.

src/libs/blueprint/conduit_blueprint_zfparray.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+        if(!compressed_data.dtype().is_unsigned_integer()) {
+            log::error(info, proto_name, "ZFP compressed-data node's dtype is incompatible with the compiled ZFP bitstream word size");
+            res = false;
+        }
+    }
+
+    log::validation(info, res);
+
+    return res;
```
xjrc: Looks perfect; thanks for adding this, @salasoom!

src/libs/relay/conduit_relay_zfp.cpp
```diff
@@ -0,0 +1,183 @@
...
+        default:
+            is_readable = false;
+            break;
+    }
+
+    if(!is_readable) {
+        return NULL;
+    }
+
+    zfp::array::header header[1];
```
xjrc: Forgive me if I'm missing something obvious, by why are you using an array containing a single item here instead of declaring a standalone object (i.e. `zfp::array::header` instead of `zfp::array::header[1]`)?

src/libs/relay/conduit_relay_zfp.cpp
```diff
@@ -0,0 +1,183 @@
...
+int
+wrap_zfparray(const zfp::array *arr,
+         Node &dest)
+{
+    // store header
+    zfp::array::header header;
+    try {
+        header = arr->get_header();
+    } catch(zfp::array::header::exception const &) {
+        return 1;
```
xjrc: Since we're returning an integer to indicate success status for this function, it would probably be better to use an integer other than 1 here to differentiate this case from the bitwidth mismatch case accounted for below.

src/tests/relay/t_relay_zfp.cpp
```diff
@@ -0,0 +1,300 @@
...
+        default:
+            FAIL() << "ZFP was compiled with an unrecognizable word type";
+    }
+
+    // fetch zfparray object from Node
+    zfp::array* fetched_arr = io::unwrap_zfparray(result);
+
+    // verify no instance returned
+    ASSERT_TRUE(fetched_arr == 0);
+}
```
xjrc: Great test coverage! Thank you for being so thorough!

src/libs/relay/conduit_relay_zfp.cpp
salasoom: I'll change it to remove the array-part

src/libs/relay/conduit_relay_zfp.cpp
salasoom: Sure I'll change the return value so we can differentiate error causes

salasoom:
@cyrush I made some changes that @xjrc suggested, and fixed the CI issues so everything is green

cyrush:
@salasoom thanks!

PR #431, opened 2019-07-01
add specsets validation and testing

xjrc:
The changes in this pull request add the `specsets` section as an officially supported section in the Blueprint schema. This includes all of the following added features:

- [x] Added documentation to outline the new `specsets` section and its role within Blueprint.
- [x] Added a verification method for `specsets` to the `conduit::bleuprint::mesh` namespace.
- [x] Added index support for the `specsets` section (both generation and verification).
- [x] Added regression tests for all of the functions implemented as part of adding `specsets` to Blueprint.


src/libs/blueprint/conduit_blueprint_mcarray.cpp
```diff
@@ -346,6 +346,18 @@ verify(const std::string &/*protocol*/,
 //----------------------------------------------------------------------------
 bool verify(const conduit::Node &n,
             Node &info)
+{
+    // TODO(JRC): Make the second parameter the largest value for the 'index_t'
+    // type (unfortunately this isn't trivial to get because 'index_t' can be
+    // signed or unsigned depending on Conduit's build options).
+    return mlarray::verify(n, info, 0, 10000);
```
xjrc: This needs to change and be made more general before this pull request is merged. Any advice on how to properly create a version of `std::numeric::limits<T>::max` that supports the build-configurable type `index_t` would be appreciated.

cyrush: @xjrc  you should be able to use that template fun with index_t, i quickly tried:

 
```
   #include <limits>
    std::cout << std::numeric_limits<conduit::index_t>::max() << std::endl;
```

Maybe you just hit a typo? (`numeric_limits` (ok) vs `numeric::limits` (not ok) ?)

xjrc: I'm not really sure why it didn't work before when I tried it; I seem to remember compiling and running the code with `std::numeric_limits` expressions, so I must've messed it in some other novel way. In any case, it seems to be working now; thank you for the advice!

PR #433, opened 2019-07-15
isend, irecv using schema

davidHysom:
added two methods:
 isend_using_schema
  irecv_using_schem
Added a unit test:
  TEST(conduit_mpi_test, isend_irecv_using_schema)

### DIFF
--- a/src/libs/relay/conduit_relay_mpi.cpp
+++ b/src/libs/relay/conduit_relay_mpi.cpp
@@ -345,7 +345,6 @@ send_using_schema(const Node &node, int dest, int tag, MPI_Comm comm)
     return mpi_error;
 }
 
-
 //---------------------------------------------------------------------------//
 int
 recv_using_schema(Node &node, int src, int tag, MPI_Comm comm)
@@ -353,7 +352,6 @@ recv_using_schema(Node &node, int src, int tag, MPI_Comm comm)
     MPI_Status status;
     
     int mpi_error = MPI_Probe(src, tag, comm, &status);
-    
     CONDUIT_CHECK_MPI_ERROR(mpi_error);
     
     int buffer_size = 0;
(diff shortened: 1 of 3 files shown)


cyrush:
@davidHysom  Sorry it took me a bit to review & thanks for taking a pass at this. 

The travis ci failure isn't related to your code. 

@bryujin helped me review, he wrote most of the conduit relay mpi support. 

reviewing, I don't think this quite accomplishes a generic `isend` / `irecv`. 

The send data needs to persist until the call to mpi `wait_send`.

Right now the Node that holds the data to send is local to the `isend_w_schema` function, and it will go out of scope after the call returns.  You can fix this by putting the message in the node in the Request object. (request->m_buffer).

The irecv is a bit tricker.   It also has the issue with the data needing to persist until the mpi `wait_recv`.
but tricker b/c I think the`wait_recv` would need to do the unpacking. 

Also the `MPI_Probe` and `MPI_Get_count` calls are effectively a synchronous call before you post the irecv. With the current structure you won't get any async benefit between the `irecv_w_schema` and the `wait_recv`.

davidHysom:
Good points. Here's what I think may work. Design a class that contains:

  - pointers or references to the Nodes that are to be sent/received
  - public methods to start the sends and receives

Internally the class can keep any needed, locally allocated  memory (e.g, "the Node that holds the data to send is local to the isend_w_schema function").

Unless you can poke a big hole in this idea, I'll code up a strawman API so you can take a look.
In the meantime I'll kill my PR, as I think it's simpler to start over.

cyrush:
@davidHysom thats basically the strategy we have with the current `Request` class for the non-schema variants of isend / irecv.  

The additional wrinkle with isend and irecv w/ schema is that the pointers to the recv data have to exist and be allocated when you can post the irecv, however we don't know how large they will be to preallocate them.

davidHysom:
Yes; so underneath the covers the class would send two messages, the first being the size to be received. This is what I have coded in LBANN's data_store_conduit class.

cyrush:
@davidHysom that certainly can work, in the past we have gotten hung up on how to make the interface understandable and convey the multi-send semantics. maybe your pass could help us converge to something solid.

Issue #437, opened 2019-08-03
check zlib  setup in config.mk


cyrush:
we have a `CONDUIT_ZLIB_DIR` section on config.mk, but it is not wired up. 

Zlib would come in from hdf5, so we should simply handle as part of that case.

cyrush:
resolved in #441

PR #441, opened 2019-08-12
config.mk improvements

cyrush:
update config.mk to include zfp
add logic for transitive static dep of hdf5 in config.mk


cyrush:
This should resolve #437

Issue #471, opened 2019-11-14
Minimum CMake required is 3.9 with the use of NATIVE_COMMAND


David-Camp:
Hello,

You need to update the CMake required version to 3.9 because in cmake/thirdparty/SetupHDF5.cmake line 230, you use separate_arguments with the NATIVE_COMMAND argument. This was not supported until 3.9.
Right now it is cmake_minimum_required(VERSION 3.0) in Conduit/src/CMakeLists.txt

David Camp

cyrush:
thanks for the report david!

cyrush:
resolved in #475

PR #475, opened 2019-12-09
py modules: defer linking and symbols to final py interp

### LINKED ISSUE #472: change python linking strategy
good advice from conda-forge gitter:

don't link against the python shared lib in extension modules
only link against it in executables that embedd python.

this follows new guidance from python 3.8 release, https://docs.python.org/3/whatsnew/3.8.html#debug-build-uses-the-same-abi-as-release-build

it could also solve our long stand off with static vs python

cyrush:
working to resolve #472


cyrush:
also resolves #471

PR #482, opened 2019-12-27
python: allow node.set_external to take a conduit node

### LINKED ISSUE #470: python interface bug with node set_external
node.set_external should accept a conduit node as well

currently: 

`TypeError: set_external requires a numpy array`

cyrush:
resolves #470


cyrush:
resolves #470

PR #495, opened 2020-01-09
Set _IMPORT_PREFIX, as needed by vcpkg machinery

agcapps:
When used through vcpkg, several LIBRARY_DIR variables get set to the value `${_IMPORT_PREFIX}`.  This change copies the logic for setting up _IMPORT_PREFIX to before it gets used.

Many thanks to @white238 for his help on this.


cyrush:
thanks @agcapps and @white238 !

PR #502, opened 2020-02-08
Update Uberenv and introduce Gitlab CI [feature/uberenv_ci]

adrienbernede:
In this PR I updated uberenv to the latest version (https://github.com/LLNL/uberenv/tree/master + https://github.com/LLNL/uberenv/pull/20), and introduce a very basic Gitlab CI pipeline.

CI simply compiles 4 conduit targets using uberenv.

I will also include https://github.com/LLNL/uberenv/pull/19, which should help preserve the current workflow.

This PR is ready for review.


scripts/uberenv/project.json
```diff
@@ -1,8 +1,10 @@
 {
 "package_name" : "conduit",
-"uberenv_package_name" : "uberenv-conduit",
-"spack_url": "https://github.com/alpine-DAV/spack",
-"spack_branch": "task/2019_10_update_conduit",
+"package_version" : "master",
+"package_final_phase" : "configure",
+"package_source_dir" : "../..",
+"spack_url": "https://github.com/spack/spack",
+"spack_branch": "develop",
```
cyrush: i know we need to use a newer spack, but I am a bit worry about doing this without testing in a few places (NERSC)

I would also want it to be pinned to some version to avoid random failures when spack develop moves on.

scripts/uberenv/packages/conduit/package.py
```diff
@@ -0,0 +1,498 @@
+# Copyright 2013-2019 Lawrence Livermore National Security, LLC and other
```
cyrush: I don't think this is the latest conduit spack pkg -- it should have version 0.5.1 in it)

scripts/uberenv/project.json
adrienbernede: OK

adrienbernede:
@cyrush I re-introduced uberenv package in, modifying it to have a "configure" phase.
However, you may prefer using the existing package in Spack.

scripts/uberenv/packages/conduit/package.py
adrienbernede: See my comment below

adrienbernede:
@cyrush I need to be added as "developer" (I’m not sure of the terminology) to create the mirroring in Gitlab.

cyrush:
I see -- I think my latest spack package does have configure:

https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/conduit/package.py

can you see if that is compatible?

scripts/uberenv/packages/conduit/package.py
adrienbernede: My mistake, I added a configure phase in ascent, conduit already at it.

adrienbernede: Update will be straightforward.

cyrush:
@adrienbernede was there some change that increased the output of the spack builds? The travis jobs failed b/c they hit the max log file size.

scripts/uberenv/project.json
cyrush: I can create a fork in the alpine space that points to a newer spack

.gitlab-ci.yml
```diff
@@ -0,0 +1,55 @@
+##############################################################################
+# Copyright (c) 2016-19, Lawrence Livermore National Security, LLC and Serac
+# project contributors. See the COPYRIGHT file for details.
+##############################################################################
```
cyrush: switch to (2014-2020) and Conduit project contributors

.gitlab/ci/build_lassen.yml
```diff
@@ -0,0 +1,38 @@
+##############################################################################
+# Copyright (c) 2016-19, Lawrence Livermore National Security, LLC and Umpire
+# project contributors. See the COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (MIT)
```
cyrush: again, switch to (2014-2020) and Conduit project contributors

mclarsen:
We found that azure will tolerate log files of .5GB.....

adrienbernede:
> @adrienbernede was there some change that increased the output of the spack builds? The travis jobs failed b/c they hit the max log file size.

@cyrush The reason is because dev-build is verbose by default on dependencies build. I will make it quiet (there’s an option for it).

scripts/uberenv/project.json
adrienbernede: I am

.gitlab-ci.yml
adrienbernede: Does it mean all the sources in conduit should switch to 2020, or is it done incrementally when the given file is modified?
Serac to Conduit: OK

scripts/uberenv/project.json
adrienbernede: v0.13.4 is in fact too old (does not include latest conduit).
So I will try using the merge commit of this branch into develop. Then we can either keep the commit in the project.json, or use a clone is ascent.

cyrush:
is uberenv still copying the resulting host-config.cmake to the root uberenv_libs dir?

(it seems that that is not the case and that is what is causing the travis ci error)

adrienbernede:
I agree that this is what is happening.

I am focusing on inconsistencies I have with the tests of Gitlab, but I will address that afterwards. I know the location changed several times when we worked on Serac. I want to remind the reasoning before taking any action.

I also have an error related to python configuration.

cyrush:
sounds good,  I didn't catch that that had changed in uberenv. 

we do need the host-config file to exist in that location -- the primary way I configure when developing (across projects) involves using the file in that location.

adrienbernede:
I understand. As far as I remember, we still had at least a link there.

cyrush:
yes - symlink is fine

adrienbernede:
Now I remember:
The host-config file is generated in the project install directory in spack. Then is used to be copied or symlinked to the uberenv-libs directory.
I changed that for a link to the project root directory because I wanted to share the same uberenv-libs between several instances of the project, (hello CI).

cyrush:
in this case what is the project root directory? 

Most of our existing CI infrastructure is going to look for it in the prior location. (Also true for Ascent)

adrienbernede:
/home/travis/build/LLNL/conduit

adrienbernede:
In fact, what is happening is:
- in dependencies only mode, `spack dev-build` generates a `spack-build` directory at the `package_source_dir` location. It means the project is directly buildable with Spack dependencies there, so the host-config file was not copied or linked to `uberenv-libs` directory any more.
- in install mode, the host-config file is sym-linked in `uberenv-libs` directory.

I could copy the generated host-config file from `spack-build` to `uberenv-libs`. I would like to learn more about your multi-project workflow though.

cyrush:
so the dev build build in directories outside of the prefix passed to ubrerenv? 

If you want to share uberenv_libs dirs between CI builds and the host-config.cmake could conflict, could your CI move the generated host config instead of changing the behavior of uberenv?

adrienbernede:
Dev-build would build the dependencies in prefix, but perform the build at the source location, either in a spack-build directory for CMake or in source for Makefile. It makes sense since the idea is to build with the current version of the sources. Is that an issue? Maybe because it breaks the rule "everything Spack does goes in uberenv-libs"?

I will find a way to preserve the original behavior.

cyrush:
I see, the host config generation process (since it is part of the final package) will happen in a directory outside of the prefix. I didn't realize the behavior was changing 

If at all possible, I'd like to preserve the old semantics of uberenv:

default mode: creates a working set of dependencies and a host config file in --prefix 
install mode: creates a full spack build of the package, a host config file and a symlink to the install dir in --prefix.

adrienbernede:
This is what I did in the last commit, at one detail:
I created `conduit/uberenv-libs/host-config.cmake` as a symbolic link on `conduit/spack-build/host-config.cmake`. It means you need to preserve `spack-build`. I will change it for a copy (like I initially said).

Remaining decision: Do you want to get rid of this `spack-build`? I guess so.

I also have to address a limitation: Spack will refuse to `dev-build` a package already installed. If that’s possible in your workflow, then we may have to uninstall the package first, or ask the developer to do so.

cyrush:
thanks for your efforts on this -- you are immersed in the details, but I admin am a little lost with the changes. 

My hope is that we could use the newer spack features without changing really much of anything related to the workflow of uberenv. Including keeping any build artifacts limited to --prefix, etc. 
Another important thing: you can run uberenv again after success and get no errors.

I don't have a good handle on the limitations of the dev-build etc. Are their any aspects of it might need to be address by spack folks before we should consider using it?

adrienbernede:
It’s OK to deal with the details, and that’s no big changes we are talking about.

It’s not a Spack limitation proper. And I with it was already the case before: `uberenv --install`, followed by `uberenv` doesn’t make sense because conduit would already be installed.

cyrush:
--install could be run multiple times and if it was already installed, it just reported success. 
I have relied on this for remote systems where builds might hang, or I might get disconnected while in transit.

cyrush:
@adrienbernede 
Sorry I haven't been able to chime in on this more. 

I had an issue with spack develop building a part of the deps on my mac that I will need to resolve before we can move forward. I also want to do sanity checks at NERSC. Unfortunately,  I have many other things in my queue before I will be able to work though these.

adrienbernede:
@cyrush, well I expected do get more time to spend on it myself...

cyrush:
@adrienbernede sounds good, just wanted to let you know it is still on my radar.

adrienbernede:
@cyrush 
I think we are close to the initial workflow. I see one thing that still needs to be handled:

Initial workflow:
- build and install modes both "install" host-config file in spack (fake or real package install)
- uberenv symlinks it in `prefix`

Current workflow:
- install mode preserved
- build mode generates a `spack-build` directory with the host-config file. The host-config file is then _copied_ in the prefix dir, and `spack-build` is _removed_.

The difference is that if one re-runs `uberenv` after a build, the host-config file will be lost. (previously only unlinked).

However I don’t think this is a big issue, because re-running uberenv in build mode for a given configuration is fast enough.

adrienbernede:
Update and clean branch

adrienbernede:
@cyrush with the latest changes, I’m addressing two things:

1. In uberenv, I’m fixing the issue with already installed packages. If a spec is already installed, and one ask for a dev-build of it, uberenv with create a link to the already installed host-config file.
2. In CI, I am introducing new jobs testing "what if I update uberenv?".

Notes on 1.: To do so, I had do change the behavior of the command executor sexe() in order to allow capturing the output while also displaying it.  

Notes on 2.: Paradoxaly, these new tests would fail since uberenv is currently newer in conduit than in uberenv repo... The next step is to add "what if I update spack". At the moment I don’t see any obstacle in introducing these tests, and my goal is to have them triggered only periodically and/or on changes in master.

adrienbernede:
Also: I still need dev access to LLNL/conduit in order to setup mirroring on LC.

cyrush:
For the uberenv check, one way we have handled things like this (in other projects) is to record a hash for the current version in a text file. Then when you update the script you also have to update the hash or the test fails. 

This is a good strategy to prevent surprise updates, but it doesn't help you test and vet a new version of uberenv -- which I think is what you are aiming for?

Like with spack, I think we will be intentional about when we update uberenv. Staying up-to-date with the latest and greatest isn't necessary, when we can always easily explicitly update and vet though our CI process.

adrienbernede:
LGTM

adrienbernede:
(Previous comment is a sign-off to trigger GitLab CI)

adrienbernede:
The goal here is not to force the adoption of latest Uberenv or Spack, but to make it possible to test Conduit with both of them.

It would allow to ease the "test and vet" phase and "document" how to do that. But it also means I will be able to trigger such a test in Conduit from Uberenv CI (or Radiuss CI), to check if the last version of Uberenv or Spack would break something in  Conduit.

Even if this isn’t the best example, I think it’s an interesting demonstrator, and allows me to experience with the best way to do it in our context.

(By the way, if you feel the need for such a cross-project integration between two Radiuss projects, just let me know. I would like to promote this for projects using submodules like BLT.)

cyrush:
@adrienbernede that sounds good, knowing the answer to the "what if" we use the latest version will be helpful.

adrienbernede:
Close and transferred to #517.

PR #506, opened 2020-02-20
Feature/agcapps/carray

agcapps:
Ideas for compressed array representation in Conduit


cyrush:
quick note: I think the term compressed array will be confusing b/c we are dealing with an explicit sparse representation in contrast to something like zfp (which conduit already supports) that offers a data driven compression

agcapps:
This was an instructive straw man.  The direction forward was a little different: see issue #511 .

PR #508, opened 2020-02-26
add unit test that save mesh bp misc examples to yaml

cyrush:


cyrush:
@agcapps you can use this to look at the "misc" examples easily via yaml files. 

After running the standard mesh blueprint examples unit test  (`tests/blueprint/t_blueprint_mesh_examples`), you will now see:

```
misc_example_adjsets.yaml
misc_example_matsets.yaml
misc_example_nestsets.yaml
misc_example_specsets.yaml
```

The matset example has volume fractions, however I think we need to enhance the matset example to include a material dependent field.

cyrush:
@xjrc just FYI, I added this as a convenient way to view the output of these examples.

agcapps:
> @agcapps you can use this to look at the "misc" examples easily via yaml files.
> 

Thanks much.

PR #522, opened 2020-03-13
add mpi enabled conduit_blueprint lib

cyrush:
work in progress


cyrush:
TODO:
- [ ] Add mpi example (take from ascent's tests?)
- [ ] Add mpi verify and test for various multi-domain across mpi tasks
- [ ] Add number of domains query (as property example)
- [ ] Solve https://github.com/LLNL/conduit/issues/559

cyrush:
still more we can add to this lib, but initial support is there

cyrush:
@xjrc  and @mclarsen  -- can you give these changes a quick look?

src/tests/blueprint/t_blueprint_mpi_relay.cpp
```diff
@@ -113,28 +116,185 @@ TEST(blueprint_mpi_relay, basic_use)
...
+                                                       comm);
+
+    // check verify
+    EXPECT_TRUE(blueprint::mpi::mesh::verify(dset,v_info,comm));
+
+    // locally, expect 3 domas for rank 0, 2 for rank 1
+    if(par_rank == 0)
+        EXPECT_EQ(blueprint::mesh::number_of_domains(dset),1);
+    else
+        EXPECT_EQ(blueprint::mesh::number_of_domains(dset),0);
```
xjrc, line 277: Haven't required braces for code blocks historically been a part of Conduit style? This seems to be the case looking briefly at [`conduit_node.cpp`](https://github.com/LLNL/conduit/blob/master/src/libs/conduit/conduit_node.cpp#L408), but I'm not certain this style is ubiquitous across the code.

src/tests/blueprint/t_blueprint_mpi_relay.cpp
```diff
@@ -113,28 +116,185 @@ TEST(blueprint_mpi_relay, basic_use)
...
+
+    Node dset, v_info;
+    blueprint::mpi::mesh::examples::spiral_round_robin(1,
+                                                       dset,
+                                                       comm);
+
+    // check verify
+    EXPECT_TRUE(blueprint::mpi::mesh::verify(dset,v_info,comm));
+
+    // locally, expect 3 domas for rank 0, 2 for rank 1
```
xjrc, line 273: This comment looks incorrect based on the checks below.

src/tests/blueprint/t_blueprint_mpi_verify.cpp
```diff
@@ -0,0 +1,149 @@
...
+    EXPECT_TRUE( conduit::blueprint::mpi::verify("mesh",mesh,info, MPI_COMM_WORLD) );
+
+    // check the number of domains
+    EXPECT_EQ( conduit::blueprint::mpi::mesh::number_of_domains(mesh,MPI_COMM_WORLD), 1);
+
+    // check hypothetical index gen
+    conduit::blueprint::mpi::mesh::generate_index(mesh,
+                                                  "",
+                                                  bp_index["mesh"],
+                                                  MPI_COMM_WORLD);
```
xjrc, line 133: Since there are two check procedures that are identical here (i.e. generate base mesh on some rank, verify, check domains, generate index), this should probably be a `for` loop instead to reduce redundancy.

src/tests/blueprint/t_blueprint_mpi_relay.cpp
```diff
@@ -113,28 +116,185 @@ TEST(blueprint_mpi_relay, basic_use)
...
+    string output_base = "tout_blueprint_mpi_relay_spiral_mpi_dist_5doms";
+    // what we want:
+    // relay::mpi::io::blueprint::save_mesh(dset, output_path,"hdf5");
+    conduit::relay::mpi::io::blueprint::save_mesh(dset,
+                                                  output_base,
+                                                  "hdf5",
+                                                  comm);
+    MPI_Barrier(comm);
+
+    // read this back using load_mesh, should diff clean
```
xjrc, line 239: Was the intent to validate the results by calling `diff` as this comment and the `n_diff_info` variables suggest? If so, I think a bit of code is missing below.

src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
```diff
@@ -0,0 +1,212 @@
...
+    n_snd.set_external(&rank_send,1);
+    n_reduce.set_external(&selected_rank,1);
+
+    relay::mpi::min_all_reduce(n_snd, n_reduce, comm);
+
+    if(par_rank == selected_rank )
+    {
+        if(::conduit::blueprint::mesh::is_multi_domain(mesh))
+        {
+            ::conduit::blueprint::mesh::generate_index(mesh.child(0),
```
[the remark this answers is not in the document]
cyrush: I think we only need it when we are using both the mpi and non mpi namespaces, but I agree it's a good practice.

src/tests/blueprint/t_blueprint_mpi_relay.cpp
cyrush: good catch, I'l resolve

src/tests/blueprint/t_blueprint_mpi_relay.cpp
```diff
@@ -113,28 +116,185 @@ TEST(blueprint_mpi_relay, basic_use)
...
+
+    Node dset, v_info;
+    blueprint::mpi::mesh::examples::spiral_round_robin(5,
+                                                       dset,
+                                                       comm);
+
+    // check verify
+    EXPECT_TRUE(blueprint::mpi::mesh::verify(dset,v_info,comm));
+
+    // locally, expect 3 domas for rank 0, 2 for rank 1
```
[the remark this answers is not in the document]
cyrush: thanks for catching this

src/tests/blueprint/t_blueprint_mpi_relay.cpp
cyrush: thanks for catching this!

src/tests/blueprint/t_blueprint_mpi_relay.cpp
cyrush: it was, I had a snag early in the process and disabled the check. I will add this back in

src/tests/blueprint/t_blueprint_mpi_verify.cpp
cyrush: good suggestion, I'll implement

nselliott:
This will be a good addition.  Not part of this PR, but I think the pieces are here to enable generate_index to automatically create the partition map array that we've discussed elsewhere.

cyrush:
> This will be a good addition. Not part of this PR, but I think the pieces are here to enable generate_index to automatically create the partition map array that we've discussed elsewhere.

yes, this provides a nice home for that functionally -- barriers to tackling that are getting taken down.

src/libs/blueprint/CMakeLists.txt
```diff
@@ -179,3 +179,48 @@ if(PYTHON_FOUND)
...
+#
+set(conduit_blueprint_mpi_deps conduit conduit_blueprint mpi conduit_relay_mpi)
+
+#
+# setup conduit blueprint mpi lib
+#
+add_compiled_library(NAME conduit_blueprint_mpi
+                     EXPORT conduit
+                     HEADERS ${conduit_blueprint_mpi_sources}
+                     SOURCES ${conduit_blueprint_mpi_headers}
```
nselliott, line 216: sources and headers are mixed up here.

cyrush: good catch, I'll fix!

PR #523, opened 2020-03-13
Add base methods for Schemas and Nodes to handle children with pathli…

HaluskaR:
…ke names

This is to open discussion on naming, further tests, gotchas, etc. Still need to add remove methods and other niceties.

### DIFF
--- a/src/libs/conduit/conduit_node.cpp
+++ b/src/libs/conduit/conduit_node.cpp
@@ -499,7 +499,7 @@ Node::set_node(const Node &node)
         for (std::vector<std::string>::const_iterator itr = cld_names.begin();
              itr < cld_names.end(); ++itr)
         {
-            Schema *curr_schema = this->m_schema->fetch_ptr(*itr);
+            Schema *curr_schema = &this->m_schema->add_child(*itr);
             size_t idx = (size_t) this->m_schema->child_index(*itr);
             Node *curr_node = new Node();
             curr_node->set_schema_ptr(curr_schema);
@@ -11966,6 +11966,11 @@ Node::children() const
 Node&
 Node::add_child(const std::string &name)
 {
+    if(has_child(name))
+    {
+        return child(name);
+    }
+
     Schema &child_schema = m_schema->add_child(name);
     Schema *child_ptr = &child_schema; 
     Node *child_node = new Node();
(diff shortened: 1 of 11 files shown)


HaluskaR:
Naming-wise, get_child vs. fetch_child is a bit confusing, since there's not a concrete difference between "get" and "fetch". Schema has both a shallow and deep check ("has_child" vs "has_path"), which I think works really well. child vs. descendant might also make sense, though in this case, the child is what the newer method is getting, and "descendant" is a bit cumbersome

HaluskaR:
I was wondering about the coveralls coverage...turns out I only committed the source, not the tests. d'oh! One sec.

cyrush:
so far this is looking good!

I am open to making an API breaking change here to get to a better set of names.
I agree `has_child `vs `has_path` is good, maybe we move to change node::fetch_child.

Let me mull it a bit more. 

One way to test this would be to make the change and see if any of the unit tests make assumptions.

src/libs/conduit/conduit_node.cpp
```diff
@@ -11962,6 +11962,33 @@ Node::children() const
...
+    if(!m_schema->has_direct_child(name))
+    {
+        CONDUIT_ERROR("Cannot fetch non-existent "
+                      << "child \"" << name << "\" from Node("
+                      << this->path()
+                      << ")");
+    }
+    size_t idx = (size_t)m_schema->child_index(name);
+    return *m_children[idx];
+}
```
cyrush: we will also need a non-const implementation, I think thats issue the CI is pointing out

HaluskaR: Can do! I expect logic to be identical between the two since it's only a getter. Given that, do you have preference for the implementation? I know the brief way is to have the non-const implementation temporarily const-cast then call the const function, _Effective C++_ style, but other projects prefer having the two full methods.

cyrush: Let's go with two methods, since they are pretty short.  Thanks!

HaluskaR: Done!

cyrush:
Looks good! I am going to add some unit tests to look at json + yaml, roundtrip.

cyrush:
Update on my thoughts on naming:
I think we can change `get_child` to just `child`, node already has `Node::child(index_t index)` that provides similar semantics to what we want for `get_child`, seems much cleaner. What do you think?

Pushing a bit further with testing, it seems  we will have to re-plum some operations, like generic sets of nodes, diffs, etc.  B/c they are using generic fetches. 

I'll try to get a unit test that covers most these cases and then we can attack.

cyrush:
@HaluskaR fixing some of the internals will be on my plate -- I am digging in on that.

cyrush:
@HaluskaR  I think I am converging on `child` as the new method names. 

I started to fix some of the tricker internal changes and a small bug with add_child.
I still need to fix the generator code  (which handles JSON and YAML parsing)

cyrush:
@HaluskaR I think this might be ready to merge, I updated the change log and added you to the contributors. Do you want to put it through any more of your own testing?

HaluskaR:
Thank you for adding the tests and all the internals! Looks like all the cases that spring to mind are covered--I'll check while I wait for the last CI to complete and then perform the merge.

PR #527, opened 2020-03-25
mesh bp, add venn example

cyrush:


src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2532,6 +2532,175 @@ void spiral(index_t ndoms,
...
+void venn(index_t nx,
+          index_t ny,
+          float64 radius, 
+          Node &res)
+{
+    res.reset();
+    // create a rectilinear coordset 
+    res["coordsets/coords/type"] = "rectilinear";
+    res["coordsets/coords/values/x"] = DataType::float64(nx+1);
+    res["coordsets/coords/values/y"] = DataType::float64(ny+1);
```
xjrc: For sections like these where you reuse a node prefix many times, I've found it to be useful to create a reference to that prefix to make things a bit more clean and concise (e.g. `conduit::Node& res_coords = res["coordsets/coords"];`).

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2532,6 +2532,175 @@ void spiral(index_t ndoms,
...
+    res["fields/radius_c/values"] = DataType::float64(nx * ny);
+    // circle b vf
+    res["fields/circle_c/association"] = "element";
+    res["fields/circle_c/topology"] = "topo";
+    res["fields/circle_c/values"] = DataType::float64(nx * ny);
+
+    // per element how many circles overlap
+    res["fields/overlap/association"] = "element";
+    res["fields/overlap/topology"] = "topo";
+    res["fields/overlap/values"] = DataType::float64(nx * ny);
```
xjrc: Given that all of these fields are initialized in the exact same way, a for-loop over field names (e.g. `radius_*`, `circle_*`, `overlap`) could be used to reduce some of the code duplication.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2532,6 +2532,175 @@ void spiral(index_t ndoms,
...
+    // centered at:
+    //   x = 1/2 of width
+    //   y = 2/3 of width
+    float64 c_center_x = 0.5;
+    float64 c_center_y = 0.3333333333;
+
+
+    float64 y = 0;
+    
+    index_t idx = 0;
```
xjrc: For index iteration situations like these, I've found embedding the iteration directly into the associated nested for-loop is useful for scoping and signalling intent, e.g.

```
for(index_t j = 0, idx = 0; j < ny; j++)
{
    for(index_t i = 0; i < nx; i++, idx++)
    {
        // ... implementation ...
    }
}
```
Given that the benefits of using this pattern are marginal and potentially offset by the nonstandard syntax, I think that either approach is fine; I just wanted to offer an alternative just in case it seemed preferable.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2532,6 +2532,175 @@ void spiral(index_t ndoms,
...
+
+            if(rad_b[idx] > 0)
+                cir_b[idx] = 1.0/olap[idx];
+            else
+                cir_b[idx] = 0.0;
+
+            if(rad_c[idx] > 0)
+                cir_c[idx] = 1.0/olap[idx];
+            else
+                cir_c[idx] = 0.0;
```
xjrc: If all of the circle centers (e.g. `a_center_*`) and their associated arrays (e.g. `cir_a`, `rad_a`) were put into parallel arrays of their own, this loop body could be transformed into a loop of its own over the individual circles, which would help eliminate some code duplication. If we choose to do this, it would probably help to name the circles by their indices in this array (e.g. `circle_0` instead of `circle_a`) in order to improve the clarity of the code.

src/tests/blueprint/t_blueprint_mesh_examples.cpp
```diff
@@ -344,7 +344,7 @@ TEST(conduit_blueprint_mesh_examples, braid_too_small_npts)
     braid_type_strings.clear();
     braid_type_strings.push_back("points");
     braid_type_strings.push_back("points_implicit");
-
```
xjrc: I'm happy to see a lot of the unused whitespace go!

cyrush:
@xjrc thanks fo the feedback, this is very much a work in progress

To answer your two general qs:

* Yes, the idea is to build from this and use different matset notions. I was sprinting to create the proper values and then to plot and verify they made sense with VisIt. This is why I used normal fields.

Even when we have proper matsets, I think we may want to keep these fields for debugging and verification. I know I will want them when testing the Mesh Blueprint Plugin updates to VisIt.

To evolve the venn function, I suggested to @agcapps that we could add a string arg along the lines of matset_type = {"full"  or "sparse_by_element"  or "sparse_by_material"} to allow folks to select the different matset cases. 

* Yes, we definitely want to have an entry in the docs with an example render, etc -- but haven't started that yet.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2532,6 +2532,300 @@ void spiral(index_t ndoms,
...
+        if (fgvf < 1.)
+        {
+            bg_val[nidx] = 1. - fgvf;
+            bg_idx[nidx] = idx;
+            nidx += 1;
+        }
+    }
+}
+
+void venn(const std::string &storage_type,
```
cyrush, line 2648: can we change `storage_type` to matset_type?

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2532,6 +2532,300 @@ void spiral(index_t ndoms,
...
+        y+=dy;
+    }
+
+    // Shape in materials.
+
+    if (storage_type == "full")
+    {
+        venn_full_matset(res, nx, ny);
+    }
+    else if (storage_type == "sparse")
```
cyrush, line 2822: to prep for the other scheme , maybe call this "sparse_by_material" ?
(even "sparse_by_material"  seems ambiguous)

agcapps: Changed to "sparse-by-material"

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
agcapps: fixed

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
agcapps: done

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
agcapps: done

cyrush:
@agcapps looking good!

I think the last thing we need is a field that has a matset_values.

For the mechanics: 

The values live in a path like:

`fields/field_name/matset_values: (values)`

For full case, this is a mcarray with all of the values per material

Both of the sparse cases use the indexing from the matset, the index description doesn't need to be duplicated:

For sparse_by_mat, it's an mcarray of just non zero-per material 

for sparse_by_element, it's a single array with the values that follows the indexing scheme in the matset 


My idea for the field we should create: Could we just use the element `area`?

area seems good for validation --  but the per-material value might not be intuitive at first.

This would need to equal the full area of each element:

`area = vf(bg) * per_mat(area,bg) + vf(a) * per_mat(area,a) + vf(b) * per_mat(area,b)  + vf(c) * per_mat(area,c) 
`
That is the per-material value is the field value the element would have if that material filled the entire element.

Example:

For a single element square quad with sizes of length 1 (an area of 1) + two materials (say 40% and 60% vfs)
Both per-material field value `area`s would be 1. 

So for the area, the per material area is just the full area of the element? Maybe a bit strange, but good for validation.

cyrush:
@xjrc It feel off my radar that Arlie added both style examples in this PR can you give them a quick review and we can move towards merging.  I'll volunteer add a basic section to the docs as the final part of the PR.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
cyrush: i think this logic is a bit verbose, but I think its easier to comeback to cold to understand the example.

Recently revisiting the nest set examples, extra looping make it harder to dust off.  So prefer the simpler but longer formation.

xjrc:
@cyrush: Sure, I'll have a look this afternoon.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2532,6 +2532,389 @@ void spiral(index_t ndoms,
...
+    for(index_t idx = 0; idx < elements; idx++)
+    {
+        mat_ca[idx] = cir_a[idx];
+        mat_cb[idx] = cir_b[idx];
+        mat_cc[idx] = cir_c[idx];
+        mat_bg[idx] = 1. - (cir_a[idx] + cir_b[idx] + cir_c[idx]);
+    }
+}
+
+void make_sparse(Node & src, index_t len, Node & n)
```
xjrc, line 2568: Instead of passing `len`, couldn't we just use `src.dtype().number_of_elements()` instead? This would simplify the function signature and make the intent of the function a bit clearer. Also, naming it something like `nsrc` or `nfull` to make it crystal clear which length its describing?

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2532,6 +2532,389 @@ void spiral(index_t ndoms,
...
+{
+    float64_array src_val = src.value();
+
+    index_t nz = 0;
```
xjrc, line 2572: The variable `nz` contains the number of non-zero entries in the data array being made sparse, but its name can be interpreted to convey the opposite (e.g. "**n**um**z**eroes" instead of "**n**on**z**eroes"). In the interest of clarify, could we rename this variable to something like `num_sparse` or `nsparse`?

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2532,6 +2532,389 @@ void spiral(index_t ndoms,
...
+    {
+        if (cir_a[idx] + cir_b[idx] + cir_c[idx] < 1.) bgcount += 1;
+    }
+
+    bg["values"].set(DataType::float64(bgcount));
+    bg["element_ids"].set(DataType::int32(bgcount));
+    float64_array bg_val = bg["values"].value();
+    int32_array bg_idx = bg["element_ids"].value();
+
+    index_t nidx = 0;
```
xjrc, line 2635: Just like `nidx` refers to the index in the `n` node in `make_sparse`, could we name this instead to `bgidx` since it refers to an index in the `bg` node?

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2532,6 +2532,389 @@ void spiral(index_t ndoms,
...
+    for (index_t idx = 0; idx < len; ++idx)
+    {
+        if (src_val[idx] > 0)
+        {
+            nz += 1;
+        }
+    }
+
+    n["values"].set(DataType::float64(nz));
+    n["element_ids"].set(DataType::int32(nz));
```
xjrc, line 2582: Looking at how this is called, it looks like it will create the following structure:

```
matsets:
  matset:
    volume_fractions:
      circle_a:
        values: [ ... ]
        element_ids: [ ... ]
      circle_b:
        values: [ ... ]
        element_ids: [ ... ]
      circle_c:
        values: [ ... ]
        element_ids: [ ... ]
```

Based on the latest from #511, I think it should instead be structured like:

```
matsets:
  matset:
    volume_fractions:
      circle_a: [ ... ]
      circle_b: [ ... ]
      circle_c: [ ... ]
    element_ids:
      circle_a: [ ... ]
      circle_b: [ ... ]
      circle_c: [ ... ]
```

The schema described in #512 is currently out of date, so this is largely my fault. I will update it presently so that this isn't unclear going forward.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2532,6 +2532,389 @@ void spiral(index_t ndoms,
...
+    res["matsets/matset/material_map/bg"] = 0;
+
+    // All the volume fractions go here ("one big buffer")
+    res["matsets/matset/volume_fractions"].set(DataType::float64(vfcount));
+    // The material measured by each volume fraction
+    res["matsets/matset/material_ids"].set(DataType::int32(vfcount));
+    // The number of volume fractions in an element
+    res["matsets/matset/sizes"].set(DataType::int32(elements));
+    // The offset of the first vf in an element
+    res["matsets/matset/offsets"].set(DataType::int32(elements));
```
xjrc, line 2686: The structure here looks good!

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
agcapps: I'll get `len` from the dtype --- that is much cleaner.  I'll also rename `n` to something explicit.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
agcapps: Thanks.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
agcapps: Also, I pass `nx` and `ny` all over the place.  Is there a way I can easily pull those out of a dtype?

xjrc: Not individually, but you can eliminate them from all the `venn_*` functions and replace the calculation for the `elements` variable with `res["fields/circle_a/values"].dtype().number_of_elements()`. It may be more proper to calculate this value directly from the topology itself, but just grabbing it from a spanning field (e.g. any `circle_*` field) seems fine to me as well.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
agcapps: I went with `nsparse`.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
agcapps: I changed the `n` and `n_` names and prefixes to `sparse` and `sparse_`.  I think it is much clearer.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
agcapps: Changed to the structure you wrote here.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
agcapps: I put `nx` and `ny` into another Node (commented as "not part of the blueprint") and pull them out every time I need them.

agcapps:
@cyrush , @xjrc , I just pushed implementations of by-element and by-material sparse representation.  I made two fields, `area` (for ease of verification) and "`importance`" (totally made-up; intended to show some contrast to `area`).  I also verified that the `area` field equals the calculated element area for all three `venn` examples.

cyrush:
FYI: I am going to work on the docs entry and get this in after we merge the new Julia AMR examples #556

xjrc:
I'm also planning to give this another look this afternoon and to do one last review sweep before merge.

xjrc:
I made some minor updates to clean up signatures for the `venn` helpers and to reduce code duplication in the `venn` regression test functions. Everything looks good to me now; I consider this PR ready to merge whenever.

cyrush:
Note: about "meta/importance"


While it's fine to have other (non-blueprint) stuff in tree --  folks could confuse this extra info for something that the blueprint expects or requires -- so I plan to pull this part of the tree out after the example is created.

xjrc:
@cyrush: I hadn't considered this when I was writing up that fix, but you have a good point. I was planning to work on Conduit this afternoon; do you want me to take another stab at it, or are you set on doing it instead?

cyrush:
@xjrc I already have a bunch of changes in flight so I'll take care of it.

Issue #548, opened 2020-04-22
schema:to_json, do we even need the detailed param?


cyrush:
Looking at the source code, this param doesn't appear to be used in the Schema class.

Seems to have propagated from the signature from Node?

cyrush:
resolved in #549

Issue #538, opened 2020-04-10
to_yaml on leaf includes newline


cyrush:
ponder if this should be the case:

```
>>> n = conduit.Node()
>>> n.set(42.0)
>>> n.to_yaml()
'42.0\n'
>>> n.to_json()
'42.0'
```

cyrush:
resolved in #549

Issue #403, opened 2019-04-09
use start after and end before in sphinx source and output includes


cyrush:
instead of exact line numbers

cyrush:
resolved in #549

PR #549, opened 2020-04-22
add to_string methods to core classes

### LINKED ISSUE #539: add Node:to_string() and change print() behavior to use yaml
1) 
Node::to_string() this would be an alias to `to_yaml()`, pending resolution to  #538 

make sure to document that that is the case in the header docs

2) 
 move print() to use to_yaml(), to respect humans

cyrush:
resolves #539


cyrush:
also resolves #538

cyrush:
also resolves #403

cyrush:
also resolves #548

cyrush:
also resolves #551

PR #558, opened 2020-05-07
Document expressions

markcmiller86:


markcmiller86:
This is in relation to https://github.com/visit-dav/visit/pull/4684

cyrush:
Thanks Mark - these docs look great!

We don't really talk about the blueprint index in the conduit docs, which is where these docs would ideally sit.  Any thoughts on how to make that clear?

The Mesh Blueprint docs page is getting big, maybe thoughts about this could also help with a reorg to make them easier to consume.

markcmiller86:
Ah, ok, so I've documented the *index* in a place where really the raw data files are being documented. If there is not currently a place where the blueprint index is documented, maybe just tuck this somewhere safe until such a place exists to stick it?

cyrush:
@markcmiller86 I think it would be super helpful to have this in the docs now, maybe move
to as part of:
https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#outputting-meshes-for-visualization

Note: PR #527  is going to change the docs a bit to improve the TOC tree.  I plan to merge soon -- after that merge we (or I) can take a pass to move

markcmiller86:
@cyrush ...so I think I am confused because it *was* that *exact* part of the manual I thought *was* changing with this PR. Now, I didn't know for sure but when I saw `blueprint_mesh.html` in RTD's URL for that page, I hit the `Edit this page on GitHub` button and it took me to the page that I modified for this PR. Am I missing something?

cyrush:
@markcmiller86 sorry for the confusion, I am suggesting a specific section of the blueprint_mesh docs, the one called:

```
Outputting Meshes for Visualization
====================================
```

https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#outputting-meshes-for-visualization

(It's hard to link to directly to a line number via github b/c of the markdown preview.)

This section doesn't yet mention the Blueprint index, but I think it's the best place to do so in the future, since it talks about how to write the data out for use in tools like VisIt.

markcmiller86:
Ah, Ok, I see. Just did that. Also, regarding linking to a specific line...sheesh just tried myself a few different ways and nothing seemed to work. Seems like a GitHub deficiency. Anyways, I've moved it now.

cyrush:
Awesome thanks! Yes, I fumbled with the linking as well, no fun!

PR #560, opened 2020-05-11
fix bugs in 'blueprint::mesh::specset::verify'

xjrc:
The changes in this pull request fix the `blueprint::mesh::specset::verify` function so that it accepts species information in line with what's described in [the Blueprint documentation](https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#species-sets) instead of strictly `mlarray`-style specifications, as was done previously. This fix allows for specifications like the following to be accepted:

```
specset:
  matset: mesh_mats
  matset_values:
    copper:
      cu: [ ... ]
      zn: [ ... ]
    steel:
      fe: [ ... ]
      cr: [ ... ]
      ni: [ ... ]
      mn: [ ... ]
```

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -4039,17 +4039,56 @@ mesh::field::index::verify(const Node &field_idx,
 //-----------------------------------------------------------------------------
 bool
 mesh::specset::verify(const Node &specset,
-                    Node &info)
+                      Node &info)
 {
     const std::string protocol = "mesh::specset";
     bool res = true;
     info.reset();
 
-    res &= verify_string_field(protocol, specset, info, "matset");
-    res &= verify_mlarray_field(protocol, specset, info, "matset_values", 2, 2);
     // TODO(JRC): Enable 'volume_dependent' once it's confirmed to be a required
     // entry for specsets.
     // res &= verify_enum_field(protocol, specset, info, "volume_dependent", mesh::booleans);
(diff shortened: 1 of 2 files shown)


src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -4039,17 +4039,56 @@ mesh::field::index::verify(const Node &field_idx,
...
     const std::string protocol = "mesh::specset";
     bool res = true;
     info.reset();
 
-    res &= verify_string_field(protocol, specset, info, "matset");
-    res &= verify_mlarray_field(protocol, specset, info, "matset_values", 2, 2);
     // TODO(JRC): Enable 'volume_dependent' once it's confirmed to be a required
     // entry for specsets.
     // res &= verify_enum_field(protocol, specset, info, "volume_dependent", mesh::booleans);
+    res &= verify_string_field(protocol, specset, info, "matset");
```
kennyweiss, line 4051: Note: `&=`  is an assignment for "bitwise and", not "logical and" and is not strictly equivalent to:

```cpp
bool b = some_predicate();
b = b && some_other_predicate();
```

Perhaps the biggest difference is related to short-circuiting, so it is probably ok here (and perhaps desirable if you want to list all problems, as opposed to just the first one).

xjrc: Yes, the intent is not to short-circuit so we can provide the user with a full list of the problems we've detected! The potential ambiguity with the short-circuit behavior isn't great, but I really like the terseness of this syntax so I think it's worth it.

Issue #567, opened 2020-06-03
add property queries for new types of matsets


xjrc:
In order to make it easier for codes to identify the different [material set variants](https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#material-sets), the `conduit::blueprint::mesh::matset` namespace should be extended to include new variant predicate functions. The following functions would be useful:

- `mesh::matset::is_multi_buffer`
- `mesh::matset::is_uni_buffer`
- `mesh::matset::is_element_dominant`
- `mesh::matset::is_material_dominant`

PR #563, opened 2020-05-26
add material-dominant relationship support to blueprint matsets

xjrc:
The changes in this pull request address #512  by making the following changes:

- [x] Update the `blueprint::mesh::matset::verify` function to recognize material-dominant schemas.
- [x] Implement test cases to validate the correctness of the new `blueprint::mesh::matset::verify` functionality.
- [x] Update the ["Material Sets"](https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#material-sets) documentation to include details for the new material-dominant option.
- [x] Update the `blueprint::mesh::examples::venn` regressions to all validate their results with `blueprint::mesh::verify`.

PR #571, opened 2020-06-15
add property queries for new types of matsets

xjrc:
The changes in this pull request directly address #567 by adding property query functions to the `blueprint::mesh::matset` namespace corresponding to the new material set types introduced in #553 and #563.


xjrc:
Thanks, @agcapps! @cyrush: Do the names/locations of these nmew property functions look reasonable? If so, I will merge this PR.

PR #578, opened 2020-06-17
Node: set_external(const T*)

### LINKED ISSUE #577: Conduit Node: set_external T const*
At least for floating point types, conduit lacks overloads for `T const*` types.

Currently seens as compile error for particles in AMReX:
`Src/Extern/Conduit/AMReX_Conduit_Blueprint_ParticlesI.H`

https://github.com/AMReX-Codes/amrex/blob/5ed64778cdca5134248b3055db037a4f25a14a8d/Src/Extern/Conduit/AMReX_Conduit_Blueprint_ParticlesI.H#L69-L70

ax3l:
Fix #577: Adds const-correctness to passed pointer data. This is mainly needed since users might pass a `const T*` that cannot be casted to a `T*` (but vice versa is possible). Also expresses actual (read-only) intent in the API :)

While on it, this improves general [const-correctness](https://isocpp.org/wiki/faq/const-correctness) for related overloads of the interface.

### DIFF
--- a/src/libs/conduit/conduit_generator.cpp
+++ b/src/libs/conduit/conduit_generator.cpp
@@ -196,7 +196,7 @@ public:
 
     static void    walk_json_schema(Node   *node,
                                     Schema *schema,
-                                    void   *data,
+                                    const void *data,
                                     const rapidjson::Value &jvalue,
                                     index_t curr_offset);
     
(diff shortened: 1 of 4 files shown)


cyrush:
thanks for taking a pass at this @ax3l, here is some more context:

We originally skipped const qualified `set_external` since `set_external` supports the a zero copy case which complicates the utility of const. 

While we can do somethings -- we can't enforce const-ness in actions done to the data after the `set_external` call. This is especially true across languages (python and fortran). 
That const use case is straight forward for the standard `set` cases, b/c the semantics imply a copy there. 

Happy to improve here, but this might not be straight forward give the `set_external` use case?

ax3l:
Thank you for the context @cyrush!

I think at least for the C++11 interface that `T const*` can be done and expresses exactly what you describe, including zero-copy data access. I'll give it a few more tries, I feel I got quite far already without major road blocks :)

Issue #504, opened 2020-02-13
refactor polygonal/polyhedral layout


xjrc:
In order to improve interoperability with client codes, the data layout for Conduit-Blueprint's polygonal/polyhdral elements needs to be updated. Specifically, the layout needs to be changed from an in-line, stream-based sequence to an out-of-line, offset-based collection of sequences. This can be accomplished by adopting Conduit's new one-to-many schema (#511) to extract the polygon sizes from the index stream.

#### Polygonal Variables ####

- `eI`: number of vertices in 2D element I
- `eIvJ`: the index of vertex J in element I

#### Current Layout ####

```
topology:
  coordset: coords
  type: unstructured
  elements:
    shape: polygonal
    connectivity: [e1, e1v1, ..., e1ve1, e2, e2v1, ..., e2ve2, ...]
```

#### New Layout ####

```
topology:
  coordset: coords
  type: unstructured
  elements:
    shape: polygonal
    connectivity: [e1v1, ..., e1ve1, e2v1, ..., e2ve2, ...]
    sizes: [e1, e2, e3, ...]
    offsets: [0, e1, e1+e2, ...]
```

This situation changes slightly in 3D because the top-level, out-of-line `begins` sequence does not fully describe the internal structure of an arbitrary element (e.g. how the faces that comprise the 3D element are laid out). For such cases, an extra offset-based sequence needs to be generated.

#### Polygonal Variables ####

- `eI`: number of sub-elements in 3D element I
- `eIsJ`: the index of sub-element (face) J of 3D element I
- `sIvJ`: the index of vertex J in sub-element (face) I 
- `|eIsJ|`: number of vertices in sub-element (face) J of 3D element I
- `eI_sJ_vK`: the index of vertex K in sub-element J in element I

#### Current Layout ####

```
topology:
  coordset: coords
  type: unstructured
  elements:
    shape: polyhedral
    connectivity: [e1, |e1_s1|, e1_s1_v1, ..., e1_s1_vs1, ..., |e1_se1|, e1_se1_v1,  ..., e1_se1_vse1, e2, ...]
```

#### New Layout ####

```
topology:
  coordset: coords
  type: unstructured
  elements:
    shape: polyhedral
    connectivity: [e1s1, ..., e1se1, e2s1, ..., e2se2, ...]
    sizes: [e1, e2, e3, ...]
    offsets: [0, e1, e1+e2, ...]
  subelements:
    shape: polygonal
[...]

PR #579, opened 2020-06-24
Refactor Polygonal / Polyhedral Layout

bmhan12:
This PR addresses #504 .


bmhan12:
@xjrc  @cyrush  I think this is ready for review.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -3145,12 +3279,53 @@ mesh::topology::unstructured::verify(const Node &topo,
             {
                 elems_res &= verify_integer_field(protocol, topo_elems, info_elems, "offsets");
             }
```
xjrc, line 3281: The "offsets" field has been rolled into the one-to-many relations construct and thus no longer needs to be separately checked here.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -3145,12 +3279,53 @@ mesh::topology::unstructured::verify(const Node &topo,
...
+            if(topo_elems["shape"].dtype().is_string() &&
+              (topo_elems["shape"].as_string() == "polygonal" ||
+               topo_elems["shape"].as_string() == "polyhedral"))
+            {
+                elems_res &= verify_integer_field(protocol, topo_elems, info_elems, "sizes");
+                elems_res &= verify_integer_field(protocol, topo_elems, info_elems, "offsets");
```
xjrc, line 3287: I don't think that these checks are necessary as they're covered by the one-to-many relationship schema check (i.e. `blueprint::o2mrelation::verify`). It is true that the schema check doesn't require that "sizes" and "offsets" exist and it's fairly senseless for them to be absent in this context, but our abstract specification requires satisfying only one-to-many relationship adherence and I feel like we should stick to the abstract spec as much as possible.

While that check doesn't explicitly require "sizes" and "offsets" variables, the schema doesn't require these either (polygonal/polyhedral are described as compliant with one-to-many relationships)

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -3145,12 +3279,53 @@ mesh::topology::unstructured::verify(const Node &topo,
...
+                        {
+                            subelems_res = false;
+                        }
+                        else
+                        {
+                            subelems_res &= verify_field_exists(protocol, topo_subelems, info_subelems, "shape") &&
+                                   mesh::topology::shape::verify(topo_subelems["shape"], info_subelems["shape"]);
+                            subelems_res &= verify_integer_field(protocol, topo_subelems, info_subelems, "connectivity");
+                            subelems_res &= verify_integer_field(protocol, topo_subelems, info_subelems, "sizes");
+                            subelems_res &= verify_integer_field(protocol, topo_subelems, info_subelems, "offsets");
```
xjrc, line 3313: Ditto on removing the separate "sizes"/"offsets" checks here.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -3174,6 +3349,64 @@ mesh::topology::unstructured::verify(const Node &topo,
                     chld_res &= verify_integer_field(protocol, chld, chld_info, "offsets");
                 }
 
+                // Polygonal & Polyhedral shape
+                if(chld.has_child("shape") && 
+                   chld["shape"].dtype().is_string() &&
+                   (chld["shape"].as_string() == "polygonal" || 
+                   chld["shape"].as_string() == "polyhedral"))
+                {
```
xjrc, line 3357: I know that there was a problem in the original code here wherein some 10+ lines were duplicated between the single shape case and the mixed shape case, but persisting this problem to ~30 lines exacerbates it. Would it be possible to abstract this behavior into some form of helper function to cut down on the code duplication?

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2193,11 +2192,11 @@ basic(const std::string &mesh_type,
...
-    Node &topo = res["topologies"].child(0);
-    if(topo.has_child("elements") && topo["elements"].has_child("offsets"))
-    {
-        topo["elements"].remove("offsets");
-    }
+    // Node &topo = res["topologies"].child(0);
+    // if(topo.has_child("elements") && topo["elements"].has_child("offsets"))
+    // {
+    //     topo["elements"].remove("offsets");
+    // }
```
xjrc, line 2199: If this code is no longer necessary, the code and its associated comment should be removed.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -3707,26 +4015,62 @@ mesh::topology::unstructured::generate_corners(const Node &topo,
...
+                    index_t face_index = subconn_data_raw.size() / corners_face_degree;
+
+                    // Check if face already exists
+                    for (index_t face_i = 0; face_i < face_index; face_i++ )
+                    {
+                        index_t face_off = face_i * corners_face_degree;
+                        has_perm |= std::is_permutation(subconn_data_raw.begin() + face_off,
+                                                        subconn_data_raw.begin() + face_off + 
+                                                        corners_face_degree,
+                                                        face.begin());
```
xjrc, line 4052: Very cool; I didn't know `std::is_permutation` even existed!

src/tests/blueprint/t_blueprint_mesh_transform.cpp
```diff
@@ -452,37 +456,90 @@ TEST(conduit_blueprint_mesh_transform, polygonal_transforms)
...
+                        &poly_subdata[poly_data[efo] * topo_findices],
+                        &poly_subdata[poly_data[efo] * topo_findices + topo_findices]);
+                        poly_index_set = polyhedral_index_set;
+                    }
+                    else
+                    {
+                        std::set<index_t> polygonal_index_set(
+                        &poly_data[efo],
+                        &poly_data[efo + topo_findices]);
+                        poly_index_set = polygonal_index_set;
```
xjrc, line 541: Please indent the contents of this conditional and the previous conditional to make the code easier to follow.

src/tests/blueprint/t_blueprint_mesh_verify.cpp
```diff
@@ -750,6 +850,35 @@ TEST(conduit_blueprint_mesh_verify, topology_unstructured)
...
+
+        n["subelements"]["e"]["shape"].set("polygonal");
+        CHECK_MESH(verify_unstructured_topology,n,info,false);
+        n["subelements"]["e"]["connectivity"].set(DataType::int32(6));
+        CHECK_MESH(verify_unstructured_topology,n,info,false);
+        n["subelements"]["e"]["offsets"].set(DataType::int32(3));
+        CHECK_MESH(verify_unstructured_topology,n,info,false);
+        n["subelements"]["e"]["sizes"].set(DataType::int32(3));
+        CHECK_MESH(verify_unstructured_topology,n,info,true);
+
```
xjrc, line 881: Thanks for doing such thorough testing of the new schemas and for filling the gaps in mixed shape topology testing!

src/libs/blueprint/conduit_blueprint_mesh.cpp
bmhan12: Ah, that makes sense, thank you for clarifying! I did see the optional checking for "sizes" and "offsets" in `o2mrelation::verify`, but wasn't sure if that extended to polygonal/polyhedral.

src/libs/blueprint/conduit_blueprint_mesh.cpp
bmhan12: Yes, I've added a helper function,`verify_poly_node()` that takes care of verifying for polygonal/polyhedral in the mixed and single shape case.

src/libs/blueprint/conduit_blueprint_mesh.cpp
bmhan12: I think this "offsets" check here is fine, for the case where a non-polygonal/non-polyhedral shape has the offsets optionally generated.

xjrc: My understanding was that we were going to get rid of the "offsets" field for non-polygonal/polyhedral topologies since it was redundant and could consequently cause inconsistencies (e.g. a "quad" topology with an "offsets" array of `[0, 4, 8, 11, ...]`. This is actually why I removed the "topology/connectivity/offsets" section from [the documentation](https://github.com/LLNL/conduit/pull/579/files#diff-2aaf2cc46f2369eddb166bf2d13ed654L414).

@cyrush: I'm curious to hear your thoughts since my understanding was informed by a discussion between you and @agcapps about [the interaction between "shape" and "sizes"](https://github.com/LLNL/conduit/issues/511#issuecomment-597341371). The outcome of that discussion seemed to be that some "shape" specifiers made "sizes"  redundant, so they shouldn't be used in those cases. Given that it's a derived/related quantity, do we want to keep the "offsets" field for potentially faster indexing on static-shaped topologies (e.g. "tris", "quads")?

cyrush: `shape` vs `sizes` isn't good b/c they could both be present and conflict. But `offsets` could be useful in the non poly case, b/c it would allow folks to zero copy via an indirection array.  Blueprint consumers will need to be upgraded to support this -- but it seems like a good thing?

src/libs/blueprint/conduit_blueprint_mesh.cpp
cyrush: that is cool, it's c++11 and forward. We still have an option for older C++, but I think for the next release we should make C++11 support min

src/tests/docs/t_conduit_docs_blueprint_demos.cpp
```diff
@@ -614,7 +616,16 @@ TEST(conduit_docs, blueprint_demo_basic_polyhedra)
...
+            "sizes": [6, 6, 6, 6, 6, 6, 6, 6],
+            "offsets": [0, 6, 12, 18, 24, 30, 36, 42]
+          },
+          "subelements":
+          {
+            "shape": "polygonal",
+            "connectivity": [0, 1, 4, 3, 0, 1, 10, 9, 1, 4, 13, 10, 4, 3, 12, 13, 3, 0, 9, 12, 9, 10, 13, 12, 1, 2, 5, 4, 1, 2, 11, 10, 2, 5, 14, 11, 5, 4, 13, 14, 10, 11, 14, 13, 3, 4, 7, 6, 4, 7, 16, 13, 7, 6, 15, 16, 6, 3, 12, 15, 12, 13, 16, 15, 4, 5, 8, 7, 5, 8, 17, 14, 8, 7, 16, 17, 13, 14, 17, 16, 9, 10, 19, 18, 10, 13, 22, 19, 13, 12, 21, 22, 12, 9, 18, 21, 18, 19, 22, 21, 10, 11, 20, 19, 11, 14, 23, 20, 14, 13, 22, 23, 19, 20, 23, 22, 13, 16, 25, 22, 16, 15, 24, 25, 15, 12, 21, 24, 21, 22, 25, 24, 14, 17, 26, 23, 17, 16, 25, 26, 22, 23, 26, 25],
+            "sizes": [4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4],
+            "offsets": [0, 4, 8, 12, 16, 20, 24, 28, 32, 36, 40, 44, 48, 52, 56, 60, 64, 68, 72, 76, 80, 84, 88, 92, 96, 100, 104, 108, 112, 116, 120, 124, 128, 132, 136, 140]
           }
```
cyrush, line 629: new setup is so much better than wall of implicitly packed indices!

cyrush:
and thanks @xjrc for your shaping of this as well!

src/libs/blueprint/conduit_blueprint_mesh.cpp
xjrc: I agree that greater flexibility is generally desirable, particularly in cases like these when it comes at a low conceptual cost. However, there are two complications I foresee if we allow for a non one-to-many `offsets` concept in this context:

1. If `offsets` is allowed to exist in contexts where `sizes` is disallowed, then we need a more fundamental form of schema infrastructure below one-to-many relationships that codifies this new concept (e.g. indirection array). This concept would likely just use `indices` instead of singular `offsets` as the result is identical and it would allow for a cleaner split between `sizes`/`offsets` and `indices` (which [already exists in one-to-many relationships](https://llnl-conduit.readthedocs.io/en/latest/blueprint_o2mrelation.html#protocol)).
2. If `offsets` indirection is allowed for topological connectivity arrays, why isn't it allowed for all the other Blueprint arrays? For example, couldn't a mesh's positions/fields/species also exist as non-linearly indirected entries in larger arrays? If we decide to support such broad applications of indirection, we'd need to expand our infrastructure in a concerted way.

With these two things in mind, I don't see it being feasible to have `offsets` in this particular PR (at least if we want to remain consistent with the one-to-many relationship concept), but we could add it in the future if the infrastructure is required.

cyrush: Great points, yes let's think about this more and address in the future, not in this PR.

bmhan12: I've removed the `offsets` check in 509235f.

PR #517, opened 2020-03-10
Update Uberenv and introduce Gitlab CI [feature/uberenv_ci]

adrienbernede:
This is a transfer of #502 from fork to in-repo.
(PRs issued form a fork can’t trigger GitLab CI.)

Goals:
- Update uberenv to preserve workflow and fix some stuff.
- Add Gitlab CI capability on LC systems.

PR #582, opened 2020-07-03
Add CI on LC Gitlab

adrienbernede:
This PR is the "Add CI" part of #517.

Uberenv has been only slightly modified to fix a bug with mirroring.

The spack configuration has been updated based on Ascent one.

The failing jobs in CI are marked as "Allowed failure" when the cause is identified as beyond the scope of this PR.

### DIFF
--- a/scripts/uberenv/spack_configs/config.yaml
+++ b/scripts/uberenv/spack_configs/config.yaml
@@ -120,7 +120,7 @@ config:
   # - If set to 16 on a 4 cores machine `spack install` will run `make -j4`
   # - If set to 16 on a 18 cores machine `spack install` will run `make -j16`
   # If not set, Spack will use all available cores up to 16.
-  build_jobs: 8
+  build_jobs: 64
 
 
   # If set to true, Spack will use ccache to cache C compiles.
--- a/scripts/uberenv/spack_configs/config.yaml
+++ b/scripts/uberenv/spack_configs/config.yaml
@@ -120,7 +120,7 @@ config:
   # - If set to 16 on a 4 cores machine `spack install` will run `make -j4`
   # - If set to 16 on a 18 cores machine `spack install` will run `make -j16`
   # If not set, Spack will use all available cores up to 16.
-  build_jobs: 64
+  build_jobs: 8
 
 
   # If set to true, Spack will use ccache to cache C compiles.
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,7 +5,7 @@
 
 variables:
   GIT_SUBMODULE_STRATEGY: recursive
-  INSTALL_ROOT: ${CI_BUILDS_DIR}/conduit/${CI_COMMIT_REF_SLUG}_${CI_PIPELINE_ID}/uberenv_libs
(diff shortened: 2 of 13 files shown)


adrienbernede:
LGTM

scripts/uberenv/spack_configs/config.yaml
```diff
@@ -120,7 +120,7 @@ config:
   # - If set to 16 on a 4 cores machine `spack install` will run `make -j4`
   # - If set to 16 on a 18 cores machine `spack install` will run `make -j16`
   # If not set, Spack will use all available cores up to 16.
-  build_jobs: 8
+  build_jobs: 64
```
cyrush, line 123: this sets the max for all builds, if you can also change per platform by adding a config.yaml

adrienbernede: I changed it to take advantage of the full node allocated to build, on both machines.
Yes, it’s the "max". So, if less cores are available (quartz 36), build will use all cores (-j 36), not 64.
Isn’t that OK of every machine?

adrienbernede:
> I really like the reorganized CI scripts! Thanks!

Thanks! I’m glad you like it!

scripts/uberenv/spack_configs/config.yaml
cyrush: All cores would be bad if if someone builds on a login node.

adrienbernede: oops

adrienbernede: Limiting the number of cores is a shame in CI context. Since someone could potentially build on a login node on any machine, should I go back to 8 cores? I’m not even sure a process using all cores would be allowed on a login node.

Could we "encourage" people to build on allocated resource? I can document it.

cyrush: In other CI contexts (azure, travis) we have very limited resources, in other projects we have run out of ram even when using multi threaded builds. 

You can always use a custom config for these CI tests?

adrienbernede: I see your point, and here is where I am in my reflexions:
- Definitely, 64 should no be applied to all contexts, as is the case right now
  -> Use specific config for quartz and lassen
- _But_, even on quartz and lassen, the 64 is not desirable when building on login node
  -> Either no build on login node, or patch config.yaml on the fly for CI.

I we choose to patch on the fly, we don’t even need a specific config for lassen and quartz.

Lastly, I don’t see another way to customize spack config for CI context, because spack config is determined by uberenv using the machine info "only".

cyrush: but the ci script is always running CI, so it could select a config-ci.yaml, or something?

Not picky about the CI solution, but i do want to avoid 64 (max) for all configs.

adrienbernede:
LGTM

scripts/uberenv/spack_configs/config.yaml
adrienbernede: Done.

adrienbernede:
Failures are due to fetching instabilities in Spack. Failing tests passed with previous commit, and change introduced have no impact outside Gitlab CI, which passed.
Merging.

cyrush:
thanks!

PR #584, opened 2020-07-09
Update uberenv

adrienbernede:
It was decided that #517 should be split in two:

- Introducing gitlab CI (done in #582)
- Updating Uberenv

This PR updates Uberenv with a more recent version, fixing some bugs and allowing to get rid of "fake packages".


adrienbernede:
LGTM

scripts/uberenv/spack_configs/config.yaml
```diff
@@ -112,7 +126,7 @@ config:
   # of Spack.  For filesystems that don't support locking, you should set
   # this to false and run one Spack at a time, but otherwise we recommend
   # enabling locks.
-  locks: false
+  locks: true
```
cyrush, line 129: I worry about this one being true, it has caused hangs in the past

cyrush:
I am getting a source mirror up and running in #585, after this seems to be working, then I can try to help with updating uberenv.

scripts/uberenv/spack_configs/config.yaml
adrienbernede: Lot of change with lock management is Spack since.
I would suggest giving it a chance. I’ve had issues with this _not_ being set to true.

adrienbernede: It’s is true that this has "only" be tested by me locally and in CI.
So I understand that merging this is less straight-forward than merging CI only.

cyrush: Ok, lets keep them on 
I guess you may actually need it for the CI workflows that have shared spack instances? 

In the past locks weren't necessary and could be a barrier (hangs) for single install. But if we are going to use shared upstreams, we need them.

adrienbernede: https://github.com/LLNL/conduit/blob/9d654df2a9f93aad23f623a991826d20573ecac9/scripts/ci/build-on-lc.sh#L39

Prefix still depends on the toolchain, I wanted to avoid concurrency at first, but it’s a potential optimization.

cyrush:
closing since #684 will accomplish this with uberenv as a submodule.

Issue #570, opened 2020-06-06
relay io handle - support read with empty path


cyrush:
`h.read("",node)`

should be equiv to:

`h.read(node)`

but that doesn't appear to be the case, at least for an IOHandle reading hdf5 files.

cyrush:
resolved in #588

Issue #559, opened 2020-05-08
relay::blueprint::save_mesh in lieu of relay::io_blueprint::save?


cyrush:
@xjrc not sure if I ran this by you yet, but I am thinking we could deprecate:

`relay::io_blueprint::save`

in favor of something like:

`relay::blueprint::save_mesh`

I recall the reason for `io_blueprint` was to avoid confusing issues when using the namespaces (we don't want `relay::blueprint::save` and `relay::save` to be ambiguous -- I think   `relay::blueprint::save_mesh` strikes a better balance. What do you think?

xjrc:
Approved!

cyrush:
Cool thanks! I'll tackle this in the near future + leave the current there as deprecated between 0.6.0 and 0.7.0

cyrush:
resolved in #588

PR #588, opened 2020-07-24
adds support for reading sidre datastore style files

cyrush:


src/libs/relay/conduit_relay_io_handle.cpp
```diff
@@ -921,7 +871,14 @@ IOHandle::read(const std::string &path,
...
-        m_handle->read(path, node);
+        if(path.empty())
+        {
+            m_handle->read(node);
+        }
+        else
+        {
+            m_handle->read(path, node);
+        }
     }
```
cyrush, line 882: resolves #570 , but we also need to add an extra unit test 

- [ ] add that test

src/tests/relay/t_relay_io_handle_sidre.cpp
```diff
@@ -0,0 +1,217 @@
...
+    res = utils::join_path(res,"data");
+    return utils::join_path(res,test_fname);
+}
+
+//-----------------------------------------------------------------------------
+TEST(conduit_relay_io_handle, test_sidre_with_root)
+{
+    return;
+    io::IOHandle h;
+    h.open(relay_test_data_path("braid_3d_sidre_examples.root"),
```
cyrush, line 74: these will fail on CI, want to swap out for a smaller example file

cyrush:
resolves #559

cyrush:
@xjrc  lots here, I don't want to burden you with a review, but FYI, this will implement the change from io_blueprint::save, to io::blueprint::save_mesh, etc

xjrc:
Thanks for the heads up; I'll give this a review in the next couple of days. Just so I know, is this still a WIP or is it pretty much finished with your latest batch of changes?

cyrush:
this batch of changes are finish. I still expect some rapid changes after this PR, including adding support to read meshs from sidre style hdf5 files. Again, I don't think you need to review, just wanted to keep you in the loop.

src/libs/relay/conduit_relay_io_handle_sidre.cpp
```diff
@@ -0,0 +1,1273 @@
...
+    {
+        if(p_first == "root")
+        {
+            read_from_root(p_next,node);
+        }
+        else
+        {
+            if(!utils::string_is_integer(p_first))
+            {
+                // TODO:ERROR!
```
cyrush, line 267: - [ ] add error here

src/libs/relay/conduit_relay_io_identify_protocol.cpp
```diff
@@ -143,6 +143,20 @@ identify_protocol(const std::string &path,
...
+void
+identify_file_type(const std::string &path,
+                  std::string &file_type)
+{
+    // default to unknown
+    file_type = "unknown";
+
+    // TODO: HARD WIRED!
+    file_type = "hdf5";
+
```
cyrush, line 156: - [ ] bring load mesh heuristic logic here

cyrush:
- [ ] add unit test for identify protocol

cyrush:
merging in this batch, so I can move on to reading blueprint meshes from sidre

PR #598, opened 2020-08-10
Added C++-style iterator

estebanpauli:
I did this as part of the hackathon last week. It provides a C++ style iterator for the Node class. Please feel free to use none/any/all of this depending on how well it fits in with the current design and goals. I'm also happy to make adjustments as needed.


cyrush:
Awesome! Would it be possible to move the iterator methods into the existing iterator classes -- do yousee any roadblocks to do so?

estebanpauli:
> Awesome! Would it be possible to move the iterator methods into the existing iterator classes -- do yousee any roadblocks to do so?

I hadn't thought about that before. Should have done a design review first :)

The main potential problem I see is that currently ConduitNode is not a class template like the new iterator. To not break existing code, we would have to extract all of its behavior to a parent class (e.g. new new template) and then provide a typedef like "using NodeIterator = node_iterator_template<false>". We could also provide "using NodeConstIterator = node_iterator_template<true>".

estebanpauli:
Looking at some of the build failures, it looks like conduit does not support C++11 features. Is this correct? If so, I'll have to rework some things.

cyrush:
We plan to make C++11 the min for the next release, I am fine with this being part of the push for that.

estebanpauli:
I still need to tackle combining the new iterators with the old ones. First I want to make sure all the tests pass in all the different configurations.

estebanpauli:
I did not mean to change BLT in the last commit. I will fix that.

src/libs/conduit/conduit_node_iterator.hpp
```diff
@@ -54,14 +54,228 @@
 //-----------------------------------------------------------------------------
 // -- conduit includes -- 
 //-----------------------------------------------------------------------------
-#include "conduit_node.hpp"
+#include "conduit_core.hpp"
```
cyrush, line 57: is there an issue with order of include?
(could this still be conduit_node, and class not be fowardwad declared below)?

cyrush (approved):
This looks good -- I agree breaking into another class is a better plan of attack. 

To finish off, can you 
* Add an entry to changelog
* Add a unit test that looks more like expected plain C++ usage
(Perhaps one with for loop with the iterator, or if C++11 support is on a foreach?) so we can point folks to those in the absence of other docs

src/libs/conduit/conduit_node_iterator.hpp
estebanpauli: `conduit_node.hpp` includes `conduit_node_iterator.hpp` since it returns instances of `NodeIterator` and `NodeConstIterator` by value. I broke the circular dependency by removing `conduit_node.hpp` here. I was about to write that we only needed a forward declaration of `Node` in this file. However, I'm not sure how that's true given that we use `Node&` and `Node*` as template parameters and then write expressions based on those parameters. Maybe this works because anyone working with `conduit_node_iterator.hpp` would also have `conduit_node.hpp`. I did make sure `node_iterator.hpp` is the first include in `node_iterator.cpp`, and that compiles fine (maybe because `conduit_node.hpp` is included before any uses).

After removing `conduit_node.hpp`, I had to add conduit_core.hpp for some other definitions (e.g. `conduit::index_t`).

Thoughts on this?

Do users include the individual `conduit_X.hpp` files, or do they just include `conduit.hpp`?

estebanpauli:
> This looks good -- I agree breaking into another class is a better plan of attack.
> 
> To finish off, can you
> 
>     * Add an entry to changelog
> 
>     * Add a unit test that looks more like expected plain C++ usage
>       (Perhaps one with for loop with the iterator, or if C++11 support is on a foreach?) so we can point folks to those in the absence of other docs

The "enhanced_for_loop" test at https://github.com/LLNL/conduit/pull/598/files#diff-e54a8c9e88d6d986e2051f2f28fb706aR287 tests enhanced for looks. Is that what you're looking for, a test with `std::for_each()`, or something different?

cyrush:
For the unit tests, it's good if some can also serve as a reference (or if we can provide examples in the docs)

The typed tests do a good job at putting things through their paces, but they aren't very readable. I think it woudl be good to just have a simpler test that doens't used as much google test magic that we can point folks to.  Does that make sense?

src/libs/conduit/conduit_node_iterator.hpp
cyrush: I took a deeper look to better understand.  This is fine b/c both the new iterator classes and the existing ones only use Node refs or Node pointers. 

If I add the following to the base of the new iterator class:

```
    // wont work
    Node WHATISGOINGON()
    {

    }

    // works ok
    Node *WHATISGOINGON_PTR()
    {

    }

```
The first one makes the compiler unhappy, and the second one is fine.
I think we are in good shape, I just wanted to understand what was at play.
Thank!

estebanpauli:
I'm not sure what to make of the test failure. If this looks spurious, could someone with permissions rerun it?

I also can't merge because I don't have write access. Once the tests are verified, if everything looks good, could someone merge this?

Thanks.

src/tests/conduit/t_conduit_node_cpp_iterator.cpp
```diff
@@ -0,0 +1,351 @@
...
+    parent.add_child("c2").set("value 2");
+    parent.add_child("c3").set("value 3");
+
+    // Ensure enhanced for loop works properly
+    index_t current_index = 0;
+    for (Node &child : parent.children()) {
+        EXPECT_EQ(&child, parent.child_ptr(current_index))
+                            << "Got wrong child at " << current_index;
+        ++current_index;
+    }
```
cyrush, line 76: Looks good!

cyrush:
merged, thanks for this!

PR #610, opened 2020-09-03
Fix include on case-sensitive-fs

jschueller:
It must be lowercase on case-sensitive file-systems, eg when cross compiling

### DIFF
--- a/src/libs/relay/CMakeLists.txt
+++ b/src/libs/relay/CMakeLists.txt
@@ -218,6 +218,9 @@ if(ZFP_FOUND)
     list(APPEND conduit_relay_deps zfp)
 endif()
 
+if (MINGW)
+  list(APPEND conduit_relay_deps ws2_32)
+endif ()
 
 add_compiled_library(NAME conduit_relay
                      EXPORT conduit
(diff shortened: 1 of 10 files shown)


src/libs/relay/CMakeLists.txt
```diff
@@ -218,6 +218,9 @@ if(ZFP_FOUND)
     list(APPEND conduit_relay_deps zfp)
 endif()
 
+if (WIN32)
+  list(APPEND conduit_relay_deps ws2_32)
```
cyrush, line 222: Looks like civetweb needs this dep for the minigw case?

Is there anyway we can detect this case more specifically, I am not sure we need it for general builds.

cyrush (commented):
Thanks for these fixes, I have a q: about the ws2_32 dep maybe we can explore more?

src/libs/relay/CMakeLists.txt
jschueller: ok, it seems to be needed only for mingw, I modified the code to use "if (MINGW)"

cyrush:
Great, thanks for that adjustment!

Issue #607, opened 2020-08-27
Incorrect docs explanation of set semantics for compatible nodes


SeifIbrahim:
The docs suggest that `a.set(b)` will not change the schema of `a` if it is compatible with the schema of `b` and will simply copy. This is not true if `a` and  `b` are nodes.

https://llnl-conduit.readthedocs.io/en/latest/tutorial_cpp_basics.html#compatible-schemas
https://llnl-conduit.readthedocs.io/en/latest/tutorial_python_basics.html#compatible-schemas
https://llnl-conduit.readthedocs.io/en/latest/tutorial_cpp_update.html#node-update-methods

cyrush:
resolved in #617

Issue #605, opened 2020-08-21
conduit python interface helpful tips


cyrush:
when walking nodes, use `fetch_existing() ` to avoid extreme frustration with misspelling path names.

cyrush:
resolved in #617

Issue #519, opened 2020-03-11
add generate_index example to docs


cyrush:
the all an one helper is handy, but it doesn't cover many situations

show some concrete generate_index examples -- maybe wait for multi-domain generate_index improvements.

cyrush:
we should add an mpi example that uses the blueprint mpi lib to generate the index, and also one that uses `mesh_save`

cyrush:
pushing to 0.7.0, still need this but new relay::io::blueprint::write_mesh and read_mesh alleviate immediate need.

cyrush:
We want folks to use the `io::blueprint::write_mesh`, lets make sure that is clear in the docs.

Issue #459, opened 2019-10-16
docs request


cyrush:
to improve tutorial:

* check  https://llnl-conduit.readthedocs.io/en/latest/tutorial_cpp_numeric.html for info about arrays and pointers

* add details about how useful Node refs are

cyrush:
resolved in #617

PR #617, opened 2020-09-16
batch of docs updates

cyrush:


cyrush:
resolves #519

cyrush:
TODO:
- [ ] #607 
- [ ] #611 
- [ ] #519

cyrush:
resolves #459

cyrush:
resolves #605

cyrush:
resolves #607

cyrush:
@xjrc @agcapps if you have time for a quick look, can you give these changes a once over. More pairs of eyes + comments could help me improve.

src/docs/sphinx/blueprint_mesh.rst
```diff
@@ -9,10 +9,19 @@ Mesh Blueprint
...
+
+
+Conduit docs don't have a Mesh Blueprint tutorial yet, if you are looking to 
+wrap your mind around the basic mechanics of describing a mesh:
+
+* The Ascent tutorial includes section on `creating Meshes using Conduit <https://ascent.readthedocs.io/en/latest/Tutorial_Intro_Conduit_Blueprint.html>`_. This is the best reference for getting started and includes C++ and Python code examples.
+
+* The :ref:`complete_uniform_example` at the end of this section shows you how to create and save a uniform grid to a file which VisIt and `Ascent's Replay utility <https://ascent.readthedocs.io/en/latest/Utilities.html?highlight=replay#replay>`_ can read.
+
+* The :ref:`examples` section details functions that to generate several flavors of exemplar meshes.
```
xjrc, line 24: These are great resources; thanks for including them, and thanks to you and/or @mclarsen for writing them!

src/docs/sphinx/t_conduit_docs_blueprint_demos_out.txt
```diff
@@ -31,7 +31,7 @@ fields:
 END_EXAMPLE("blueprint_demo_basic_uniform")
 [/Users/harrison37/Work/github/llnl/conduit/src/tests/docs/t_conduit_docs_blueprint_demos.cpp : 67]
  Testing Basic Example 'uniform'
-[       OK ] conduit_docs.blueprint_demo_basic_uniform (12 ms)
+[       OK ] conduit_docs.blueprint_demo_basic_uniform (10 ms)
```
xjrc, line 34: For the future, would it be possible to omit the timings for these test cases so that they aren't modified each time we update the example annotations in the documentation?

src/docs/sphinx/t_conduit_docs_blueprint_examples_out.txt
```diff
@@ -79,7 +79,7 @@ transformed to 'interleaved' mcarray
 }
 
 mem_spaces: 
-  0x7fc814e027e0: 
+  0x7f8890702090:
```
xjrc, line 82: Excluding this from the log would also be ideal as it will likely change each update. The easiest workaround probably involves just verifying the addresses with assert statements instead of printing them out.

src/docs/sphinx/t_conduit_docs_tutorial_python_basics_out.txt
```diff
@@ -9,115 +9,165 @@ Updating test list for fixtures
...
-111: 
-111: END_EXAMPLE("py_basics_bw_style")
-111: BEGIN_EXAMPLE("py_basics_bw_style_from_native")
-111: 
-111: {
-111:   "test": {"dtype":"int64","number_of_elements": 1,"offset": 0,"stride": 8,"element_bytes": 8,"endianness": "little"}
-111: }
-111: END_EXAMPLE("py_basics_bw_style_from_native")
-1/1 Test #111: t_conduit_docs_tutorial_python_basics ...   Passed    0.46 sec
+113: Test command: /Users/harrison37/Work/github/llnl/conduit/uberenv_libs/spack/opt/spack/darwin-mojave-x86_64/clang-10.0.0-apple/python-3.7.6-zd24bom2hns4o42jlzdptijz54ogai7c/bin/python3.7 "-B" "-m" "unittest" "-v" "t_conduit_docs_tutorial_python_basics"
```
xjrc, line 15: This is also pretty hard to validate; all of these lines got updated because the prefix was updated (i.e. `111` to `113`).

src/docs/sphinx/t_conduit_docs_tutorial_python_relay_mpi_examples_out.txt
```diff
@@ -9,102 +9,102 @@ Updating test list for fixtures
...
-116: [rank: 1] received: 
-116: a: 
-116:   data: 1.0
-116:   more_data: 2.0
-116:   b: 
-116:     my_string: "value"
-116: 
-116: [rank: 1] received: [0, 1, 4, 9]
-1/1 Test #116: t_conduit_docs_tutorial_python_relay_mpi_examples ...   Passed    0.48 sec
+118: Test command: /Users/harrison37/Work/github/llnl/conduit/uberenv_libs/spack/opt/spack/darwin-mojave-x86_64/clang-10.0.0-apple/mpich-3.3.1-kkqsshrtda4dgcnbbwxndombmgfsmp4f/bin/mpiexec "-n" "2" "/Users/harrison37/Work/github/llnl/conduit/uberenv_libs/spack/opt/spack/darwin-mojave-x86_64/clang-10.0.0-apple/python-3.7.6-zd24bom2hns4o42jlzdptijz54ogai7c/bin/python3.7" "-B" "-m" "unittest" "-v" "t_conduit_docs_tutorial_python_relay_mpi_examples"
```
xjrc, line 15: Same for all of the rest of these. I don't know what the best solution is here; can we just omit the line number from the output?

src/docs/sphinx/tutorial_cpp.rst
```diff
@@ -9,6 +9,8 @@ C++ Tutorial
 This short tutorial provides C++ examples that demonstrate the Conduit's Core 
 API. Conduit's unit tests (*src/tests/{library_name}/*) also provide a rich set 
 of examples for Conduit's Core API and additional libraries. 
+Ascent's Tutorial also provides a brief `intro to Conduit basics <https://ascent.readthedocs.io/en/latest/Tutorial_Intro_Conduit_Basics.html>`_
+with C++ and Python examples.
```
xjrc, line 13: As an aside question, would it make sense for us to co-opt this into the Conduit documentation proper? It seems like a useful set of demos that would be a great supplement to our introductory material.

src/docs/sphinx/tutorial_cpp_basics.rst
```diff
@@ -159,7 +163,65 @@ You can check if a Schema is compatible with another Schema using the **Schema::
...
+   :start-after: BEGIN_EXAMPLE("basics_node_refs_bad")
+   :end-before:  END_EXAMPLE("basics_node_refs_bad")
+
+
+const Nodes
+--------------------------------
+
+If you aren't careful, the ability to easily create dynamic trees can
+also undermine your process to consume them. 
+For example, asking for a expected but non-existent path will return
```
xjrc, line 204: There's a typo here; this should be:

```
For example, asking for a*n* expected but non-existent...
```

src/docs/sphinx/tutorial_cpp_basics.rst
```diff
@@ -159,7 +163,65 @@ You can check if a Schema is compatible with another Schema using the **Schema::
...
+a reference to an empty Node. Surprise!
+
+Methods like `fetch_existing` allow you to be more explicit
+when asking for expected data. In C++, ``const Node`` references are also common
+way to process trees in an read-only fashion. ``const`` methods will not modify
+the tree structure, so if you ask for a non-existent path, you will receive
+an error instead of reference to an empty Node.
```
xjrc, line 211: I get a lot of questions about this, so I'm very pleased to see this documentation; great addition!

xjrc (changes requested):
This looks great, @cyrush; I quite like all of the new additions. I have some questions that I left as review comments and a few typo suggestions, but I think we can merge once those are addressed. Going forward, it would be good if we could minimize our diffs for the generated pieces of documentation (e.g. `t_conduit_docs_blueprint_demos_out.txt`) so that only important changes show up (e.g. no line number changes, time changes, etc.), but we can worry about that another time. Thanks for making all of these improvements!

src/docs/sphinx/t_conduit_docs_tutorial_python_relay_mpi_examples_out.txt
cyrush: I agree there is a lot of noise here. But those parts of the files aren't pulled into the docs, where other parts are automatically pulled in (via tags).  It is a bit annoying to skip as part of the review, but trying to avoid the changes in the first place would be hard, i worry that creating system to do so could cause surprises.

src/docs/sphinx/blueprint_mesh.rst
```diff
@@ -1254,59 +1263,101 @@ Outputting Meshes for Visualization
...
+
+    /// Options accepted via the `opts` Node argument:
+    ///
+    ///      mesh_name: "{name}"
+    ///          provide explicit mesh name, for cases where bp data includes
+    ///           more than one mesh.
+    ///
+    conduit::relay::io::blueprint::read_mesh(const std::string &root_file_path,
+                                             const conduit::Node &opts,
+                                             conduit::Node &mesh);
```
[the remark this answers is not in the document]
cyrush: I have a habit of using input nodes as the first arg, and output nodes as the final arg. 

That said, I don't think I am completely consistent.  This also causes confusion with the python interface, where since there are kwargs and differente considerations related to those, the order can be differente between the C++ and Python  interfaces.

src/docs/sphinx/blueprint_mesh.rst
```diff
@@ -1254,59 +1263,101 @@ Outputting Meshes for Visualization
...
+.. code:: cpp
+
+    /// Options accepted via the `opts` Node argument:
+    ///
+    ///      file_style: "default", "root_only", "multi_file"
+    ///            when # of domains == 1,  "default"   ==> "root_only"
+    ///            else,                    "default"   ==> "multi_file"
+    ///
+    ///      suffix: "default", "cycle", "none" 
+    ///            when # of domains == 1,  "default"   ==> "off"
```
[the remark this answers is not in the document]
cyrush: Good catch, I'll figure this out

- [ ] resolve this mystery

src/docs/sphinx/t_conduit_docs_blueprint_examples_out.txt
cyrush: For this case, it is used in the docs to demo the differences between how the memory is alloced by the nodes.
I don't think we can get away with hiding it.

(https://llnl-conduit.readthedocs.io/en/latest/blueprint.html#top-level-blueprint-interface)


Side Note: Not too happy with the schema formatting in many of these examples either, but thats a side effect of a prior change.

src/docs/sphinx/t_conduit_docs_tutorial_python_relay_mpi_examples_out.txt
xjrc: Could we just trim all non-tagged segments of these files, or is there something important we want to retain in the interim segments?

src/docs/sphinx/tutorial_cpp.rst
cyrush: Yes, I think we should think about developing a more comprehensive tutorial, this content can help.

src/docs/sphinx/t_conduit_docs_tutorial_python_relay_mpi_examples_out.txt
cyrush: I am sure we could create a solution. But there are bound to be corner cases, they scare me. 

Adding the script to regen these was a huge improvement and lowered the barrier big time. Any clean up could be processed by that script.  (scripts/regen_docs_outputs.py)

cyrush: Added this issue so we can continue to discuss: 
https://github.com/LLNL/conduit/issues/620

xjrc (approved):

src/tests/docs/t_conduit_docs_tutorial_basics.cpp
```diff
@@ -137,3 +137,142 @@ TEST(conduit_tutorial, basics_bw_style_from_native)
...
+
+    // setup a node
+    Node root;
+    // set data in hierarchy 
+    root["my/nested/path"] = 0.0;
+
+    // display the contents
+    root.print();
+
+    // In this case, notice we aren't use a reference.
```
agcapps, line 183: typo: should be "aren't using a reference."

src/tests/docs/t_conduit_docs_tutorial_python_basics.py
```diff
@@ -92,4 +92,74 @@ def test_007_basics_bw_style_from_native(self):
...
+    def test_008_basics_numpy_or_node(self):
+        BEGIN_EXAMPLE("py_basics_numpy_or_node")
+        # setup a node with a leaf array
+        n = conduit.Node()
+        data = numpy.zeros((5,),dtype=numpy.float64)
+        n["my/path/to/data"] = data
+        
+        # this will be a ndarray
+        my_data = n["my/path/to/data"]
+        print("== this will be a ndarray == ")
```
agcapps, line 104: typo: should be "this will be an ndarray"   (fix also lines 102, 125; line 123 has "an ndarray" already)

agcapps (changes requested):
Aside from the few typos I caught, it looks very good.  After those fixes I'll happily approve.

While including line numbers, timings, and addresses in regenerated example output files makes for messy merges and updates, they may be worth providing to users as real examples.

cyrush:
thanks @xjrc  and @agcapps for the feedback! I made adjustments to address these issues.

src/docs/sphinx/tutorial_python_basics.rst
```diff
@@ -167,7 +167,10 @@ You can check if a Schema is compatible with another Schema using the **Schema::
 Differences between C++ and Python APIs
 ----------------------------------------
 
-In Python, Node objects are referenced counted containers that hold C++ Node pointers.
+In Python, Node objects are referenced-counted containers that hold C++ Node
```
agcapps, line 170: @cyrush  Thanks--- please take out the d from "referenced"

agcapps (approved):
Approved (please catch the typo in tutorial_python_basics.rst).  Thanks!

PR #621, opened 2020-09-23
mesh bp gen index, bcast result to all ranks

### LINKED ISSUE #618: conduit::blueprint::mpi::mesh::generate_index
Just a comment/suggestion about this function in conduit::blueprint::mpi::mesh

```
void CONDUIT_BLUEPRINT_API generate_index(const conduit::Node &mesh,
                                          const std::string &ref_path,
                                          Node &index_out,
                                          MPI_Comm comm);
```

This is an MPI collective call, so all ranks must pass an `index_out` Node, but only one rank fills the Node with any data.  Looks like it will be rank 0 as long as it has any domains on that rank, but in general you don't know.  This means that the calling code needs to inspect `index_out` before doing anything with it subsequently, especially if it intends to pass it along to something that expects a non-empty Node.

I can deal with this by inspecting the DataType of the Node, and if it's an object, it must contain the generated index, but would it make sense to add something to to API that explicitly tells us which rank has the index in the Node?  Some possibilities would be to make the rank the return value, or to add an `Node& info` to the signature as a place to put a small amount of metadata.

cyrush:
resolves #618

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
@@ -120,6 +120,11 @@ generate_index(const conduit::Node &mesh,
                                                        index_out);
         }
     }
+
+    // broadcast the resulting index to all other ranks
+    relay::mpi::broadcast_using_schema(index_out,
+                                       selected_rank,
+                                       comm);
 }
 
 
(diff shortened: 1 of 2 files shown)


cyrush:
@nselliott this should fix #618

nselliott:
Thanks @cyrush

PR #624, opened 2020-10-10
Record buffer length needed for face data

agcapps:
The blueprint mesh representation uses a helper class, `TopologyMetadata`, to find and record the relations from element to faces and nodes.  There was a test in that computation that was too restrictive, resulting in a segfault in test t_blueprint_mesh_verify when it was compiled Debug on Windows by either MSVC or Intel 19 compilers.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -1201,7 +1201,7 @@ struct TopologyMetadata
                 dim_buffer.insert(dim_buffer.end(), entity_indices.begin(), entity_indices.end());
                 dim_entity_map[entity] = dim_offset++;
 
-                if (dim_shape.is_polygonal())
+                if (dim_shape.dim == 2)
                 {
                   dim_size.push_back(entity.size());
                 }


agcapps:
@xjrc , this change removes the check whether `dim_shape.embedding == NULL`.

Why do `is_poly()`, `is_polyhedral()`, `is_polygonal()` check whether `embedding == NULL`?  Put differently, what does `foo.is_poly*() -> true` mean about `foo`?

xjrc:
> Why do `is_poly()`, `is_polyhedral()`, `is_polygonal()` check whether `embedding == NULL`?

If a 2+-D entity doesn't have an embedding, that implies that the entity's sub-entity composition isn't statically defined, which is only the case for polytopal entities in Conduit. This ordering cannot be statically defined for such entities because the number of sub-entities is variable, and thus any sub-entity creation code must necessarily be implicit/generative rather than static. For example, the implicit/generative rules for polygons is that each pair of adjacent indices (including at the boundary of the arrays, so `0` and `n-1`) refer to vertices that comprise a single edge of a particular polygon.

>  Put differently, what does `foo.is_poly*() -> true` mean about `foo`?

Here is what each of the `is_poly*` functions mean:

- `is_poly()`: The topology is polytopal, which means it consists of either polygons (2D) or polyhedra (3D).
- `is_polyhedral()`: The topology consists of polyhedrea (3D).
- `is_polygonal()`: The topology consists of polygons (2D).

agcapps:
@xjrc , @cyrush , I don't know why the Main Docker_ubuntu build failed.  Could you get it to retry please?

cyrush:
@agcapps with @bmhan12's changes in #632, this should be resolved, so I am  closing this PR.

Can you double check on windows, and if there are more issues we can attack with a new PR?

agcapps:
Closing: superseded by @bmhan12 's changes in #632 .

PR #627, opened 2020-10-14
mesh bp fix struct topo optional origin names

### LINKED ISSUE #622: mesh bp struct topo: i,j,k origin, instead of i0,j0,k0 ?
For logical offsets in topology entry, should we be using `{i,j,k}` instead of `{i0,j0,k0}`

What `{i0,j0,k0}` implies seems slightly more confusing in a dist-mem context, so `{i,j,k}` might be better.

Discuss.

cyrush:
resolves #622

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 #### Blueprint
 - Refactored the Polygonal and Polyhedral mesh blueprint specification to leverage one-to-many concepts and to allow more zero-copy use cases.
 - The `conduit_blueprint_mpi` library now depends on `conduit_relay_mpi`.
+- Optional Mesh Blueprint structured topology logical element offsets are now `{i,j,k}` instead of `{i0,j0,k0}`.
 
 
 ## [0.5.1] - Released 2020-01-18
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 #### Blueprint
 - Refactored the Polygonal and Polyhedral mesh blueprint specification to leverage one-to-many concepts and to allow more zero-copy use cases.
 - The `conduit_blueprint_mpi` library now depends on `conduit_relay_mpi`.
-- Optional Mesh Blueprint structured topology logical element offsets are now `{i,j,k}` instead of `{i0,j0,k0}`.
+- The optional Mesh Blueprint structured topology logical element origin is now specified using `{i,j,k}` instead of `{i0,j0,k0}`.
 
 
 ## [0.5.1] - Released 2020-01-18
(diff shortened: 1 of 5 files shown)


cyrush:
@xjrc any thoughts on this? From #622 I think `{i,j,k}` seems more consistent with the rest of the BP? I am pretty sure I was the one that injected `{i0,j0,k0}` and I can't think of a strong reason to keep that going forward.

xjrc:
I'm honestly pretty ambivalent about this. I do agree that `{i, j, k}` is generally more consistent than `{i0, j0, k0}`, but Blueprint has a number of similar deviations that are used to provide context (e.g. `origin/{x, y, z}` vs. `spacing/{dx, dy, dz}` for uniform grid coordinate sets). I'm unfamiliar with the terminology used in the distributed memory domain, but if it causes ambiguity for the `{i0, j0, k0}` paths then I'm fine with the change (though I personally find the clarification to be useful).

As an open-ended thought exercise, I think it's useful to consider: what differentiates `{i0, j0, k0}` in its clarification from `{dx, dy, dz}`, and is that difference large enough to merit omitting the context in the former but not the latter?

cyrush:
My thoughts: for the `{dx,dy,dz}`, those actually do imply deltas - so I think that's straight forward and a bit more intuitive.

For`{i0,j0,k0}` it's the logical index of the first element of the current topo in the context of a larger index space that helps define a logically consistent space that spans multi-domains. The `0` tries to imply that, but I don't think it does a good job, so shifting to `{i,j,k}` is marginally better.

gzagaris:
> For`{i0,j0,k0}` it's the logical index of the first element of the current topo in the context of a larger index space that helps define a logically consistent space that spans multi-domains. The `0` tries to imply that, but I don't think it does a good job, so shifting to `{i,j,k}` is marginally better.

You could consider `imin` , `jmin` and `kmin` if you wanted to be explicit, which is also what is typically used in codes for this.

Just my 2c....

xjrc:
Speaking personally, I find `{i0, j0, k0}` to be useful in further clarifying that "this is the first element" in a very similar way to how `{dx, dy, dz}` clarifies "these are deltas." Maybe the latter has greater clarity as there's less ambiguity (0 means first in the computer science domain, but not in the physics domain), but I do find the attempt at clarification to have some utility (probably because I'm looking at it from a computer science lens).

I do like @gzagaris 's suggestion of `{imin, jmin, kmin}` as it feels like a disambiguating improvement, and it synchronizes nicely with some AMR patch path considerations we've had.

nselliott:
The reason I like `{i,j,k}` is that these are not in a vacuum.  They are `origin/{i,j,k}`, so the full path tells us they are the origin coordinates of the topology.  Without the `origin` indicator, they would need something more.

cyrush:
@xjrc  -- yes, `origin/{i,j,k}` matches how we specify windows for structured `adjsets` and `nestsets`. 

We also have things like `origin/{x,y,z}` for coordsets -- vs `origin/{xmin,ymin,zmin}`, or `origin/{x_min,y_min,z_min}`, etc

The fact that it matches the `nestsets` and `adjsets` is enough to convince me to commit to `origin/{i,j,k}`.

Any last strong objections :-) ?

xjrc:
Nope; I think it's fine to go ahead with the change to `{i, j k}`.

PR #632, opened 2020-10-16
Remove dim_sizes from TopologyMetadata

bmhan12:
This PR removes `dim_sizes` object from TopologyMetadata. 
`dim_sizes` logic was originally added to handle the case where if given a `shape: polyhedral` Node, the `elements/sizes` field needed to be initialized for the polygonal Node of `dim_topos` , the topology cascade.
Besides the polygonal 2D index, the rest of `dim_sizes` is uninitialized.
`dim_sizes` logic has been replaced for logic that handles the specific polyhedral Node input case.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -1077,7 +1077,6 @@ struct TopologyMetadata
         dim_topos[topo_shape.dim]["elements/offsets"].set(topo_offsets);
 
         std::vector< std::vector<int64> > dim_buffers(topo_shape.dim + 1);
-        std::vector< std::vector<int64> > dim_sizes(topo_shape.dim + 1);
         std::vector< int64 > dim_offsets(topo_shape.dim + 1);
 
         // Prepare Initial Values for Processing //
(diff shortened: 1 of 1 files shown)


cyrush:
@bmhan12 thanks for diving in to resolve and providing the additional info!

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1305,16 +1298,15 @@ struct TopologyMetadata
             dim_conn.set(DataType(int_dtype.id(), dim_buffers[di].size()));
             data_conn.to_data_type(int_dtype.id(), dim_conn);
 
-            // Generate size for polygonal mesh
-            if (di == 2)
+            // Initialize element/sizes for polygonal mesh using polyhedral's
+            // subelement/sizes
+            if (di == 2 && topo_shape.is_polyhedral())
```
xjrc, line 1303: Just to clarify @bmhan12, the code within this conditional basically just hooks up the `dim_topos` polygonal `Node` such that it points to the `dim_topos` polyhedral subelement `Node`, which allows the `offsets` generated to properly propagate in all places that it's needed (i.e. the polyhedral 3D "subelements" entry in `dim_topos` and the polygonal 2D "elements" entry in `dim_topos`), correct? Just want to make sure I understand; thanks for looking into this so quickly!

to hook up the correct path in the 3D case so that offsets are generated for the

xjrc (approved):

src/libs/blueprint/conduit_blueprint_mesh.cpp
bmhan12: Yes!

xjrc: Excellent, thanks Brian!

cyrush:
thanks @bmhan12 !

Issue #635, opened 2020-10-17
t_blueprint_mesh_verify fails on Windows


agcapps:
After @bmhan12 's fixes in #632 , I recompiled using MS Visual Studio and still had failures:

- Debug: test topology_unstructured reports failure.
  - Any call to CHECK_MESH()
    - after setting `n["subelements"][ (something) ]`, expecting the verify function to return `false`
    - or at lines 734 or 738 (also expect the verify function to return `false`)
- Release: all tests report success.

agcapps:
@cyrush , @xjrc , @bmhan12 , these are the remaining failures I see on Windows after #623 is resolved.

cyrush:
@agcapps I think this is now resolved, I just merged #636 and tested a debug build on windows and the tests passed. 

If there is anything else we should try to fix, can you throw up a new issue?

agcapps:
> 
> 
> @agcapps I think this is now resolved, I just merged #636 and tested a debug build on windows and the tests passed.
> 
> If there is anything else we should try to fix, can you throw up a new issue?

Thanks!  Looks great.  I'll put in a new issue if anything else crops up (which it hasn't so far).

Issue #277, opened 2018-04-03
validity status isn't always consistent in blueprint info nodes


xjrc:
There's currently an issue wherein the `valid` field contained in each subtree of the `info` node produced by the `blueprint::verify` can provide inconsistent and misleading results. This behavior has been observed particularly in `fields` nodes that contain a reference to an invalid topology, which can produce results similar to the following:

```
{
  /* ... */
  "fields": {
    "field": {
      "topology": {
        "valid": "false"
      }
      /* ... */
      "valid": "true"
    }
  },
  /* ... */
}
```

This should be fixed so that the a `valid` field value of `false` in any subtree causes each parent to also have a `valid` value of false.

PR #636, opened 2020-10-19
has_consistent_validity check fix in t_blueprint_mesh_verify

bmhan12:
For polygonal topology, propagate `elements` field's `valid` info to parent.

Related to https://github.com/LLNL/conduit/issues/635

The `has_consistent_validity` check (from https://github.com/LLNL/conduit/issues/277 and  https://github.com/LLNL/conduit/pull/283) was failing because the `info` node was inconsistent - topmost `valid` state was `true` despite a subtree (in this case, `elements`) being in a `valid` state of `false`.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -729,7 +729,7 @@ bool verify_poly_node(bool is_mixed_topo,
             elems_res &= subnode_res;
         }
     }
-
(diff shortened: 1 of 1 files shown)


cyrush:
FYI: we can ignore the docker build failure.

xjrc (approved):
Thanks for making this fix, @bmhan12 !

cyrush:
thanks @bmhan12  I ran a test on windows in a debug build with these changes and all the tests passed

PR #641, opened 2020-10-29
pass at resolving warnings for 0.6.0 release

cyrush:


src/libs/blueprint/conduit_blueprint_o2mrelation.cpp
```diff
@@ -189,7 +189,7 @@ void compact_to(const conduit::Node &o2mrelation,
     else
     {
         O2MIterator o2miter(o2mrelation);
-        const std::vector<std::string> o2m_paths =
+        const std::vector<std::string> o2m_paths_curr =
             conduit::blueprint::o2mrelation::data_paths(o2mrelation);
```
cyrush, line 193: @xjrc  can you take a look at this one? Prior to this change, this var was shadowing a global var name. I think it's fine with the name change fix (the intent is to use the local var, not the global one) but a check by you could give me confidence the warning wasn't pointing to something worse that I overlooked.

xjrc: Hey Cyrus, I took a quick look over the code to refresh my memory and your correction looks to be valid. In this scope, `o2m_paths_curr` (previously `o2m_paths`) refers to the data paths (e.g. `values`) in this one-to-many subtree, where `o2m_path` (at the global scope) refers to the paths reserved by one-to-many relations (e.g. `sizes`, `offsets`, etc.), which is consistent with the code attempting to compact data into the data paths below.

cyrush: thanks for checking!

Issue #476, opened 2019-12-11
extend Blueprint index to support AMR


xjrc:
In order to support AMR and other problems with nontrivial rank-domain distributions, the Blueprint index needs to be extended to include sparser and more descriptive domain metadata. More specifically, the rank/file-domain ownership relationship needs to be made more explicit and the ability to flexibly remove/exclude domains (e.g. those uninvolved in a particular segment of a calculation) needs to be added.

In order to accomplish these aims, the schema for the Blueprint index needs to be updated as follows:

### Current Schema ###
```
{
    "number_of_files": /* int */, // number of owner files 
    "number_of_trees": /* int */, // number of domain trees per file
    "file_pattern": /* string */, // template for owner file strings
    "tree_pattern": /* string */, // template for domain tree paths
}
```

### New Schema ###
```
{
    "domain_index": [ /* int */ ], // map of domain id to domain sparse index
    "domain_owner_map": [ /* int */ ], // map of domain sparse index to rank/file id
    "owner_pattern": /* string */, // template for owner rank/file strings
    "tree_pattern": /* string */, // template for domain tree paths
}
```

#### Problem ####
```
rX := rank with id X
dY := domain with id Y

|===================|===================|===================|
|         r0        |        r1         |        r2         |
|===================|===================|===================|
|  +----+   +----+  |  +----+           |  +----+   +----+  |
|  | d0 |   | d1 |  |  | d3 |           |  | d5 |   | d9 |  |
|  +----+   +----+  |  +----+           |  +----+   +----+  |
|===================|===================|===================|
```

#### Blueprint ####
```
{
    "domain_index": [ 0, 1, -1, 2, -1, 3, -1, -1, -1, 4 ],
    //                |  |      |      |              |
    //                |  +---+  |  +---+              |
    //                +---v  v  v  v  v---------------+
    "domain_owner_map": [ 0, 0, 1, 2, 2 ],
    "owner_pattern": "r%05d",
    "tree_pattern": "problem/domains",
}
```

[...]

PR #642, opened 2020-10-29
WIP polygonal/polyhedral transformations

nselliott:
Showing current state of the transformations.  2D is pretty much there, lots to do for 3D.


xjrc:
Thanks for putting this together, @nselliott; I'll give all this code a review tomorrow afternoon.

xjrc:
I've had a brief look over the code and I have a request to make the review a bit easier: would you mind giving a brief 1-2 sentence synopsis on each new function and its role in the AMR-polytopal conversion process? I've compiled a list of all of the new functions I could find below:

- `conduit::blueprint::mpi::mesh::to_polygonal(const Node&, Node&, const std::string&)`
- `conduit::blueprint::mpi::mesh::to_polyhedral(const Node&, Node&, const std::string&)`
- `conduit::blueprint::mpi::mesh::generate_partition(const Node&, Node&, MPI_Comm)`
- `conduit::blueprint::mesh::connectivity::make_element_2d(std::vector<int64_t>&, int64_t, int64_t)`
- `conduit::blueprint::mesh::connectivity::make_element_3d(PolyElemType&, int64_t, int64_t, int64_t, int64_t, std::map<int, std::vector<int64_t>>&, std::map<int, std::vector<int64_t>>&, std::map<int, std::vector<int64_t>>&)`
- `conduit::blueprint::mesh::connectivity::create_elements_2d(const Node&, int64_t, int64_t, int64_t, std::map<int, std::vector<int64_t>>&)`
- `conduit::blueprint::mesh::connectivity::create_elements_3d(const Node&, int64_t, int64_t, int64_t, int64_t, int64_t, int64_t, std::map<int, PolyElemType>&, std::map<int, std::vector<int64_t>>&, std::map<int, std::vector<int64_t>>&, std::map<int, std::vector<int64_t>>&)`
- `conduit::blueprint::mesh::connectivity::connect_elements_2d(const Node&, int64_t, int64_t, int64_t, int64_t, int64_t&, std::map<int, std::vector<int64_t>>&)`
- `conduit::blueprint::mesh::connectivity::connect_elements_3d(const Node&, int64_t, int64_t, int64_t, int64_t, int64_t, int64_t&, std::map<int, PolyElemType>&)`

nselliott:
> I've had a brief look over the code and I have a request to make the review a bit easier: would you mind giving a brief 1-2 sentence synopsis on each new function and its role in the AMR-polytopal conversion process? I've compiled a list of all of the new functions I could find below:

Certainly.  I suggest you mostly focus on the 2D functions, since the 3D functions are incomplete and may end up completely different.
 
>     * `conduit::blueprint::mpi::mesh::to_polygonal(const Node&, Node&, const std::string&)`

This is the user API call for the 2D transformation from a structured mesh to a polygonal mesh.

>     * `conduit::blueprint::mpi::mesh::to_polyhedral(const Node&, Node&, const std::string&)`

Likewise for 3D transformations from structured to polyhedral.  (This is not complete.)

>     * `conduit::blueprint::mpi::mesh::generate_partition(const Node&, Node&, MPI_Comm)`

This is not directly part of the transformations.  It is intended to be part of generate_index, to automatically set up the domain-to-rank partition map for the new multi-domain index format.
 
>     * `conduit::blueprint::mesh::connectivity::make_element_2d(std::vector<int64_t>&, int64_t, int64_t)`

This creates a single 2D quadrilateral polygonal element based on information from the structured topology.  The vertices go into the vector to be later accumulated into the larger connectivity array.  If the element needs to have extra vertices added at a domain boundary, that happens elsewhere.

>     * `conduit::blueprint::mesh::connectivity::make_element_3d(PolyElemType&, int64_t, int64_t, int64_t, int64_t, std::map<int, std::vector<int64_t>>&, std::map<int, std::vector<int64_t>>&, std::map<int, std::vector<int64_t>>&)`

This creates a single polyhedral element for a hexahedron computed from the structured topology.  It also collects some metadata that is intended for figuring out connections of faces across domain boundaries, but this will probably be revamped.
 
>     * `conduit::blueprint::mesh::connectivity::create_elements_2d(const Node&, int64_t, int64_t, int64_t, std::map<int, std::vector<int64_t>>&)`

This creates the 2d elements that are specifically located on the boundaries of 2D domains where they touch finer domains.  It calls make_elements_2d to set them up as quadrilaterals and to prepare them to have extra vertices added.

>     * `conduit::blueprint::mesh::connectivity::create_elements_3d(const Node&, int64_t, int64_t, int64_t, int64_t, int64_t, int64_t, std::map<int, PolyElemType>&, std::map<int, std::vector<int64_t>>&, std::map<int, std::vector<int64_t>>&, std::map<int, std::vector<int64_t>>&)`

Same thing for 3D.

>     * `conduit::blueprint::mesh::connectivity::connect_elements_2d(const Node&, int64_t, int64_t, int64_t, int64_t, int64_t&, std::map<int, std::vector<int64_t>>&)

This figures out the connections between the coarse elements that touch the domain boundary and their adjacent fine elements.  It adds new vertices to the coarse elements at the appropriate locations.

> 
>     * `conduit::blueprint::mesh::connectivity::connect_elements_3d(const Node&, int64_t, int64_t, int64_t, int64_t, int64_t, int64_t&, std::map<int, PolyElemType>&)`

Same for 3D, but it needs to figure out both new faces and new vertices, and it doesn't work.

src/libs/blueprint/conduit_blueprint_mesh.hpp
```diff
@@ -86,6 +86,70 @@ void CONDUIT_BLUEPRINT_API generate_index(const conduit::Node &mesh,
                                           index_t num_domains,
                                           Node &index_out);
 
+namespace connectivity
```
xjrc: Given their nature as helper functions for the greater AMR-to-polytopal process, I think it would be best for all of these functions to be excluded from external-facing headers (similar to [`ShapeType`](https://github.com/LLNL/conduit/blob/develop/src/libs/blueprint/conduit_blueprint_mesh.cpp#L174) and its associated functionality).

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2386,6 +2386,536 @@ void mesh::to_multi_domain(const conduit::Node &n,
     }
 }
 
+//-----------------------------------------------------------------------------
+void
+mesh::connectivity::make_element_2d(std::vector<int64_t>& elem,
+                                    int64_t element,
+                                    int64_t iwidth)
+{
```
xjrc: I feel like the interface for this function is a bit ambiguous. In particular, the contents of the out variable `elem` could be one of many things (a list of *just* this 2D element's vertices etc.), where they are in fact an accumulated list of potentially many 2D elements. In order to make this behavior clearer, could we rename this function to something like `append_element_2d`?

xjrc: This question extends to the `{i|j|k}faces` output arguments to the `make_element_3d` function.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2386,6 +2386,536 @@ void mesh::to_multi_domain(const conduit::Node &n,
...
+void
+mesh::connectivity::make_element_3d(PolyElemType& connect,
+                                    int64_t element,
+                                    int64_t iwidth,
+                                    int64_t jwidth,
+                                    int64_t kwidth,
+                                   std::map<int, std::vector<int64_t> >& ifaces,
+                                   std::map<int, std::vector<int64_t> >& jfaces,
+                                   std::map<int, std::vector<int64_t> >& kfaces)
+
```
xjrc: A minor nitpick, but the indentation for the `{i|j|k}faces` arguments is slightly off.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2386,6 +2386,536 @@ void mesh::to_multi_domain(const conduit::Node &n,
...
+            auto& khi_face = kfaces[hi_face];
+            khi_face.push_back(ilo + jlo_offset + khi_offset);
+            khi_face.push_back(ihi + jlo_offset + khi_offset);
+            khi_face.push_back(ihi + jhi_offset + khi_offset);
+            khi_face.push_back(ilo + jhi_offset + khi_offset);
+        }
+        connect.m_subelems[kface_start+hi_face] = kfaces[hi_face];
+        connect.m_elems.push_back(kface_start + lo_face);
+        connect.m_elems.push_back(kface_start + hi_face);
+    }
```
xjrc: I feel like there has to be some good way to abstract the behavior used for each dimension in this function into a general routine across dimensions (e.g. something like [`grid_id_to_ijk`](https://github.com/LLNL/conduit/blob/develop/src/libs/blueprint/conduit_blueprint_mesh.cpp#L313)).

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2386,6 +2386,536 @@ void mesh::to_multi_domain(const conduit::Node &n,
...
+                                       int64_t i_lo,
+                                       int64_t j_lo,
+                                       int64_t k_lo,
+                                       int64_t iwidth,
+                                       int64_t jwidth,
+                                       int64_t kwidth,
+                                    std::map<int, PolyElemType>& elems,
+                                  std::map<int, std::vector<int64_t> >& ifaces,
+                                  std::map<int, std::vector<int64_t> >& jfaces,
+                                  std::map<int, std::vector<int64_t> >& kfaces)
```
xjrc: The argument indentation is off here as well.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2386,6 +2386,536 @@ void mesh::to_multi_domain(const conduit::Node &n,
...
+
+
+
+//-----------------------------------------------------------------------------
+void
+mesh::connectivity::create_elements_2d(const Node& ref_win,
+                                       int64_t i_lo,
+                                       int64_t j_lo,
+                                       int64_t iwidth,
+                                       std::map<int, std::vector<int64_t> >& elems)
```
xjrc: > This creates the 2d elements that are specifically located on the boundaries of 2D domains where they touch finer domains. It calls make_elements_2d to set them up as quadrilaterals and to prepare them to have extra vertices added.

I'm having difficulty mapping the above description onto the arguments presented for the `create_elements_2d` function. In my mind, it seems sufficient to pass just the structured AMR mesh window and a direction to generate all of the boundary elements for that window. I can square this circle by mapping `i_lo` and `j_lo` onto some sort of boundary anchor, but the existence of `iwidth` (which, to me, seems redundant relative to the information in `ref_win`) makes me question whether I understand this function at all.

If you could Noah, could you please explain how this function selects the boundary from within the given `ref_win` using `i_lo`, `j_lo`, and `iwidth`?

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2386,6 +2386,536 @@ void mesh::to_multi_domain(const conduit::Node &n,
...
+    for (int kidx = kstart; kidx < kend; ++kidx)
+    {
+        for (int jidx = jstart; jidx < jend; ++jidx)
+        {
+            for (int iidx = istart; iidx < iend; ++iidx)
+            {
+                int offset = kidx*iwidth*jwidth + jidx*iwidth + iidx;
+                auto& elem_conn = elems[offset];
+                elem_conn.m_elems.push_back(new_vertex);
+                ++new_vertex;
```
xjrc: The use of the output variable `new_vertex` here seems highly suggestive that this whole set of functions should be contained within a struct/class like [`ShapeType`](https://github.com/LLNL/conduit/blob/develop/src/libs/blueprint/conduit_blueprint_mesh.cpp#L174), as the purpose of such variables and how they're initialized/used becomes too opaque otherwise.

xjrc: Also, I was tempted to ask how you were able to get away with this function being simpler than `connect_elements_2d` (a bit of a hopeful leading question toward simplifying `connect_elements_2d`), but the fact that this function doesn't work indicates that it may become more complex as it becomes more complete.

src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
```diff
@@ -103,6 +103,28 @@ generate_index(const conduit::Node &mesh,
...
 
+    Node do_par_map;
+    do_par_map.set_uint8(0);
+    if(par_rank == selected_rank )
+    {
+        if(::conduit::blueprint::mesh::is_multi_domain(mesh))
+        {
+            if(mesh.child(0).has_child("state") &&
+               mesh.child(0)["state"].has_child("domain_id"))
+            {
```
xjrc: Can't all of these cascading `if` statements be combined into a single large clause via `&&`? For example:

```
if(par_rank == selected_rank &&
        ::conduit::blueprint::mesh::is_multi_domain(mesh) &&
        mesh.child(0).has_child("state") &&
        mesh.child(0)["state"].has_child("domain_id")) {
    do_par_map.set_uint8(1);
}
```

I feel like this style more clearly signals the intent; with the cascading `if` statements, I'm always on the lookout for extra behavior within each individual clause.

src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
```diff
@@ -119,6 +141,12 @@ generate_index(const conduit::Node &mesh,
                                                        global_num_domains,
                                                        index_out);
         }
+
+        if (!partition.dtype().is_empty())
+        {
+            index_out["state/partition_size"].set(par_size);
+            index_out["state/domain_to_partition_map"].set(partition);
```
xjrc: For my own benefit, this naming scheme appears to have been derived from #476 (and specifically, [this comment](https://github.com/LLNL/conduit/issues/476#issuecomment-662120834)).

src/libs/blueprint/conduit_blueprint_mpi_mesh.hpp
```diff
@@ -66,10 +66,22 @@ void CONDUIT_BLUEPRINT_API generate_index(const conduit::Node &mesh,
                                           Node &index_out,
                                           MPI_Comm comm);
 
+
+void CONDUIT_BLUEPRINT_API to_polygonal(const conduit::Node &n,
+                                   conduit::Node &dest,
+                                   const std::string& name);
+void CONDUIT_BLUEPRINT_API to_polyhedral(const conduit::Node &n,
+                                   conduit::Node &dest,
+                                   const std::string& name);
```
xjrc: On the one hand, I realize that these functions are peculiar because they apply mesh-wide transformations and thus don't fit nicely within an embedded namespace (e.g. `conduit::blueprint::mpi::mesh::topology::xyz::to_polygonal`), but I feel like more clarification is necessary to properly scope the set of functions to which this can be applied (i.e. structured AMR meshes). As an initial suggestion to start a conversation, would something like `from_amr_to_poly{gonal|hedral}` be an improvement?

src/tests/blueprint/t_blueprint_mpi_mesh_transform.cpp
```diff
@@ -0,0 +1,381 @@
...
+    group_1_0["windows/window_000001/dims/j"] = 9;
+    group_1_0["windows/window_000001/ratio/i"] = 2;
+    group_1_0["windows/window_000001/ratio/j"] = 2;
+    group_1_0["windows/window_000000/level_id"] = 0;
+    group_1_0["windows/window_000000/origin/i"] = 4;
+    group_1_0["windows/window_000000/origin/j"] = 0;
+    group_1_0["windows/window_000000/dims/i"] = 1;
+    group_1_0["windows/window_000000/dims/j"] = 5;
+    group_1_0["windows/window_000000/ratio/i"] = 2;
+    group_1_0["windows/window_000000/ratio/j"] = 2;
```
xjrc: As this example is a bit complicated, would it be possible to accompany it with an ASCII diagram in a code comment?

src/tests/blueprint/t_blueprint_mpi_mesh_transform.cpp
```diff
@@ -0,0 +1,381 @@
...
+        group_0_1["windows/window_000000/dims/j"] = 5;
+        group_0_1["windows/window_000000/ratio/i"] = 2;
+        group_0_1["windows/window_000000/ratio/j"] = 2;
+        group_0_1["windows/window_000001/level_id"] = 1;
+        group_0_1["windows/window_000001/origin/i"] = 8;
+        group_0_1["windows/window_000001/origin/j"] = 0;
+        group_0_1["windows/window_000001/dims/i"] = 1;
+        group_0_1["windows/window_000001/dims/j"] = 9;
+        group_0_1["windows/window_000001/ratio/i"] = 2;
+        group_0_1["windows/window_000001/ratio/j"] = 2;
```
xjrc: It looks like this code was lifted from the previous example; would it be possible to abstract this into a helper function?

xjrc (changes requested):
Sorry it took me so long to get to this @nselliott, but I was able to go through most of it. Most of my comments are questions and some software architecture/organization suggestions, but overall the functionality looks good. I think we should converge on the final placement of some of this functionality before we should merge, but I think that it's close.

src/libs/blueprint/conduit_blueprint_mesh.hpp
nselliott: So just put these helpers completely in conduit_blueprint_mpi_mesh.cpp?  That is fine, as I will defer to you on these kinds of code organization matters.  I separated them because they have no MPI, but if their only use is to enable MPI-based functionality, it still makes sense.

src/libs/blueprint/conduit_blueprint_mesh.cpp
nselliott: This always creates one element defined by four vertices.  There is no accumulation nor appending of elements inside this function.

nselliott: `make_element_3d` is different because it has to handle both elements and subelements.  Each call creates a single element, but that element is made up of subelements (faces) that can be shared by two elements.  The `{i|j|k}faces` arguements are there so that the function can first figure out if a needed subelement already exists, then use it if it does, or create a new one and append it if it doesn't.

src/libs/blueprint/conduit_blueprint_mesh.cpp
nselliott: Probably so.  I brute-forced this a bit when I ripped out an earlier version that used the previous layout from the Conduit 0.5.1 documentation.

src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
nselliott: Yes, it's just my style preference to not have an if statement go beyond two lines, but I don't object to changing it to your way.

src/libs/blueprint/conduit_blueprint_mpi_mesh.hpp
nselliott: Could be `structured_to_polygonal`?  Strictly speaking, the input mesh doesn't have to be an AMR mesh.

src/tests/blueprint/t_blueprint_mpi_mesh_transform.cpp
nselliott: Let's see if this works.  2 domains of differing resolutions, and the coarse elements at the interface will get extra vertices in the transformation.

```
_______________________
|    |    |__|__|__|__|
|____|____|__|__|__|__|
|    |    |__|__|__|__|
|____|____|__|__|__|__|
```

src/tests/blueprint/t_blueprint_mpi_mesh_transform.cpp
nselliott: Yes, this is true.

src/libs/blueprint/conduit_blueprint_mesh.cpp
nselliott: `ref_win` comes from the `adjset` and defines a 1d boundary interface with an adjacent domain, but it doesn't contain information about which side of the domain (left, right, lower, upper) that it's on, so the function figures it out by comparing it to information from the domain's topology.  It figures out whether ref_win is a i face or a j face based on which direction is width 1, and then it compares its origin coordinates to `i_lo` or `j_lo` to figure out if it is on the lower or upper side.  It then traverses the length of `ref_win` and calls `make_element_2d` for to create every domain element along that side of the domain.  `iwidth` is used to map each element to an offset within the domain based on the formula `offset = j*iwidth + i`.

An alternative would be to pass the topology itself, and then extract needed values inside the function, but these three values had already been extracted before the code gets to the one place that this function is called.

src/libs/blueprint/conduit_blueprint_mesh.cpp
nselliott: I probably should have deleted `connect_elements_3d` as it doesn't really do anything at this point.  Connecting the 2D elements is complicated, but it's ultimately just adding new vertices to polygons.  In 3D there are new faces created, new vertices added to old faces, and then this information has to be sent to any element that touches an affected face.

src/libs/blueprint/CMakeLists.txt
```diff
@@ -90,7 +90,7 @@ add_compiled_library(NAME conduit_blueprint
                      EXPORT conduit
                      HEADERS ${blueprint_headers} ${blueprint_c_headers}
                      SOURCES ${blueprint_sources} ${blueprint_c_sources} ${blueprint_fortran_sources}
-                     DEPENDS_ON conduit
+                     DEPENDS_ON conduit mpi
```
cyrush: quick note: we don't want the base blueprint lib to depend on mpi, only the blueprint_mpi lib

src/libs/blueprint/conduit_blueprint_mesh.cpp
xjrc: After reviewing your comments and the content of `conduit::blueprint::mesh::connectivity::create_elements_2d`, I understand how this works now. As I have alluded to elsewhere, I feel like this interface suggests that there should be some form of superstructure that encapsulates all of these bits of functionality in a more legible way (e.g. `struct`/`class` a la `ShapeTopology`).

src/tests/blueprint/t_blueprint_mpi_mesh_transform.cpp
```diff
@@ -20,158 +20,196 @@ using namespace conduit;
...
+//
+//  _____________________________________________
+//  |    |    |    |    |__|__|__|__|__|__|__|__|
+//  |____|____|____|____|__|__|__|__|__|__|__|__|
+//  |    |    |    |    |__|__|__|__|__|__|__|__|
+//  |____|____|____|____|__|__|__|__|__|__|__|__|
+//  |    |    |    |    |__|__|__|__|__|__|__|__|
+//  |____|____|____|____|__|__|__|__|__|__|__|__|
+//  |    |    |    |    |__|__|__|__|__|__|__|__|
+//  |____|____|____|____|__|__|__|__|__|__|__|__|
```
xjrc: Beautiful; thank you for adding this, Noah!

src/tests/blueprint/t_blueprint_mpi_mesh_transform.cpp
```diff
@@ -364,6 +257,11 @@ TEST(conduit_blueprint_mesh_transform, amr_2d_transform_parallel)
     conduit::blueprint::mpi::mesh::to_polygonal(mesh, poly, "topo");
 
     EXPECT_TRUE( conduit::blueprint::mpi::verify("mesh",poly,info, MPI_COMM_WORLD));
```
xjrc: I think it would be good if there were a few more tests here to ensure that the new mesh has the expected structure, e.g.:

```
EXPECT_TRUE(conduit::blueprint::mesh::topology::unstructured::verify(poly["path/to/topo"]));
EXPECT_EQ(poly["path/to/topo/elements/shape"].as_string(), "polygonal");
```

xjrc: Also, some broad-level checks to ensure that the contents look correct:

```
EXPECT_EQ(/*# elements in polygonal topology*/, /*# elements in original structured AMR topology*/);
```

src/libs/blueprint/CMakeLists.txt
nselliott: Good catch--I temporarily had some MPI code in conduit_blueprint_mesh before conduit_blueprint_mpi_mesh existed, but this was not intended to be merged.

cyrush:
@nselliott  @xjrc  -- what do we still have pending here, are there more changes in the works or do we think this is ready to go in?

nselliott:
> @nselliott @xjrc -- what do we still have pending here, are there more changes in the works or do we think this is ready to go in?

The one remaining missing piece is the correct handling of the case where a coarse element receives new hanging vertices on more than one of its faces.  The new vertices on the edges shared between those faces are duplicated rather than treated as a single vertex.  This is not easy to explain in text, so I will try to make a crude PowerPoint drawing and send email.

nselliott:
@cyrush @xjrc OK, I consider this ready to be merged.

cyrush:
thanks @nselliott !

PR #651, opened 2020-11-11
fix minor bugs w/ blueprint's `unstructured::generate_*` functions

xjrc:
The changes in this pull request make the following changes to the Blueprint source code:

- [x] Update the output map variables for all of the `conduit::blueprint::mesh::topology::unstructured::generate_*` functions to adhere to [the O2M schema](https://llnl-conduit.readthedocs.io/en/latest/blueprint_o2mrelation.html).
- [x] Update the `generate_vertices`, `generate_edges`, and `generate_faces` functions to use more sensible orderings for 2D and 3D.
- [x] Update the `generate_sides` and `generate_corners` functions to generate in orders that match the `generate_edges` function and the per-element vertex listings, respectively.
- [x] Improve the `generate_sides` and `generate_corners` functions to generate topological entities that follow the same orientation rules as their source meshes (e.g. counter-clockwise faces in 2D, outward normal faces in 3D).
- [x] Add test cases for the new orientation retention behavior of `generate_sides` and `generate_corners`.


xjrc:
@cyrush, @terryhaut: Just as an FYI, this PR is now quite close to completion. I'm going to do one final pass over the source tomorrow (i.e. 12/08/20) to make some refactoring improvements, but the base implementation is there and seems to be working (based on cursory checks and regression tests, at least).

xjrc:
@cyrush: This pull request is now ready for review.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -338,6 +345,45 @@ std::vector<index_t> intersect_sets(const std::set<index_t> &s1,
     return std::vector<index_t>(si.begin(), si_end);
 }
 
+//-----------------------------------------------------------------------------
+std::vector<index_t> intersect_sets(const std::vector<index_t> &v1,
+                                    const std::vector<index_t> &v2)
+{
```
cyrush, line 352: how big are these expected to be? If large, we migh want to pass in refs for output to avoid extra copies.

cyrush (approved):
Hi Joe,

This looks good to me.

The only thing I can thing that could improve is if there were comments that outlined the big picture a bit more, and some of the considerations you navigated. I can follow the logic, but I can't help verify it follows your intended path.

Not a blocker for this PR, just a thought.

src/libs/blueprint/conduit_blueprint_mesh.cpp
xjrc: Generally they're expected to be small, but I agree that we should avoid any unnecessary copies. I was blindly hoping that the compiler avoid the copy via [copy elision](https://en.cppreference.com/w/cpp/language/copy_elision) in this case, but I should double check on that.

xjrc:
> The only thing I can thing that could improve is if there were comments that outlined the big picture a bit more, and some of the considerations you navigated. I can follow the logic, but I can't help verify it follows your intended path.

That's a fair point; I'll go over this again and make my intent for the more complicated sections more clear.

src/libs/blueprint/conduit_blueprint_mesh.cpp
xjrc: I did some reading [here](https://www.fluentcpp.com/2016/11/28/return-value-optimizations/) and found that whether or not optimizations will be applied is based largely on the compiler in these cases, so it's probably safest to do output reference instead to ensure that unnecessary copies aren't being made.

xjrc:
@cyrush: I tried to address both your points by:

- Improving documentation for the `generate_corners`, `generate_sides`, and `TopologyMetadata::IndexType` components, which I felt were the parts most in need of explanation.
- Forcing the `set_*` operations in the Blueprint source to use RVO instead of NRVO, which [I'm led to believe](https://shaharmike.com/cpp/rvo/) is supported by all major compilers and should eliminate any potential inefficient by-value return copies.

Let me know what you think!

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -381,7 +382,7 @@ std::vector<index_t> subtract_sets(const std::vector<index_t> &v1,
             res.push_back(v1[i1]);
         }
     }
-    return res;
+    return std::vector<index_t>(std::move(res));
 }
```
cyrush, line 386: looks good, now that we are c++11 we do this. 

(I will still probably be a dinosaur and use out refs, b/c they are a known qty)

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1133,8 +1178,9 @@ struct TopologyMetadata
         // associated conversations (data).
         Node temp, data;
 
-        // NOTE(JRC): A 'deque' is used in order to allow for queue behavior
-        // to be used in deciding the identifiers for the cascade of entities.
+        // NOTE(JRC): A 'deque' is used so that queue behavior (FIFO)
+        // is responsible for ordering the identifiers in the cascade of entities,
+        // which more closely follows the average intuition.
         const index_t bag_num_elems = topo_num_coords + topo_num_elems;
```
cyrush, line 1184: I like this term.
(FYI: i am below average intuition :-) )

cyrush (approved):
These extra comments look good. Thanks for taking the time to put them in, I think they will help us with future understanding!

src/libs/blueprint/conduit_blueprint_mesh.cpp
xjrc: I think the phrase I was looking for was "standard intuition," but calling this "average" (as opposed to enlightened forms such as LIFO, etc.) is more comical.

PR #658, opened 2020-11-23
fix minor bug in default argument of 'Node::to_string_stream'

xjrc:
The changes in this pull request fix a minor bug wherein `conduit::Node::to_string_stream(const std::string&, ...)` was being output to the JSON format instead of the YAML format by default.


cyrush (approved):
Thanks!

PR #660, opened 2020-11-24
addition of custom allocator

mclarsen:
Notes / Design

- Allocator should be sticky/inherited, meaning that all child nodes inherit the same allocator.
- Setting an allocator on a node destroys it.
  - In the conduit node class, this means the first thing you have to do is set the allocator or else it breaks things.
  - I supposed its possible to add a behavior that copies itself to a node using the new allocator rather than complete destruction.

Allocator Interface:

- create memory (e.g., `calloc`)
- destroy memory (e.g., `free`)
- memset (to have nice 0 initializations)
- memcpy

 
Use cases:

- Changing conduit to use a pool allocator. We have had users try to use conduit to store complex data structures where the leaves are not arrays. When using it in practice, there were performance issues where the majority of the time was spent creating and destroying memory.
- Changing conduit to allocate GPU memory. There is the potential to create a smart memcpy that queries pointer attributes and can be interleaved with CPU pointers without the use of unified memory.


mclarsen:
@cyrush Should I create another pull request that removes all trailing whitespace from these files since this diff is unreadable?

xjrc:
> @cyrush Should I create another pull request that removes all trailing whitespace from these files since this diff is unreadable?

I would personally prefer a separate PR for those changes since they affect so many lines.

src/libs/conduit/conduit_node.cpp
```diff
@@ -4385,7 +4393,8 @@ Node::set_external_data_using_schema(const Schema &schema,
 {
     reset();
     m_schema->set(schema);
-    walk_schema(this,m_schema,data);
+#warning "should set_external_data_using_schema inherit allocator?"
```
mclarsen: What needs to be done here?

src/libs/conduit/conduit_node.cpp
```diff
@@ -12468,6 +12477,7 @@ Node::to_base64_json(std::ostream &os,
     Node bb64_data;
     bb64_data.set(DataType::char8_str(enc_buff_size));
 
+#warning "I don't think to_base64_json should use the allocator, but it references mem"
```
mclarsen: What happens if this is a GPU allocator?

src/libs/conduit/conduit_node.cpp
```diff
@@ -466,6 +467,7 @@ Node::set_node(const Node &node)
             Schema *curr_schema = &this->m_schema->add_child(*itr);
             size_t idx = (size_t) this->m_schema->child_index(*itr);
             Node *curr_node = new Node();
+            curr_node->set_allocator(m_allocator_id);
```
mclarsen, line 470: In the current implementation, setting the allocator MUST be first, since its calls `reset()`. I don't like that there is unintuitive ordering.

src/libs/conduit/conduit_utils.cpp
```diff
@@ -72,6 +73,84 @@ namespace conduit
...
+int32 register_mem_handler( void*(*allocate) (size_t, size_t),
+                            void(*free)(void *),
+                            void(*copy)(void*,const void *,size_t),
+                            void(*memset)(void*,int,size_t))
+{
+  static int32 allocator_id = 1;
+  allocator_map[allocator_id] = allocate;
+  free_map[allocator_id] = free;
+  memcpy_map[allocator_id] = copy;
+  memset_map[allocator_id] = memset;
```
mclarsen: I think I need to change these names. I can totally see a compiler using the other free, not the provided one.

src/libs/conduit/conduit_node.cpp
cyrush: in this case, Node bb64_data is not a child of the current node, so it would use the default alloc and I think we are ok?

src/libs/conduit/conduit_node.cpp
```diff
@@ -420,7 +420,8 @@ Node::mmap(const std::string &stream_path,
     m_mmaped = false;
 
     m_schema->set(schema);
-    walk_schema(this,m_schema,m_data);
+#warning "I don't know if this is ok(mmap). It might need to be set to the defautl alloc_id"
+    walk_schema(this,m_schema,m_data, m_allocator_id);
```
cyrush, line 425: mmap itself is like using a special allocator. When mmap is used -- it should dictate the the default alloc is used. 

the way mmap works now -- you setup it up, and if you deviate (set that isn't compat, etc) you are back to default

src/libs/conduit/conduit_node.cpp
cyrush: for internal construction, should we have a helper that sets the allocator w/o reset?

src/libs/conduit/conduit_node.cpp
cyrush: i do think it should inherit -- external vs alloced are orthogonal paths

src/libs/conduit/conduit_node.cpp
```diff
@@ -15185,6 +15200,29 @@ Node::as_long_double_array() const
 //
 //-----------------------------------------------------------------------------
 
+void Node::set_allocator(int32 allocator_id)
```
cyrush: should these handles be index_t?

src/libs/conduit/conduit_utils.cpp
cyrush: agreed, can we also name this `register_memory_handler`

src/libs/conduit/conduit_node.cpp
cyrush: ahh, i see below: we need to think about this -- call the memcpy method?

mclarsen:
@cyrush so i have a fancy allocator that automagically does the right thing with memcpy, but I can't figure out a way to use it to automatically copy data from either another node or from a pointer.

In the case of a data array, a generic data array is constructed and it does the copying, not knowing anything about the destination. This is backwards from what I was hoping. Can you see a clean way to have the data array use the allocator of the destination to do the copying?

src/libs/conduit/conduit_utils.cpp
cyrush: i updated these names

src/libs/conduit/conduit_node.cpp
cyrush: i updated these to use `index_t`

src/libs/conduit/conduit_utils.hpp
```diff
@@ -238,6 +238,76 @@ namespace utils
...
+
+    // conduit uses a single pair of memset and memcpy functions to
+    // manage data movement. 
+
+    // this strategy allows downstream users to support complex cases
+    // like moving between memory spaces non accessible on the host. 
+    //
+    // These methods aren't bound to allocators b/c allocators
+    // won't be tied into all of the places where source and dest pointers
+    // need to be located.
```
xjrc, line 279: I understand the need to decouple `memset` and `memcpy` from allocators, but I don't understand why Conduit only supports a single pair at any given point in time. Isn't it conceivable that a user may want multiple functions that they could choose between with minimal hassle, like a `memcpy` function for transferring from host to device and vice versa?

src/tests/conduit/t_conduit_mem_allocator.cpp
```diff
@@ -0,0 +1,81 @@
...
+  {
+    std::cout<<"set bananas\n";
+    memset(ptr,value,num);
+  }
+
+  static void banana_copy(void * destination, const void * source, size_t num)
+  {
+    std::cout<<"copy bananas\n";
+    memcpy(destination,source,num);
+  }
```
xjrc: This is a good way to quickly and manually verify that the custom allocation code is working, but we probably want a different mechanism so that the testing can be automated with `ASSERT_*` checks.

xjrc (commented):
This is a very exciting new capability! I've added some general comments and questions on the current status, and I'm happy to do a more in-depth review once the code is ready.

src/libs/conduit/conduit_utils.hpp
cyrush: The first pass had these pinned to allocators. 

The problem is while a Node may know its allocator (and it may even know the allocator of a Node that is being set into it) there is no good notion of how to move data between allocators. Umpire provides a platform interface to solve this, but we don't want to go that far.  There are also cases where allocator is completely unknown and those are most of the interface. 
 
So -- In a bunch of the cases where we need to move data, the allocator (or more broadly where the data lives), both source or dest, won't be known explicitly by any tagging or metadata that conduit has access to. The only solution is pointer inspection. The solve this --- folks doing fancy things will create a memcpy that uses pointer inspection to handle all situations in play. So the one registered instance will have to be smart enough to move between any funky mem spaces in use. Matt is doing this in Ascent, so that will be a concrete example to test these ideas.

mclarsen: Note: printing a node containing device values segfaults, but I didn't expect that to work. `bool danger()`

mclarsen: perhaps `bool calamitous()`

mclarsen: What also doesn't work (which I expected) is setting a gpu allocator to a node and passing it to the blueprint examples. I don't think we need to handle this case, because its easy enough to `Node device_node = host_node;`

mclarsen:
Another potential issue is how to handle `as_int32()` and friends. For example, I created a device memory backed data set which included the `state/domain_id`. When I tried to get the values inside ascent, it accesses the pointer directly, which segfaults. Since we never intend for conduit Nodes to be accessed in device code, all methods that return integral types (and maybe strings) should probably call memcpy into a result value to be safe.

Edit: most definitely strings (i.e., `association`)

src/libs/conduit/conduit_node.cpp
cyrush: I didn't yet update these to use `index_t`

src/libs/conduit/conduit_node.cpp
cyrush: I think this case is fine b/c, above the data is always compacted to a node that uses the default allocator.

src/libs/conduit/conduit_node.cpp
cyrush: flip flopped on this, it should propagate the alloc like set_external does

src/tests/conduit/t_conduit_mem_allocator.cpp
cyrush: I added some assert checks that verify tracked counts

src/libs/conduit/conduit_node.hpp
```diff
@@ -4320,7 +4343,8 @@ class CONDUIT_API Node
     // work horse for complex node hierarchical setup
     static void      walk_schema(Node   *node,
                                  Schema *schema,
-                                 void   *data);
+                                 void   *data,
+                                 int32 allocator_id);
```
cyrush: - [x] TODO for cyrush: Change to index_t

src/libs/conduit/conduit_utils.cpp
```diff
@@ -75,6 +76,179 @@ namespace conduit
...
+//                          void(*conduit_hnd_copy)(void*,const void *,size_t),
+//                          void(*conduit_hnd_memset)(void*,int,size_t))
+// {
+//   static index_t allocator_id = 1;
+//   allocator_map[allocator_id] = conduit_hnd_allocate;
+//   free_map[allocator_id]      = conduit_hnd_free;
+//   memcpy_map[allocator_id]    = conduit_hnd_copy;
+//   memset_map[allocator_id]    = conduit_hnd_memset;
+//   return allocator_id++;
+// }
```
cyrush: - [x] TODO for cyrush, remove commented out code

src/libs/conduit/conduit_node.cpp
```diff
@@ -16096,7 +16319,8 @@ Node::identify_protocol(const std::string &path,
 void
 Node::walk_schema(Node   *node,
                   Schema *schema,
-                  void   *data)
+                  void   *data,
+                  int32   allocator_id)
```
cyrush: - [x] TODO for cyrush `index_t`

src/libs/conduit/conduit_node.cpp
```diff
@@ -16212,21 +16440,41 @@ Node::compact_elements_to(uint8 *data) const
        dtype_id == DataType::LIST_ID ||
        dtype_id == DataType::EMPTY_ID)
     {
-        // TODO: error
+        // TODO: error ?
+        // NOTE: WE RELY ON THIS NOT BEING AN ERROR
+        // FIND OUT WHY!
+        // CONDUIT_ERROR("Node::compact_elements_to cannot compact "
+        //               " Object, List, or Empty Node to array.");
```
cyrush, line 16456: - [x] TODO cyrush, add ticket for this post merge

src/libs/conduit/conduit_node.cpp
```diff
@@ -16212,21 +16440,41 @@ Node::compact_elements_to(uint8 *data) const
...
+        // uint8 *data_ptr = data;
+        // for(index_t i=0;i<num_ele;i++)
+        // {
+        //   utils::conduit_memcpy(data_ptr,
+        //                         element_ptr(i),
+        //                         (size_t)ele_bytes,
+        //                         m_allocator_id);
+        //     data_ptr+=ele_bytes;
+        // }
+
```
cyrush: - [x] TODO cyrush, remove commented out code

src/libs/conduit/conduit_utils.hpp
```diff
@@ -263,6 +263,76 @@ namespace utils
...
+
+//-----------------------------------------------------------------------------
+/// Primary interface used by the conduit API to move memory.
+//-----------------------------------------------------------------------------
+
+    // conduit uses a single pair of memset and memcpy functions to
+    // manage data movement. 
+
+    // this strategy allows downstream users to support complex cases
+    // like moving between memory spaces non accessible on the host.
```
cyrush: - [x] TODO cyrush `non` -> `not`

cyrush:
mesh bp verify is crashing in some CI cases, I want to track this down. 

  58/80 Test #58: t_blueprint_mesh_verify ....................***Exception: SegFault  0.01 sec

cyrush:
the issue is with static builds, I'll hunt it down

cyrush:
```
  frame #10: 0x0000000100003849 t_blueprint_mesh_verify`get_log_keywords() at t_blueprint_mesh_verify.cpp:31:5
  frame #11: 0x00000001000c7a24 t_blueprint_mesh_verify`::__cxx_global_var_init.69() at t_blueprint_mesh_verify.cpp:38:47
```


As I feared, I am pretty sure we have triggered a C++ static object init nightmare. 

https://isocpp.org/wiki/faq/ctors#static-init-order


```
std::vector<std::string> get_log_keywords()
{
    Node log_node;
    log::info(log_node,"","");
    log::optional(log_node,"","");
    log::error(log_node,"","");
    log::validation(log_node,false);
    return log_node.child_names();
}

const std::vector<std::string> LOG_KEYWORDS = get_log_keywords();
```


Node needs to use the default allocator, which is in a map, but there is no guarantee that is inited yet. 

Knowing is most of the battle here, I'll work on a solution.

cyrush:
I fixed the crash on static object init, however we still have a crash on exe clean up (static destruction)

This one is going to be harder to understand and mitigate. 

```
 	t_blueprint_mesh_generate.exe!std::_Tree<std::_Tmap_traits<__int64,void (__cdecl*)(void *),std::less<__int64>,std::allocator<std::pair<__int64 const ,void (__cdecl*)(void *)>>,0>>::_Find_lower_bound<__int64>(const __int64 & _Keyval) Line 1607	C++
 	t_blueprint_mesh_generate.exe!std::map<__int64,void (__cdecl*)(void *),std::less<__int64>,std::allocator<std::pair<__int64 const ,void (__cdecl*)(void *)>>>::_Try_emplace<__int64 const &>(const __int64 & _Keyval) Line 177	C++
 	t_blueprint_mesh_generate.exe!std::map<__int64,void (__cdecl*)(void *),std::less<__int64>,std::allocator<std::pair<__int64 const ,void (__cdecl*)(void *)>>>::operator[](const __int64 & _Keyval) Line 325	C++
 	t_blueprint_mesh_generate.exe!conduit::utils::conduit_free(void * ptr, __int64 allocator_id) Line 181	C++
 	t_blueprint_mesh_generate.exe!conduit::Node::release() Line 16201	C++
 	t_blueprint_mesh_generate.exe!conduit::Node::cleanup() Line 16222	C++
 	t_blueprint_mesh_generate.exe!conduit::Node::~Node() Line 119	C++
 	[External Code]	
 	t_blueprint_mesh_generate.exe!conduit::Node::release() Line 16187	C++
 	t_blueprint_mesh_generate.exe!conduit::Node::cleanup() Line 16222	C++
 	t_blueprint_mesh_generate.exe!conduit::Node::~Node() Line 119	C++
 	[External Code]	
 	t_blueprint_mesh_generate.exe!conduit::Node::release() Line 16187	C++
 	t_blueprint_mesh_generate.exe!conduit::Node::cleanup() Line 16222	C++
 	t_blueprint_mesh_generate.exe!conduit::Node::~Node() Line 119	C++
 	[External Code]	
 	t_blueprint_mesh_generate.exe!_execute_onexit_table::__l2::<lambda>() Line 206	C++
 	t_blueprint_mesh_generate.exe!__crt_seh_guarded_call<int>::operator()<void <lambda>(void),int <lambda>(void) &,void <lambda>(void)>(__acrt_lock_and_call::__l2::void <lambda>(void) && setup, _execute_onexit_table::__l2::int <lambda>(void) & action, __acrt_lock_and_call::__l2::void <lambda>(void) && cleanup) Line 204	C++
 	t_blueprint_mesh_generate.exe!__acrt_lock_and_call<int <lambda>(void)>(const __acrt_lock_id lock_id, _execute_onexit_table::__l2::int <lambda>(void) && action) Line 975	C++
 	t_blueprint_mesh_generate.exe!_execute_onexit_table(_onexit_table_t * table) Line 231	C++
 	t_blueprint_mesh_generate.exe!common_exit::__l2::<lambda>() Line 226	C++
 	t_blueprint_mesh_generate.exe!__crt_seh_guarded_call<void>::operator()<void <lambda>(void),void <lambda>(void) &,void <lambda>(void)>(__acrt_lock_and_call::__l2::void <lambda>(void) && setup, common_exit::__l2::void <lambda>(void) & action, __acrt_lock_and_call::__l2::void <lambda>(void) && cleanup) Line 224	C++
 	t_blueprint_mesh_generate.exe!__acrt_lock_and_call<void <lambda>(void)>(const __acrt_lock_id lock_id, common_exit::__l2::void <lambda>(void) && action) Line 975	C++
 	t_blueprint_mesh_generate.exe!common_exit(const int return_code, const _crt_exit_cleanup_mode cleanup_mode, const _crt_exit_return_mode return_mode) Line 259	C++
>	t_blueprint_mesh_generate.exe!exit(int return_code) Line 294	C++
```

cyrush:
Trick I tried was not tricky enough. Still needs work.

cyrush:
We now avoid static object init issues for the alloc and free function maps

To avoid static destruction order dependence (which appeared on windows only with static linking) -- we use a singleton with a never destructed instance.

cyrush:
We now avoid static object init issues for the alloc and free function maps.

To avoid static destruction order dependence (which appeared on windows only with static linking) -- we use a singleton with a never destructed instance.

cyrush:
Looks like we are ready.

@xjrc @mclarsen any final requests before we merge?

PR #661, opened 2020-11-24
war on whitespace

mclarsen:


xjrc:
Thanks for doing this, @mclarsen!

xjrc (approved):

mclarsen:
I needed to do this so my other pull request could be readable. I could do the entire source tree.

xjrc:
If it isn't too much trouble, please feel free; however, we don't have strict enforcement against trailing whitespace, so I fear that the problem will creep back in again without CI enforcement. That being said, I'm quite happy to update the whole tree as we'll be able to enjoy minimal whitespace at least for a time.

PR #671, opened 2020-12-18
add Node::describe()

cyrush:
requested by @mclarsen 

add a describe method, like r or pandas data frames provide. 

for each leaf, it provides summary stats, but also a truncated display of the values.

For larger datasets, using the describe output is helpful for debugging and understand vs wall of text from standard to yaml. 

TODO:

[x] settle on `count` vs `number_of_elements`
[x] add to python interface

### DIFF
--- a/src/libs/conduit/conduit_data_array.cpp
+++ b/src/libs/conduit/conduit_data_array.cpp
@@ -280,6 +280,81 @@ DataArray<T>::diff_compatible(const DataArray<T> &array, Node &info, const float
 }
 
 //---------------------------------------------------------------------------// 
+///
+/// Summary Stats Helpers
+///
+//---------------------------------------------------------------------------// 
+
+//---------------------------------------------------------------------------// 
+template <typename T>
+T
+DataArray<T>::min()  const
+{
+    T res = std::numeric_limits<T>::max();
+    for(index_t i = 0; i < number_of_elements(); i++)
+    {
+        const T &val = element(i);
+        if(val < res)
+        {
+            res = val;
+        }
+    }
+
+    return res;
+}
+
+//---------------------------------------------------------------------------// 
+template <typename T>
+T
+DataArray<T>::max()  const
+{
+    T res = std::numeric_limits<T>::min();
+    for(index_t i = 0; i < number_of_elements(); i++)
+    {
+        const T &val = element(i);
+        if(val > res)
+        {
+            res = val;
+        }
+    }
+
+    return res;
+}
+
+
+//---------------------------------------------------------------------------// 
+template <typename T>
+T
+DataArray<T>::sum()  const
+{
+    T res =0;
+    for(index_t i = 0; i < number_of_elements(); i++)
+    {
+        const T &val = element(i);
+        res += val;
+    }
+
+    return res;
+}
+
+//---------------------------------------------------------------------------// 
+template <typename T>
+float64
+DataArray<T>::mean()  const
+{
+    float64 res =0;
+    for(index_t i = 0; i < number_of_elements(); i++)
+    {
+        const T &val = element(i);
+        res += val;
+    }
+
+    res = res / float64(number_of_elements());
+    return res;
+}
+
+
+//---------------------------------------------------------------------------// 
 template <typename T>
 std::string 
 DataArray<T>::to_string(const std::string &protocol) const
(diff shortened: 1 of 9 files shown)


cyrush:
Added python support. 

For prior `count` vs `number_of_elements` quandry:

`count` is used much more widely in this context, so I think we should stick with it.

cyrush:
@mclarsen @xjrc can you give this a quick review when you get a chance?

src/libs/conduit/conduit_data_array.cpp
```diff
@@ -1436,6 +1511,135 @@ DataArray<T>::compact_elements_to(uint8 *data) const
...
+        if( (threshold % 2) > 0)
+        {
+            bottom++;
+        }
+
+        if(nele > 1)
+            os << "[";
+
+        bool first = true;
+        bool done  = false;
```
xjrc: It seems to me like something weird will happen when the list is empty as the `while` loop will evaluate at least once; thus, I think this should be `bool done = nele == 0;`.

src/libs/conduit/conduit_data_array.cpp
```diff
@@ -1436,6 +1511,135 @@ DataArray<T>::compact_elements_to(uint8 *data) const
...
+
+        if( (threshold % 2) > 0)
+        {
+            bottom++;
+        }
+
+        if(nele > 1)
+            os << "[";
+
+        bool first = true;
```
xjrc: To reduce the number of variables that we need to keep track of, would it make sense to use `idx == 0` as the check for the first element instead of a separate `first` variable?

src/tests/conduit/t_conduit_array.cpp
```diff
@@ -609,6 +609,60 @@ TEST(conduit_array, print_bells_and_whistles)
...
+    EXPECT_EQ(v,"[1, 2, 3, 4, 5]");
+
+    v = va_int64.to_summary_string(2);
+    std::cout << v << std::endl;
+    EXPECT_EQ(v,"[1, ..., 5]");
+
+    v = va_int64.to_summary_string(3);
+    std::cout << v << std::endl;
+    EXPECT_EQ(v,"[1, 2, ..., 5]");
+
```
xjrc, line 665: I think it'd be good to add some more test cases here for the corner cases, e.g. when `threshold` is bigger than the array size, when the array is empty, etc.

src/tests/conduit/t_conduit_node.cpp
```diff
@@ -1162,3 +1162,60 @@ TEST(conduit_node, add_child)
...
+    EXPECT_EQ(d["f/min"].to_float(),1.0);
+
+    EXPECT_EQ(d["a/max"].to_int(),5);
+    EXPECT_EQ(d["b/max"].to_int(),3);
+    EXPECT_EQ(d["c/max"].to_int(),6);
+    EXPECT_EQ(d["d/max"].to_int(),7);
+    EXPECT_EQ(d["e/max"].to_int(),12);
+    EXPECT_EQ(d["f/max"].to_float(),7.0);
+
+    EXPECT_EQ(d["g/mean"].to_float(),3.0);
```
xjrc, line 1209: Just so we have test cases for some of the `int` arrays, I think adding `EXPECT_EQ(d["a/mean"].to_float(),3.0);` and `EXPECT_EQ(d["b/mean"].to_float(),2.0);` here would be beneficial.

src/libs/conduit/conduit_node.hpp
```diff
@@ -3708,10 +3708,18 @@ class CONDUIT_API Node
...
     /// TODO: this is inefficient w/o move semantics, but is very
     /// convenient for testing and example programs.
     Node             info() const;
 
+    ///
+    /// describe() creates a node that replaces each leaf with
+    ///  descriptive statistics (count, mean, min, max) and a string repd 
+    ///  values summary
+    void             describe(Node &nres) const;
+    void             describe(const Node &opts, Node &nres) const;
```
xjrc, line 3721: Could we also add `to_summary_string` and `to_summary_string_default` to the `conduit::Node` type? This would make it much easier to print summary information from debugging tools like `gdb`.

xjrc (changes requested):
These new functions look great Cyrus; they should make debugging large Conduit trees much easier! I have a few minor questions regarding tests and interface changes, but once we hash those out I think this is good to go.

src/libs/conduit/conduit_node.hpp
cyrush: This is an awesome idea, I will address in a new PR, here is an issue associated with this 

https://github.com/LLNL/conduit/issues/674

src/libs/conduit/conduit_data_array.cpp
cyrush: good catch, i made this change.

src/libs/conduit/conduit_data_array.cpp
cyrush: agreed, done

src/tests/conduit/t_conduit_array.cpp
cyrush: thanks for keeping me honest -- i added several mode

src/tests/conduit/t_conduit_node.cpp
cyrush: agreed, I added mean checks for all arrays

xjrc (approved):
Thanks for all of the fixes!

Issue #672, opened 2021-01-04
Overrite existing hdf5 files


mclarsen:
Currently, conduit throws an error when an hdf5 file already exists, which is a change from the previous behavior. This new change can cause MPI deadlock when coordinating writes among many mpi ranks. I think the default behavior should be to overwrite, with an option to complain.

cyrush:
resolved in #673

PR #673, opened 2021-01-04
add generic io handle append and truncate options, add bp save_mesh t…

cyrush:
…o provide replay save (truncate) semantics


cyrush:
resolves #672

PR #677, opened 2021-01-12
enhancements to matset to_silo

cyrush:


cyrush:
@xjrc let me know what you think, FYI -- I want to improve some of the tests

cyrush:
One issue I need to resolve, the size of the `matlist` ends up being too big in the uni-buffer case:


```
topology: "topo"
material_map: 
  circle_a: 1
  circle_b: 2
  circle_c: 3
  background: 0
volume_fractions: [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 0.333333333333333, 0.333333333333333, 0.333333333333333, 1.0, 1.0, 1.0, 0.5, 0.5, 1.0]
material_ids: [0, 0, 0, 0, 0, 0, 3, 0, 0, 1, 1, 2, 3, 2, 0, 1, 1, 2, 2]
sizes: [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 3, 1, 1, 1, 2, 1]
offsets: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 13, 14, 15, 16, 18]


topology: "topo"
material_map: 
  circle_a: 1
  circle_b: 2
  circle_c: 3
  background: 0
matlist: [1, 1, 1, 1, 1, 1, 4, 1, 1, 2, -1, 3, 1, 2, -4, 3, 0, 0, 0]
mix_next: [2, 3, 0, 5, 0]
mix_mat: [1, 2, 3, 1, 2]
mix_vf: [0.333333333333333, 0.333333333333333, 0.333333333333333, 0.5, 0.5]
```
(see the extra three `0`s)

I think this value is coming from:
```

        blueprint::o2mrelation::O2MIterator mat_iter(mat_vfs);
        matset_num_elems = mat_iter.elements(O2MIndex::ONE);
```

I could just use the number of elements of the `offsets` or `sizes`, but I want to make sure the O2M stuff is working as we expect.

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -207,6 +200,7 @@ to_silo(const conduit::Node &n,
             }
         }
 
+        // TODO, IS THIS CORRECT?
```
xjrc: Looking over this again, I think there's one error in the logic of this branch, but the code at the bottom here is correct. The whole aim of this branch is to put element-dominant material sets into an O2M-esque structure (which is optional, but the most general form of representation) so that the number of elements can be counted in a uniform way (i.e. via `O2MIterator::elements(O2MIndex::ONE)`.

That part that needs fixing is line 188 (i.e. `mat_vfs["values"].set_external(n["volume_fractions"]);`); this line should be changed to `mat_vfs.set_external(n);`, which will allow for the code to pick up an O2M in the unibuffer case.

xjrc (changes requested):
Looks good Cyrus, but I think there's still one thing incorrect with my initial implementation; please take a look at my comment and let me know what you think. Once we get that issue resolved, I think we're ready to merge this.

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
cyrush: that was the key, thanks for helping figure it out!!

xjrc (approved):

PR #680, opened 2021-01-15
add conduit::utils::format

cyrush:
Adds `conduit::utils::format` methods that support fmt style formatting with args encoded in conduit::Nodes. 

While generally useful, the primary aim of these functions is to help us support for non-trivial domain maps (https://github.com/LLNL/conduit/issues/476)

- [x] add basic docs
- [x] add python support


cyrush:
@xjrc @agcapps was hoping for a quick review from someone else, but the update of the docs outputs includes a lot of noise. The most helpful thing would be to read the new docs and see if they make sense. I suspect the easiest way to do so will be after this PR is merged :-).

PR #684, opened 2021-01-21
use new uberenv + spack

cyrush:


cyrush:
almost there, but currently blocked by:
https://github.com/spack/spack/issues/22530

cyrush:
Error in Docker test:

```
ImportError: /conduit/uberenv_libs/spack/opt/spack/linux-ubuntu18.04-broadwell/gcc-7.5.0/python-3.8.8-rcva2ue7focsi7wr3ayizkg327hrcmfy/lib/python3.8/site-packages/mpi4py/MPI.cpython-38-x86_64-linux-gnu.so: undefined symbol: MPI_UNWEIGHTED
```

fixed, there were two mpis in play due to a mixup with how the container was setup.

cyrush:
error in other tests:

```
[activating dependent packages] 
[ERROR: failed to find package named 'uberenv-conduit']
```

uberenv shouldn't be trying to activate this package, looking into it ... 

This is confusing b/c the error is *not* related to activating, it happens later. 
Adding some better status to uberenv to avoid future tail chasing.

cyrush:
this is ready + ci test for newer cmake in #747 requires this, so I am merging.

PR #696, opened 2021-02-05
generate_partition for index of multi-domain meshes

nselliott:
generate_partition automatically generates the `domain_partition_map` described in issue #476 for parallel multi-domain meshes.  This is extracted from the polygonal/polyhedral PR because it's orthogonal to the main focus there.  WIP because a test needs to be added for this.


cyrush:
Looks like a good path.

A few thoughts:

Generating the domain partition map is generally useful for the in-memory case.  How should we handle that -- does gen_part fully meet that need (It may)

The map from domains to files may not match the map from domains to ranks, so we might need to keep gen_part logic decoupled from gen_indnex (or make it optional some how)

nselliott:
@cyrush Yes, this generate_partition should be callable from elsewhere in the code if you just need it in memory.  Do you have a use case in mind?

For your second point, it sounds like the output of generate_partition needs to be processed through another indirection map (rank-to-file).  Perhaps there should be an option added to generate_partition so that it can optionally produce a domain-to-file map instead of its current domain-to-rank.

cyrush:
In #476 cases shown a  `partition` was a generic construct for the index pattern, but this is more specialized. 

Should we rename `generate_partition`  to `generate_domain_to_rank_map()`  to be more explicit? 

We also want this function to work if each mpi task presents a single domain, not just if they present multi domain. 
(Restated -- Non domain overloaded cases can present single domain per MPI task)

For the index itself, the end goal should be the `partion_map` + `partition_pattern`.  I don't think we need an explicit `partiton_size`. The new form is closer to self describing. 

To keep things orthogonal -- instead of always creating this in the index, can folks can augment the index?

How about this as a strawman: (needs improvement) 

For the case where the partition of domains across files will match the partition of domains across MPI Ranks:

```
conduit::blueprint::mpi::mesh::generate_index(mesh,
         "",
        bp_index["mesh"],
        MPI_COMM_WORLD);

// add partition info
bp_index["mesh/state/partition_pattern"] = "myfile_{files:05d}.root:/my_domain_{domains:05d}/mesh"

conduit::blueprint::mpi::mesh::generate_domain_to_rank_map(mesh,
        bp_index["mesh/state/partition_map/files"],
        MPI_COMM_WORLD);

// create domains [0->N)
// maybe this can be implicit or easier 
bp_index["mesh/state/partition_pattern/domains"].set(DataType::int64(num_domains))
int64_array vals = bp_index["mesh/state/partition_pattern/domains"].value()
// loop to fill 0 - > N-1
// ...
``

**Comment by nselliott** [comment]
@cyrush I just realized that the output I create here uses a schema suggested earlier in the #476 discussion rather than the simplified file and domain maps at the end of the discussion.  So I need some changes here to deal with that.  The code that figures everything out is fine, but the end result needs to be presented differently.

There should be no trouble handling the single domain/rank case.  That just means that the map will look like `domain: [0, 1, 2, 3, 4, 5, ...]`

**Comment by nselliott** [comment]
I removed the generate_partition call from generate_index--at this point it leaves it to the calling code to decide what to do with the output Node.  Also I added logic in generate_partition to handle meshes that return false for `is_multi_domain`.

**Comment by xjrc** [comment]
> Should we rename `generate_partition` to `generate_domain_to_rank_map()` to be more explicit?

I like this name better; it's verbose, but far more descriptive.

File: src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp [hunk 1]
```diff
@@ -128,6 +128,61 @@ generate_index(const conduit::Node &mesh,
...
+    }
+    else
+    {
+        int par_size = relay::mpi::size(comm);
+        partition.set_dtype(conduit::DataType::int64(par_size));
+        conduit::int64_array part_array = partition.as_int64_array();
+        for (int i = 0; i < par_size; ++i)
+        {
+            part_array[i] = i;
+        }
```
xjrc: This code makes the following assumptions in the uni-domain case:

- If one processor has a uni-domain mesh, then all processors have a uni-domain mesh.
- A uni-domain mesh always has a `state/domain_id` corresponding to its MPI rank.

This is probably generally true of the codes that use Blueprint, but I don't think we should make these assumptions in general; more than likely, they won't hold in at least one case.

src/libs/blueprint/conduit_blueprint_mpi_mesh.hpp
```diff
@@ -66,6 +66,11 @@ void CONDUIT_BLUEPRINT_API generate_index(const conduit::Node &mesh,
                                           Node &index_out,
                                           MPI_Comm comm);
 
+//-------------------------------------------------------------------------
+void CONDUIT_BLUEPRINT_API generate_partition(const conduit::Node &mesh,
```
xjrc: I'd like to see this changed to `generate_domain_to_rank_map`, as per @cyrush's suggestion. What are your thoughts, @nselliott?

xjrc (changes requested):
I think we should discuss the assumptions in the uni-domain case before merging, but otherwise I think this looks good and am glad to see the parallel code being built out!

src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
nselliott: You are correct that this has an assumption that all processors have a uni-domain, so we do need to define how to handle this if it needs to handle processors without any mesh.

As for the `state/domain_id`, I wrote this with the assumption that it is _not_ there in the uni-domain case.  That value is only checked in the multi-domain case.  I suppose that as currently written it is effectively assigning an implicit `domain_id` that is equal to the rank, which I think makes sense as long as all ranks have a mesh, but might fall apart if some ranks don't.

src/libs/blueprint/conduit_blueprint_mpi_mesh.hpp
nselliott: Yes, that's fine.  Tells exactly what it is.

src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
nselliott: For the sparse uni-domain situation, could we have the map be an array that containing only the ranks that hold a domain?  So if there are 8 processors, but only the even ranks have domains, it's `[0,2,4,6]`.

cyrush: thats a good question, we may need to arrays in this case, they would have the same values:

[0,2,4,6]  -> [0,2,4,6]?

nselliott: What do you see the two arrays representing?

cyrush: a map from domain ids to the ranks they live on (in this case they are the same), but we should think about more general cases.

nselliott: I think domain ids may not exist (or may be only implicit) for this case.  You could also conceptualize my example as [0,1,2,3] -> [0,2,4,6].  That is, the four domains have a compact range of ids but they exist on a sparse set of ranks.

src/tests/blueprint/t_blueprint_mpi_mesh_verify.cpp
```diff
@@ -74,7 +87,79 @@ TEST(blueprint_mpi_smoke, ranks_with_no_mesh)
...
+
+    // all ranks should have index data.
+    EXPECT_TRUE(bp_index["mesh"].dtype().is_object());
+
+    conduit::Node domain_to_rank_map;
+    conduit::blueprint::mpi::mesh::generate_domain_to_rank_map(mesh,domain_to_rank_map,MPI_COMM_WORLD);
+
+    EXPECT_TRUE(domain_to_rank_map.dtype().is_int64());
+
+    EXPECT_EQ(domain_to_rank_map.dtype().number_of_elements(), num_domains);
```
cyrush, line 201: this looks ready, only suggestion would be to check the actual values in the unit tests, can we add that?

cyrush (approved):

src/tests/blueprint/t_blueprint_mpi_mesh_verify.cpp
nselliott: The latest commit adds this.

nselliott:
This might be ready to merge, but I don't know why the gitlab CI is failing.

src/tests/blueprint/t_blueprint_mpi_mesh_verify.cpp
cyrush: awesome, thanks!

cyrush:
don't worry about the gitlab tests, I think the train went off the rails with all of the lc gitlab changes. 

@xjrc is there anything else you think should be addressed for this PR?

xjrc:
> @xjrc is there anything else you think should be addressed for this PR?

I think it's much closer, but there are still some edge case issues. In particular, the way that this function handles empty nodes and delineates between uni- and multi-domain meshes concern me. To explain in a bit more detail:

- **Empty Nodes**: If this function follows the Blueprint pattern of assuming `blueprint::mesh::verify`-compliant inputs (like `blueprint::mesh::number_of_domains`, etc.), then the empty node check is redundant because an empty node will not return true when passed to `blueprint::mesh::verify`.
- **Mixed Meshes**: The current implementation assumes that clients use a uniform representation for their mesh nodes (i.e. all uni-domain or all multi-domain). While this may generally be true, I don't think it's safe to assume that it will always be the case. One prominent example is related to empty nodes; an otherwise uni-domain-conformant client would need to use an empty multi-domain representation in order to represent a rank without domains as this is the only `blueprint::mesh::verify`-compliant way to do so.

With those two points in mind, I think we should tweak the function a bit more before merging. I'll create a pull request on top of these changes with a fix to the issues that I outlined above and we can workshop further from there.

xjrc (approved):

xjrc:
@cyrush: Would we be able to remove Adrien's pipleine by rebasing? It looks like it's failing because of shebang paths/third-party library installs, which I don't think are related to the specific changes in this PR.

nselliott:
@xjrc Something is broken in a use case I created where I have 4 ranks with rank 0 empty and the other ranks each holding a domain in uni-domain mode.  It has to do with how it sets the domain id to the rank if there isn't a `state/domain_id` entry.  Your PR removed some code that handled this, and I had forgotten about it.  I see there isn't a conduit test for this use case--I found this with something I am developing in axom.

cyrush:
@xjrc  -- lets just ignore the LC gitlab tests for now

xjrc:
> @xjrc Something is broken in a use case I created where I have 4 ranks with rank 0 empty and the other ranks each holding a domain in uni-domain mode. It has to do with how it sets the domain id to the rank if there isn't a `state/domain_id` entry. Your PR removed some code that handled this, and I had forgotten about it. I see there isn't a conduit test for this use case--I found this with something I am developing in axom.

Would there be any way for us to get a reproducer for this particular case? I attempted to retain the `state/domain_id` logic that you implemented for uni-domain entries (i.e. [this](https://github.com/LLNL/conduit/blob/2db4c31247b66ca608678ae686a470488a579e2e/src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp#L178)), albeit in a refactored form (i.e. see [this](https://github.com/LLNL/conduit/blob/feature/nselliott/gen-partition/src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp#L136), in particular `int64 domain_id = par_rank;`). I'm sorry for breaking this, but I'm happy to put in a fix if there was some element of the behavior that I got wrong with my changes.

nselliott:
@xjrc I pushed a change that triggers the error.  You need to manually run `t_blueprint_mpi_mesh_verify` on 4 ranks.

This only happens when there is no `state/domain_id` entry for any of the domains and it assigns the rank as the domain id.  One option could be to not allow this, to require `state/domain_id` if someone wants to use `generate_domain_to_rank_map`.

The `local_map[par_rank] = -1` stuff is what I used to deal with the ranks that were empty.

xjrc:
After reading over both implementations more carefully, I think I see what the issue is. To illustrate, let's look at the problem you described earlier:

```
rX := rank with id X
dY := uni-domain with arbitrary identifier Y (no explicit 'state/domain_id')

|===================|===================|===================|===================|
|         r0        |        r1         |        r2         |        r3         |
|===================|===================|===================|===================|
|                   |  +----+           |  +----+           |  +----+           |
|                   |  | d1 |           |  | d2 |           |  | d3 |           |
|                   |  +----+           |  +----+           |  +----+           |
|===================|===================|===================|===================|
```

The latest implementation of `generate_domain_to_rank_map` crashes when processing this problem because the domain identifiers are not compact (e.g. there are 3 domains total and the compacted domain-to-rank map (i.e. `std::vector<int64> local_map(3, 0);`) has an out-of-bounds error when assigning the rank for domain 3 (i.e. `local_map[3] = 3;`)). The original implementation didn't encounter this issue because it simply skipped over the domains with missing ranks and compacted values when generating the final map (e.g. `[-1, 1, 2, 3]` became `[1, 2, 3]`). Based on my understanding of what the `domain_to_rank_map` is meant to represent (i.e. a list where a domain ID maps to that domain's host rank), both of these representations are inadequate; the actual final list should include holes if they exist (e.g. `[-1, 1, 2, 3]` for the problem above, which properly states that domain with index 0 has an invalid host rank).

Does this implementation seem reasonable given your understanding of what should happen in this case, @nselliott? I think that this is important to consider because Conduit [doesn't force domains to include `state` information](https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#state) and we need some reasonable way of producing a final mapping when they exist for some domains and not others.

nselliott:
@xjrc Yes, that makes sense.

cyrush:
Looks like everything is worked and ready? 

From this process, are they any docs changes we should make (we can add issues as follow on, don't need to do in this PR) ?

xjrc:
> Looks like everything is worked and ready?
> 
> From this process, are they any docs changes we should make (we can add issues as follow on, don't need to do in this PR) ?

Yes, we'll need to update the documentation to indicate that empty nodes are interpreted as empty multi-domain meshes, but I agree that this can be a separate PR. I've added an issue for this; see #742.

cyrush:
@xjrc thanks! @nselliott just want to double check you are ready, let me know and we can merge.

nselliott:
@cyrush I have one thing I want to verify.  I should get to it before the end of the day.

nselliott:
@cyrush @xjrc I verified the thing I wanted to see, so I fully approve a merge now.

cyrush:
@nselliott awesome thanks!

PR #698, opened 2021-02-05
add node summary string methods

### LINKED ISSUE #674: add node to_summary_string
follow on to suggestion made by @xjrc in #671 

add to_summary_string at the node level

useful options:
- leaf # elements threshold
- # children threshold

cyrush:
resolves #674 

todo:
- [x] add python interface and tests
- [x] change python repr to use to_summary_string


src/libs/conduit/conduit_node.cpp
```diff
@@ -11981,6 +11981,341 @@ Node::to_string_default() const
...
+{
+    // unpack options and enforce defaults
+    index_t num_children_threshold = 7;
+    index_t num_elements_threshold = 5;
+    index_t indent = 2;
+    index_t depth = 0;
+    std::string pad = " ";
+    std::string eoe = "\n";
+
+    if(opts.has_child("num_children_threshold") &&
```
xjrc, line 12020: I know we're using `num_children` and `num_elements` because they're standard Conduit vernacular, but could we use something a bit shorter than `threshold` since it isn't as standard? Maybe something short, like `max`?

src/libs/conduit/conduit_node.cpp
```diff
@@ -11981,6 +11981,341 @@ Node::to_string_default() const
...
+std::string
+Node::to_summary_string()const
+{
+    Node opts;
+    return to_summary_string(opts);
+}
+
+//-----------------------------------------------------------------------------
+std::string
+Node::to_summary_string(const conduit::Node &opts)const
```
xjrc, line 12000: Is the idea that the other `conduit::Node::to_*` methods are moving to an options `conduit::Node` approach? This seems like a break from convention otherwise.

src/libs/conduit/conduit_node.hpp
```diff
@@ -3508,6 +3508,42 @@ class CONDUIT_API Node
...
+    std::string         to_summary_string(const conduit::Node &opts) const;
+    void                to_summary_string_stream(std::ostream &os,
+                                                 const conduit::Node &opts) const;
+    void                to_summary_string_stream(const std::string &stream_path,
+                                                 const conduit::Node &opts) const;
+
+    // NOTE(cyrush): In this case to_summary_string() is easily callable
+    // in debugging tools, but still provide the to_summary_string_default()
+    // variant to present a similar interface to other `to_` methods.
+    std::string         to_summary_string_default() const;
```
xjrc, line 3544: As an FYI, I found a way to use even the default-parameter `conduit::Node::to_*` functions in `gdb`, so I think we can delete these `conduit::Node::to_*_default` functions in 0.7.X.

src/libs/conduit/conduit_node.cpp
```diff
@@ -11981,6 +11981,341 @@ Node::to_string_default() const
...
+    std::string pad = " ";
+    std::string eoe = "\n";
+
+    if(opts.has_child("num_children_threshold") &&
+       opts["num_children_threshold"].dtype().is_number())
+    {
+       num_children_threshold = (index_t)opts["num_children_threshold"].to_int32();
+    }
+
+    if(opts.has_child("num_elements_threshold") &&
```
xjrc, line 12026: Might it be useful to include a max/threshold on tree `depth` as well? I could see it being useful to get a broad idea of the basic building blocks of a schema. Not saying we need to do it for this PR, but something to think about.

src/tests/conduit/t_conduit_to_string.cpp
```diff
@@ -44,3 +44,299 @@ TEST(conduit_to_string, simple_1)
...
+    texpect = R"ST(
+a: [0, 1, ..., 8, 9]
+b: [1, 1, ..., 8, 9]
+... ( skipped 4 children )
+g: 
+  aa: [0, 1, ..., 8, 9]
+  bb: [1, 1, ..., 8, 9]
+  ... ( skipped 1 child )
+  dd: [3, 1, ..., 8, 9]
+)ST";
```
xjrc, line 128: If regex doesn't work out, I think I'm going to use this strategy on my output checks as well.

src/tests/conduit/t_conduit_to_string.cpp
```diff
@@ -44,3 +44,299 @@ TEST(conduit_to_string, simple_1)
...
+    EXPECT_EQ(tres,texpect);
+
+    // neg or zero for thresholds should trigger default yaml case
+    opts.reset();
+    opts["num_elements_threshold"] = -1;
+    opts["num_children_threshold"] = -1;
+    opts.print();
+    tres = n.to_summary_string(opts);
+    std::cout << tres << std::endl;
+    EXPECT_EQ(tres,n.to_yaml());
```
xjrc, line 138: I hadn't considered this case; good coverage of the possible inputs!

src/libs/conduit/conduit_node.cpp
```diff
@@ -11981,6 +11981,341 @@ Node::to_string_default() const
...
+
+        if( (threshold % 2) > 0)
+        {
+            bottom++;
+        }
+
+        bool done  = (nchildren == 0);
+        int idx = 0;
+
+        while(!done)
```
xjrc, line 12138: The code in these loops is rather hard for me to follow. I think using the lambdas we're afforded by C++11 could make it much more legible, e.g.:

```
const auto to_summary = [] (const conduit::Node& n, ...)
{
    // ... //
}

for(index_t idx = 0; idx <= bottom; idx++)
{
    to_summary(node);
}

// ... skipped print logic goes here ... //

for(index_t idx = nchildren - top; idx < nchildren; idx++)
{
    to_summary(node);
}
```

Just my 2 cents here; I don't mind keeping this as is for the sake of expediency.

xjrc (approved):
I left a few comments, but I didn't find any deal-breakers. I'm looking forward to using this for debugging!

src/libs/conduit/conduit_node.cpp
cyrush: I would like to move in that direction,  number of options is getting large.

src/libs/conduit/conduit_node.hpp
cyrush: ok, lets discuss further the path post 0.7.0 release. I'll keep this here for now.

src/libs/conduit/conduit_node.cpp
cyrush: i agree, we should do that in the future.

src/libs/conduit/conduit_node.cpp
cyrush: I used `threshold` b/c that is what I settled on for the DataArray class (mirrors numpy's param name)

cyrush: https://numpy.org/doc/stable/reference/generated/numpy.set_printoptions.html

src/libs/conduit/conduit_node.cpp
cyrush: added: #700 to track this

src/libs/conduit/conduit_node.cpp
cyrush: added https://github.com/LLNL/conduit/issues/699 to track this.

src/libs/conduit/conduit_node.cpp
cyrush: added #701 to track this

src/libs/conduit/conduit_node.hpp
cyrush: added #702 to track this

src/libs/conduit/conduit_node.cpp
cyrush: (i think this is great idea)

Issue #682, opened 2021-01-20
add support for mpi-based `generate_*` functions


xjrc:
In order to simplify the user burden for calling any of the `blueprint::mesh::topology::unstructured::generate_*` functions on a multi-domain mesh, the Blueprint library should include multi-domain/MPI extensions for these functions, which would handle the non-trivial task of generating new `adjsets` entries based on the function invocation and original mesh contents. Here's a rough sketch of the new functions to be introduced:

- `blueprint::mpi::mesh::generate_{points|lines|faces}(conduit::Node& mesh, const std::string& src_topo_path, const std::string& dst_topo_path, const std::string& dst_adjset_path, conduit::Node& s2dmap, conduit::Node& d2smap)`
- `blueprint::mpi::mesh::generate_{sides|corners}(conduit::Node& mesh, const std::string& src_topo_path, const std::string& dst_topo_path, const std::string& dst_cset_path, const std::string& dst_topo_adjset_path, const std::string& dst_cset_adjset_path, conduit::Node& s2dmap, conduit::Node& d2smap)`

The sketch prototypes above deviate from their uni-domain counterparts in a few key ways, which are highlighted and explained below:

- **Destination `std::string` Over `conduit::Node`**: In the uni-domain case, the caller is given more freedom in specifying the location of a derived topology because this topology exists in isolation (or, at least, it is presumed to). In the multi-domain case, these derived topologies need to share a path to signal that they are linked across domains, hence the added constraint of a path specification instead of an arbitrary node.
- **Change in Source/Destination Maps**: Since the multi-domain interfaces process per-domain topologies, the source/destination maps need to accommodate per-domain information. Thus, the corresponding `conduit::Node`s will have the mappings embedded by domain identifier (e.g. `s2dmap={"domain0": {"values": [...]}, "domain1": {"values": [...]}}`).
- **Potential for Per-Domain Configuration**: The above interfaces can be extended to support more flexible cross-domain configuration/specification by layering an additional interface that accepts `conduit::Node` objects with per-domain entries for each spec (e.g. `src_topo_path={"domain0": "topo0", "domain1", "topo1", ...}`). This interface requires additional tooling and may become confusing for users (each `path` presents the opportunity for a disjoint per-domain mapping), but the possibility is open if required.

These interfaces/stipulations will be fleshed out more as their usage within real applications is further explored/manifested.

xjrc:
[...]

PR #713, opened 2021-02-19
make blueprint helpers available to multiple source files

xjrc:
This pull request aims to clean up and abstract the Blueprint helper code (e.g. internal code for querying information about a mesh, like its dimsnion, striding, etc.) so that it's available to source code outside of `conduit_blueprint_mesh.cpp`. This refactoring will be particularly useful when tackling #682 and finalizing #642 as it will make this helper functionality available to the `conduit_blueprint_mpi_mesh.cpp` source code.

At a more fine-grained level, this pull request includes the following modifications:

- [x] Refactor all Blueprint static containers to use initializer lists.
- [x] Create a new internal object file that is imported by both `conduit_blueprint_mesh.cpp` and `conduit_blueprint_mpi_mesh.cpp`.
- [x] Incorporate a small amount of Blueprint helper code into `conduit_blueprint_mpi_mesh.cpp` and write associated test cases.


xjrc:
@cyrush, @nselliott: I think that this PR is ready for review. Here's an overview of the main changes:

1. The Blueprint topology utility code that lived in `conduit_blueprint_mesh.cpp` has been moved to `conduit_blueprint_util_mesh.cpp`, and exposed via `conduit_blueprint_util_mesh.hpp`. This header can be used from within any Blueprint source file.
2. The `conduit_blueprint_util_o2mrelation.hpp` is an analogous module for general one-to-many utilities (currently just constant variables).
3. The `conduit::blueprint::mesh::coordset::length` and `conduit::blueprint::mesh::topology::{dims|length}` functions have been added to the public Blueprint Mesh interface.
4. Many test cases have been refactored to use the new constants/functions defined in `conduit_blueprint_util_mesh.hpp` and `conduit_blueprint_util_o2mrelation.hpp`.

This should help with #682 and some of the abstractions @nselliott has been working on. Let me know what you think!

src/libs/blueprint/conduit_blueprint_util_o2mrelation.cpp
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) Lawrence Livermore National Security, LLC and other Conduit
+// Project developers. See top-level LICENSE AND COPYRIGHT files for dates and
+// other details. No copyright assignment is required to contribute to Conduit.
+
+//-----------------------------------------------------------------------------
+///
+/// file: conduit_blueprint_util_mesh.cpp
+///
+//-----------------------------------------------------------------------------
+
```
nselliott: Since this file has no code that does anything, are you including it as a place for further additions?

xjrc: That's correct; I want the small amount of code that's in `conduit_blueprint_util_o2mrelation.hpp` and I figured that I'd include a corresponding source file just in case we ever need to put anything in it.

nselliott (approved):
Thanks @xjrc , this is good.

cyrush:
Joe looks good,  One overarching question I have 


Should we use:

`util::mesh` or `mesh::utils`

I feel like a case could be made for `mesh::utils` -- they are utils that apply to things that confirm to the mesh bp?

Maybe the same should apply to o2m, but I didn't catch it there?

Want to make sure we consider it.

xjrc:
> Joe looks good, One overarching question I have
> 
> Should we use:
> 
> `util::mesh` or `mesh::utils`
> 
> I feel like a case could be made for `mesh::utils` -- they are utils that apply to things that confirm to the mesh bp?
> 
> Maybe the same should apply to o2m, but I didn't catch it there?
> 
> Want to make sure we consider it.

I like the look of `mesh::utils` better, and I think it better describes the scope of the functionality (though, some things may ultimately live at some higher `blueprint::utils` level); I will make this change.

xjrc:
@cyrush: I just remembered why I chose the current ordering: to better match with the nesting scheme used for `blueprint::mpi::mesh`. I still prefer your naming suggestion, though I suppose my question would be how does the `utils` namespace differ from the `mpi` namespace to justify the difference in ordering? If there is no substantial difference, my preference would be to reverse the order in both places (i.e. `blueprint::mesh::mpi` and `blueprint::mesh::utils`).

cyrush:
Hmm, lets ponder a bit more.

cyrush:
Still not sure what is best here. 

Note: Everything under `blueprint::mpi` exists in the mpi-enabled lib -- that's one type of logical grouping that is useful in a different way b/c it has linking implications. 

Also  what do we think about `blueprint::mesh::utils`  `blueprint::mpi::mesh::utils`? 

Honestly -- we can also consider placing those methods simply in `blueprint::mesh`, unless you are worried they will clutter the namespace? Lots of choices to consider (?)

xjrc:
> Still not sure what is best here.
> 
> Note: Everything under `blueprint::mpi` exists in the mpi-enabled lib -- that's one type of logical grouping that is useful in a different way b/c it has linking implications.

That seems to me like a reasonable factor of differentiation.

> Also what do we think about `blueprint::mesh::utils` `blueprint::mpi::mesh::utils`?
> 
> Honestly -- we can also consider placing those methods simply in `blueprint::mesh`, unless you are worried they will clutter the namespace? Lots of choices to consider (?)

I think that the former is fine, and I certainly prefer it to the latter. I'd like to separate out these functions to help identify them as utilities that have lower-level use cases and interfaces. I will change the namespace ordering.

cyrush:
sorry one more nitpick: we have `conduit::utils`, so should these be `utils`

xjrc:
> sorry one more nitpick: we have `conduit::utils`, so should these be `utils`

Point well taken; the changes have been made.

cyrush (approved):
thanks @xjrc , looks good!

xjrc:
@cyrush: Can you merge (if Travis looks okay; it failed on grabbing packages, so it's likely on their end)? It looks like I don't have the proper permissions.

cyrush:
yes, I just re-started travis.

But it is strange you can't merge, we should figure that out.

PR #709, opened 2021-02-17
Feature/relay io read write offset

kevinhkhuynh:
PR for issue #689 

TODO:

The API needs to be extended to the HDF5Handler and IOHandler class. Currently, the new feature is only implemented for the functions within the conduit_relay_io_hdf5.cpp class. I can get started on this next week.

Issue #689, opened 2021-02-03
Relay IO - generic support for read and write with offset


cyrush:
An approach to resolve:

#587 

Relay's IO Handle will provide a better interface than the top level unbound calls.

Target support for  HDF5, but support for basic io (YAML, etc) can be emulated with current in memory cached node strategy. 

Goal
--------------------------

Allow user to write a leaf from a conduit node to an HDF5 data array using H5UNLIMITED dims and an explicit offset
 
C++ User Interface example:
--------------------------

```
Node data;
data.set({0,1,2,3,4});

// open our hdf5 file for read / write
conduit::relay::io::IOHandle h;
h.open("my_file.hdf5");

// write initial data (currently supported)
h.write(data,"my_array");

// look at what we wrote
Node res;
h.read("my_array",res);
res.print();

//---------------
// New Support
//---------------

// write data to same hdf5 dataset, but at given offset  (extend the dataset)
Node opts;
opts["offset"] = 5;
h.write(data, "my_array", opts);

// look at the entire array:
h.read("my_array", res);
res.print();

// To provide a symmetric interface, also provide read with offset

// read subset of the array:
h.read("my_array", opts, res);
res.print();

// We also want a way to get the info about a given path
// this way we can probe the info to provide a useful offset in the append case
h.read_info("my_array",res);
res.print()

// there is nothing special about "appending" the aim is to support generic offset writes
// when then can be used to accomplish appending
opts["offset"] = 10 
data.set({10,20,30,40,50});
h.write(data, "my_new_array", opts);
// look at the entire file
h.read(res);
res.print();

```

Hypothetical Output
--------------------------

```
[0, 1, 2, 3, 4]

[0, 1, 2, 3, 4, 0, 1, 2, 3, 4]

[0, 1, 2, 3, 4]


dtype: int64
number_of_elements: 10

my_array: [0, 1, 2, 3, 4, 0, 1, 2, 3, 4]
my_new_array: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 20, 30, 40, 50]

```

Implementation Approach
--------------------------

Most of the logic will end up living in:

https://github.com/LLNL/conduit/blob/develop/src/libs/relay/conduit_relay_io_hdf5.cpp

These methods use hdf5 handles directly and manage all aspects of using the HDF5 api. 

Create hdf5_write + hdf5_read variants that support the offset option.

Use H5UNLIMITED when declaring leaves, work out issues related to heuristics for compact, etc.
[...]

PR #715, opened 2021-02-22
Feature/relay io read write offset io handle support

kevinhkhuynh:
PR for issue #689 

A continuation of PR #709 

1. Extended the HDF5 read/write with offset API to the HDF5Handler and IOHandler class.
2. Added a new test using the io handler for read/write with offset.


cyrush (approved):
Looks good!

PR #728, opened 2021-03-17
improve logic related to hdf5 mpi deps

cyrush:


cyrush:
resolves h5z-zfp smoke test compile issue reported here:
https://github.com/spack/spack/pull/22233

PR #731, opened 2021-03-19
fix `generate_domain_to_rank_map` for mixed-domain environments

xjrc:
The changes in this pull request contains fixes to mixed-domain issues identified in #696.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
@@ -127,20 +127,7 @@ void generate_domain_to_rank_map(const conduit::Node &mesh,
 {
     int64 par_rank = relay::mpi::rank(comm);
 
-    std::vector<const Node *> domains;
-    if(!::conduit::blueprint::mesh::is_multi_domain(mesh))
-    {
-        domains.push_back(&mesh);
-    }
-    else
-    {
-        NodeConstIterator itr = mesh.children();
-        while(itr.has_next())
-        {
-            domains.push_back(&itr.next());
-        }
-    }
-
+    std::vector<const Node *> domains = ::conduit::blueprint::mesh::domains(mesh);
     std::vector<int64> local_domains;
     for(index_t di = 0; di < (index_t)domains.size(); di++)
     {
--- a/src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
@@ -51,19 +51,11 @@ verify(const conduit::Node &n,
        conduit::Node &info,
        MPI_Comm comm)
 {
-    // some MPI tasks may not have data, that is fine
-    // but blueprint verify will fail, so if the
-    // input node is empty skip verify
-    int local_verify_ok = 0;
-    if(!n.dtype().is_empty())
-    {
-        if(conduit::blueprint::mesh::verify(n,
-                                            info))
-        {
-            local_verify_ok = 1;
-        }
-    }
+    int par_size = relay::mpi::size(comm);
 
+    // NOTE(JRC): MPI tasks without any domains should use a multi-domain
+    // format with empty contents (i.e. an empty object or list node).
+    int local_verify_ok = conduit::blueprint::mesh::verify(n, info) ? 1 : 0;
     int global_verify_ok = 0;
 
     Node n_snd, n_reduce;
@@ -72,7 +64,7 @@ verify(const conduit::Node &n,
     n_reduce.set_external(&global_verify_ok,1);
 
     relay::mpi::sum_all_reduce(n_snd, n_reduce, comm);
-    return global_verify_ok > 0;
+    return global_verify_ok == par_size;
 }
 
 //-------------------------------------------------------------------------
--- a/src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
@@ -135,76 +135,48 @@ void generate_domain_to_rank_map(const conduit::Node &mesh,
 {
     int64 par_rank = relay::mpi::rank(comm);
 
-    if (!mesh.dtype().is_empty() &&
-        ::conduit::blueprint::mesh::is_multi_domain(mesh))
+    std::vector<const Node *> domains;
+    if(!::conduit::blueprint::mesh::is_multi_domain(mesh))
     {
-
-        std::vector<conduit::int64> local_domains;
-
-        conduit::NodeConstIterator itr = mesh.children();
-        while (itr.has_next())
-        {
-            const conduit::Node &chld = itr.next();
-            if (chld.has_child("state"))
-            {
-                const conduit::Node &state = chld["state"];
-                if (state.has_child("domain_id"))
-                {
-                    conduit::int64 dom_id = state["domain_id"].as_int64();
-                    local_domains.push_back(dom_id);
-                }
-            }
-        }
-
-        Node num_local, num_global;
-        num_local.set_int64(local_domains.size());
-        num_global.set_int64(0);
-        relay::mpi::sum_all_reduce(num_local, num_global, comm);
-
-        std::vector<int64> local_map(num_global.as_int64(), 0);
-        for (auto m_itr = local_domains.begin(); m_itr != local_domains.end();
-             ++m_itr)
-        {
-            local_map[*m_itr] = par_rank;
-        }
-
-        Node local_par;
-        local_par.set_external(&local_map[0], local_map.size());
-
-        relay::mpi::max_all_reduce(local_par, domain_to_rank_map, comm);
+        domains.push_back(&mesh);
     }
     else
     {
-        int64 par_size = relay::mpi::size(comm);
-
-        std::vector<int64> local_map(par_size, 0);
-        if (mesh.dtype().is_empty())
+        NodeConstIterator itr = mesh.children();
+        while(itr.has_next())
         {
-            local_map[par_rank] = -1;
+            domains.push_back(&itr.next());
         }
-        else
-        {
-            local_map[par_rank] = par_rank; 
-        }
-
-        Node local_full;
-        local_full.set_external(&local_map[0], local_map.size());
-
-        Node global_full;
-        relay::mpi::sum_all_reduce(local_full, global_full, comm);
-
-        conduit::int64_array global_array = global_full.as_int64_array();
-        std::vector<int64> compact_map;
-        for (int i = 0; i < par_size; ++i)
-        {
-            if (global_array[i] != -1)
-            {
-                compact_map.push_back(i);
-            }
-        } 
-
-        domain_to_rank_map.set(compact_map);
     }
+
+    std::vector<int64> local_domains;
+    for(index_t di = 0; di < (index_t)domains.size(); di++)
+    {
+        const conduit::Node &domain = *domains[di];
+
+        int64 domain_id = par_rank;
+        if(domain.has_child("state") && domain["state"].has_child("domain_id"))
+        {
+            domain_id = domain["state/domain_id"].as_int64();
+        }
+        local_domains.push_back(domain_id);
+    }
+
+    Node num_local, num_global;
+    num_local.set_int64(local_domains.size());
+    num_global.set_int64(0);
+    relay::mpi::sum_all_reduce(num_local, num_global, comm);
+
+    std::vector<int64> local_map(num_global.as_int64(), 0);
+    for(auto m_itr = local_domains.begin(); m_itr != local_domains.end(); ++m_itr)
+    {
+        local_map[*m_itr] = par_rank;
+    }
+
+    Node local_par;
+    local_par.set_external(&local_map[0], local_map.size());
+
+    relay::mpi::max_all_reduce(local_par, domain_to_rank_map, comm);
 }
 
 //-----------------------------------------------------------------------------
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -769,19 +769,26 @@ verify_multi_domain(const Node &n,
     bool res = true;
     info.reset();
 
-    if(!n.dtype().is_object() && !n.dtype().is_list())
+    if(!n.dtype().is_object() && !n.dtype().is_list() && !n.dtype().is_empty())
     {
-        log::error(info, protocol, "not an object or a list");
+        log::error(info, protocol, "not an object, a list, or empty");
         res = false;
     }
     else
     {
-        NodeConstIterator itr = n.children();
-        while(itr.has_next())
+        if(n.dtype().is_empty() || n.number_of_children() == 0)
         {
-            const Node &chld = itr.next();
-            const std::string chld_name = itr.name();
-            res &= verify_single_domain(chld, info[chld_name]);
+            log::info(info, protocol, "is an empty mesh");
+        }
+        else
+        {
+            NodeConstIterator itr = n.children();
+            while(itr.has_next())
+            {
+                const Node &chld = itr.next();
+                const std::string chld_name = itr.name();
+                res &= verify_single_domain(chld, info[chld_name]);
+            }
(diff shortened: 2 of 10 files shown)


src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
```diff
@@ -135,76 +135,48 @@ void generate_domain_to_rank_map(const conduit::Node &mesh,
...
+        const conduit::Node &domain = *domains[di];
+
+        int64 domain_id = par_rank;
+        if(domain.has_child("state") && domain["state"].has_child("domain_id"))
         {
-            local_map[par_rank] = par_rank; 
+            domain_id = domain["state/domain_id"].as_int64();
         }
+        local_domains.push_back(domain_id);
+    }
```
nselliott, line 142: The logic here is not correct for multi-domain meshes.  Multi-domain meshes need the loop over children, as in the one starting on line 145 in the left column.

nselliott (changes requested):
@xjrc I don't see how this change produces different results for the cases when there are empty ranks.  This type of mesh is tested in the `ranks_with_no_mesh` test in `t_blueprint_mpi_mesh_verify.cpp` and it works with some ranks being empty and others holding the mesh in uni-domain format.

Additionally, as noted above, this breaks the general multi-domain case, as it removes the loop over children that finds all of the domains on a rank in the multi-domain format.

src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
xjrc: The logic on line 145 is being applied to multi-domain meshes; the original conditional has been changed to check for uni-domain meshes. More specifically, the logic now looks like this:

```
if(!::conduit::blueprint::mesh::is_multi_domain(mesh))
{
    // uni-domain case
}
else
{
    // multi-domain case
}
```

xjrc:
> @xjrc I don't see how this change produces different results for the cases when there are empty ranks. This type of mesh is tested in the `ranks_with_no_mesh` test in `t_blueprint_mpi_mesh_verify.cpp` and it works with some ranks being empty and others holding the mesh in uni-domain format.

I didn't notice this before, but you have a good point here. Looking into this a bit more deeply, it appears as though my real concern isn't so much with your implementation as it is with the current implementation of `blueprint::mpi::mesh::verify`.

In particular, this function currently returns `true` when *any* processor has a valid Blueprint mesh instead of *all* processors. In my mind, this is a fundamentally flawed way of implementing this check. As it stands, if all but one rank has an ill-formed Blueprint mesh (e.g. all missing `coordset` entries), this function will still return `true`. Instead, I think we should force ranks with no data to use an empty object/list Node as these are valid Blueprint meshes (i.e. empty multi-domain) and they would allow us to write a more robust all-or-nothing version of `blueprint::mpi::mesh::verify`. I'll illustrate what I mean with my next set of changes.
 
> Additionally, as noted above, this breaks the general multi-domain case, as it removes the loop over children that finds all of the domains on a rank in the multi-domain format.

I don't know that this is true; in fact, this source code still passes the `multi_domain` MPI test case. If I'm missing something, please let me know and we can have an analogue chat to figure things out.

nselliott:
> 
> 
> > @xjrc I don't see how this change produces different results for the cases when there are empty ranks. This type of mesh is tested in the `ranks_with_no_mesh` test in `t_blueprint_mpi_mesh_verify.cpp` and it works with some ranks being empty and others holding the mesh in uni-domain format.
> 
> I didn't notice this before, but you have a good point here. Looking into this a bit more deeply, it appears as though my real concern isn't so much with your implementation as it is with the current implementation of `blueprint::mpi::mesh::verify`.
> 
> In particular, this function currently returns `true` when _any_ processor has a valid Blueprint mesh instead of _all_ processors. In my mind, this is a fundamentally flawed way of implementing this check. As it stands, if all but one rank has an ill-formed Blueprint mesh (e.g. all missing `coordset` entries), this function will still return `true`. Instead, I think we should force ranks with no data to use an empty object/list Node as these are valid Blueprint meshes (i.e. empty multi-domain) and they would allow us to write a more robust all-or-nothing version of `blueprint::mpi::mesh::verify`. I'll illustrate what I mean with my next set of changes.

I see what you're saying here.  The reduction in the parallel `verify` ends up being a check that at least one rank has a verified Blueprint mesh.  That could be changed to require all ranks have either a verified mesh or valid empty data.

> 
> > Additionally, as noted above, this breaks the general multi-domain case, as it removes the loop over children that finds all of the domains on a rank in the multi-domain format.
> 
> I don't know that this is true; in fact, this source code still passes the `multi_domain` MPI test case. If I'm missing something, please let me know and we can have an analogue chat to figure things out.

In my branch, a multi-domain mesh goes into a loop over children to find all of the domains, and in particular it finds the value of `domain_id` for every domain.  In this PR, the multi-domain mesh is simply added to the `domains` vector via `domains.push_back(&mesh)`.  Thus when it gets to the loop at line 153, `domains` is guaranteed to be of size 1, and there is no way that loop is going to find the potentially numerous values of `domain_id` that could be contained in the mesh.

I haven't built the PR branch to see it all in action.  Did it work with the latest version of the tests that I added last week?

nselliott:
Oh, I just realized my misunderstanding--I missed the `'!'` in the `if-else` logic, so I thought the first block was multi-domain and the else block was single-domain, which is exactly backwards.  I need to look at this again with a correct understanding.  Sorry about that.

cyrush:
> I see what you're saying here. The reduction in the parallel verify ends up being a check that at least one rank has a verified Blueprint mesh. That could be changed to require all ranks have either a verified mesh or valid empty data.

Agreed, we should make this change.

xjrc:
> @nselliott: I see what you're saying here. The reduction in the parallel `verify` ends up being a check that at least one rank has a verified Blueprint mesh. That could be changed to require all ranks have either a verified mesh or valid empty data.

Yes, precisely; I've just changed the implementation to require all ranks to have a valid mesh for a `true` return from `verify`. A rank without any assigned domains can just present an empty object or list node; see the updated code for the `ranks_with_no_mesh` test for details.

> @cyrush: Agreed, we should make this change.

Done; please give it a look and let me know what you think.

nselliott (approved):
Again, sorry for that misunderstanding.  When I parse the logic correctly, everything makes sense.  Also the changes to `verify` look good.

cyrush:
empty itself should still be valid --  in addition to an empty object or list, is that the case with the new logic?

xjrc:
> empty itself should still be valid -- in addition to an empty object or list, is that the case with the new logic?

No, an empty node will be newly rejected by this logic. The reason for this is that the base function `blueprint::mesh::verify` rejects empty nodes in following with the definitions of uni-domain and multi-domain meshes. If a rank has no mesh data, it seems easy enough for that rank to list its mesh as an empty multi-domain mesh (i.e. an empty list or object).

cyrush:
Ok, I don't think that is how folks publish data to ascent currently.

If they don't have data on an MPI task, they would just pass:
```
Node data;
```

I think empty in an mpi context should be valid, you can have a single MPI task with a valid domain and the others empty.

Your logic  does match the single domain case.  Life is too confusing :-(

xjrc:
> I think empty in an mpi context should be valid, you can have a single MPI task with a valid domain and the others empty.
> 
> Your logic does match the single domain case. Life is too confusing :-(

The only strong stance I'll take on this is on interface consistency: either both `blueprint::mesh::verify` and `blueprint::mpi::mesh::verify` should accept empty nodes, or both should reject them. I'm partial to rejecting empty nodes because (1) empty nodes are hard for me to justify logically as a valid Blueprint mesh and (2) allowing empty nodes provides surface area for mistakes (e.g. passing `info` instead of `mesh` to `verify`).

Ultimately, I feel like this comes down to a tradeoff between slight convenience (i.e. `Node data;` vs `Node data(DataType::object())`) and unlikely potential error (i.e. `verify(info)` vs. `verify(mesh)`). Given the rather low marginal cost/benefit in either case, I'm open to going either way on accepting/rejecting empty nodes.

**EDIT**: After mulling this over a bit, it also occurred to me that accepting empty nodes as valid Blueprint meshes complicates a lot of our internal handling functions. In particular, in addition to checking for uni/multi-domain cases, these functions will all need to adapt to handle the empty case as well. This is another fairly minor point, but an important one to consider I think.

cyrush:
`blueprint::mesh::verify` and `blueprint::mpi::mesh::verify` are a bit different since the MPI case has to handle ranks with no data.  It's that across all MPI ranks there has to be a valid domain somewhere.  If that's the case, there is a mesh that can be processed. 

I think of verify == true as the all clear sign the rest of the interface can be used. 

I guess I am a little surprised that an empty object or list in the non mpi case are valid.  Number of domains does work (returns zero), but I am not sure what else useful could be done with it?

In practice:
- For the non mpi case, I don't think we are relying on blueprint verify to return true when an empty object or list is passed. 
(I doubt this case is exercised by clients of blueprint)

- For the mpi case, I do think we are relying on blueprint verify to return true if some MPI rank has a valid domain (including the case where no data is passed via an empty node on some ranks)
(This case is exercised by clients of blueprint)

In all cases: empty objects, empty lists, empty can only signal no data. 

It seems that a consistent answer is that  verify == true implies that here is a `coordset` (the min entry for the mesh blueprint) somewhere.  Or that a coordset is optional.

xjrc:
The assertion that a mesh must contain at least some data is new to me; I've always been under the impression that an empty mesh is also a mesh. In fact, this "empty is valid" interpretation is quite useful as it seamlessly enables the "multi-rank with some empty ranks" use case at the center of this contention. Without it, we must introduce additional complexity, like allowing empty nodes but only in the parallel case.

If we accept the "empty is invalid" interpretation, then I would ask why [Conduit supports data arrays of length 0](https://github.com/LLNL/conduit/blob/develop/src/tests/conduit/t_conduit_array.cpp#L795). I feel like similar arguments on utility could be levied against such concepts (e.g. what useful thing can we do with an empty list?), with similar justifications (e.g. sometimes you simply have no data, and providing a uniform interface simplifies things). Perhaps I am missing something; I invite your comments.

cyrush:
I think empty is fine as long as it's clear that coordsets (and topos) are optional, and that we change the docs to give guidance if an empty mesh is valid case. 

Right now we state: one or more for both coordsets and topos. 
```
The Blueprint protocol defines a single-domain computational mesh using one or more Coordinate Sets (via child coordsets), one or more Topologies 
```

This is the reason why I am surprised they are valid.  But we can change the docs.  We need statements about empty and something like since an empty mesh is valid, always check number of domains to see if you have data.  In practice for many cases folks already have logic for single vs multi domain, so it wouldn't be hard to adapt that logic. 

For consistency, I think an empty node should also verify as true. I don't see a case for an empty list or empty object to be treated different than an empty node (but I think an empty node currently verifies as false).

my statements about the mpi vs non mpi cases above are about how I gauge things are being leveraged, just so we understand the implications of changes.

xjrc:
Okay, to summarize some of the points we've discussed so far:

### The Problem ###

- MPI codes with uni-domain meshes want to be able to present one or more ranks with no data. The current solution is to provide an empty node for such ranks.
- The `blueprint::mesh::verify` function doesn't recognize an empty node as a valid specification. To account for this, `blueprint::mpi::mesh::verify` returns true if any non-empty node has a true return from `blueprint::mesh::verify` among all ranks, which can cause bad domains to sneak through.
- We need an implementation of `blueprint::mpi::mesh::verify` that respects Blueprint's API guarantees (i.e. if `a::b::c::verify(n)` works, then `a::b::c::d(n, ...)` can be called).

### Potential Solutions ###

1. Modify `blueprint::mesh::verify` to recognize an empty node as a multi-domain mesh with no entries.
    - Retains backward-compatibility with current MPI uni-domain users.
    - Enables `blueprint::mpi::mesh::verify` to return true if all meshes on all ranks return true from `blueprint::mesh::verify`, satisfying the Blueprint API guarantee.
    - Increases the error surface area for users by allowing for more potential false positive verifications (e.g. `blueprint::mesh::verify(info)`).
2. Others?

### Conclusion ###

We need to choose one solution that will allow uni-domain MPI codes to painlessly communicate/aggregate domains while seamlessly supporting a sensible solution/implementation for the `blueprint::mpi::mesh::verify` function.

@nselliott: If you've been watching this discussion, do you have any thoughts on how we should solve this issue?

nselliott:
@xjrc Approaching this from the perspective of a user who will be constructing these meshes in application codes, the ability to leave a Node empty and have it pass is easier to deal with, though if there is a requirement to set its `DataType` to object, that is not a huge burden.  From the outside perspective I see this as a library documentation issue where I as a user can set up my data to meet the requirement, as long as the requirement is clearly stated.  From the inside perspective I need to make sure any of these `generate*` functions that I am writing work correctly.  In the case of the one in this pull request, it's not difficult to make it work either way.

One possible modification to your proposal:  Would it be a good idea to make `blueprint::mpi::mesh::verify` require at least one rank somewhere to have a genuine compliant domain in order to return `true`?  This could protect against the danger of having calls such as `blueprint::mpi::mesh::verify(info)` return `true`.  I know this would complicate the implementation logic a bit, since each local `blueprint::mesh::verify` call would return true, but you would still need to make the parallel result be false when there are zero domains across the full set of ranks.

xjrc:
> @xjrc Approaching this from the perspective of a user who will be constructing these meshes in application codes, the ability to leave a Node empty and have it pass is easier to deal with, though if there is a requirement to set its `DataType` to object, that is not a huge burden. From the outside perspective I see this as a library documentation issue where I as a user can set up my data to meet the requirement, as long as the requirement is clearly stated. From the inside perspective I need to make sure any of these `generate*` functions that I am writing work correctly. In the case of the one in this pull request, it's not difficult to make it work either way.

I agree with all of your points here, and I've come around on the empty Node issue. It's easy to cast an empty node as a multi-domain mesh with no domains, and this makes life simpler for application codes while not drastically complicating life for library methods.

> One possible modification to your proposal: Would it be a good idea to make `blueprint::mpi::mesh::verify` require at least one rank somewhere to have a genuine compliant domain in order to return `true`? This could protect against the danger of having calls such as `blueprint::mpi::mesh::verify(info)` return `true`. I know this would complicate the implementation logic a bit, since each local `blueprint::mesh::verify` call would return true, but you would still need to make the parallel result be false when there are zero domains across the full set of ranks.

I foresee a few problems with the "require some data in the MPI case" approach; in particular:

1. It causes a bifurcation in how the non-MPI and MPI `verify` functions work; the former will accept an empty mesh where the latter will reject it.
2. It changes `blueprint::mpi::mesh::verify` from a simple reduction (i.e. logical `&&` of all ranks' `blueprint::mesh::verify`) to a reduction with some additional logic. This criticism is in some ways isomorphic to (1), but I think the increased complexity is an important factor to highlight.

As with my prior arguments, my primary concern is consistency. In my eyes, it's important for `blueprint::mesh::verify` and `blueprint::mpi::mesh::verify` to work analogously at a high-level. I think we're all in agreement that accepting empty in the non-MPI case is a good idea (though please correct me if I'm wrong), but I don't agree with breaking this consistency in the MPI case for the sake of safety/error checking. I invite your thoughts/comments.

nselliott:
OK, well stated.  Making it consistently return `true` when empty will be fine.

xjrc:
@nselliott, @cyrush: I've updated the code base on all of our discussion. Here's a brief overview of what changed:

- `blueprint::mesh::verify` now accepts empty nodes.
- `blueprint::mpi::mesh::verify` is now just a simple map-reduce of per-processor `blueprint::mesh::verify` evaluations.
- `blueprint::mesh::domains` is a new function that returns a `std::vector` containing pointers to all of the domains in a mesh (see `blueprint::mpi::mesh::generate_domain_to_rank_map` for a use case).

nselliott (approved):
This looks good, @xjrc, I think it can be merged into my pull request branch.  I see some CI failures, but if those are real, it's probably easier to fix them within just one pull request.

xjrc:
> This looks good, @xjrc, I think it can be merged into my pull request branch. I see some CI failures, but if those are real, it's probably easier to fix them within just one pull request.

Thanks, @nselliott. The CI failures are real, and I've just attempted to rectify them with my latest change. We can merge even if the code is still broken and I'll just push future fixes to your branch.

PR #743, opened 2021-04-14
use custom namespace for rapidjson to avoid sym collision 

cyrush:


cyrush:
@xjrc  change to fix collision issue.  Is there a way to verify this fix solves the issue you saw?

xjrc:
@cyrush: Yes, I can run some tests this afternoon.

cyrush:
awesome, thanks!

xjrc (approved):
@cyrush: I tested this out and it works! Thanks for getting to this so quickly.

PR #746, opened 2021-04-19
Fix the FindMPI command so that it includes both the C and CXX

bvanessen:
compiler and linker components.


cyrush (approved):

cyrush:
thanks @bvanessen!

cyrush:
(I will merge pending ci)

bvanessen:
Really we should thank @benson31, he was the one who found it for me.

cyrush:
thanks @benson31! Just swapped to new use of FindMPI on Friday, thanks for helping us iron this out.

PR #751, opened 2021-04-28
mesh bp index gen: guard against empty mesh case

cyrush:
@xjrc  we hit a case in ascent where mesh index gen would fail when trying to fetch `coordsets` from an empty mesh. 

I think our cases was due to changes in verify now lets emptys through -- which is now expected.

I added a new test that triggers a similar case, and would fail w/o these changes.  Still not 100% sure this is best solution.


cyrush:
To avoid surprises across codes, I think that `generate_index` should throw an exception in this case, instead of creating an empty index.

cyrush:
post azure os update, need to set env vars equiv `--oversubscribe` to allow the tests to run with more mpi tasks than there are cores.

cyrush:
also heals appveyor and azure ci

PR #759, opened 2021-05-12
support hdf5 cmake system, possibly legacy

white238:
I am still testing this fix in axom but didn't want to forget to get this into conduit.


cyrush:
@white238 everything looks good here and in the related axom PR, so I am merging.
Thanks for this!

PR #764, opened 2021-05-19
Update conduit_fortran_obj.f90 to initialize node pointer to NULL.

aaroncblack:
Small change to initialize the cnode pointer to null.

GNU is not always giving the correct result of a
c_associated(cnode)

when that isn't set to null beforehand.


src/libs/conduit/fortran/conduit_fortran_obj.f90
```diff
@@ -10,12 +10,12 @@
...
 !------------------------------------------------------------------------------
     use conduit
-    use, intrinsic :: iso_c_binding, only : C_PTR
+    use, intrinsic :: iso_c_binding, only : C_PTR, C_NULL_PTR
     implicit none
     
     !--------------------------------------------------------------------------
     type node
-        type(C_PTR) cnode
+        type(C_PTR) cnode => C_NULL_PTR
```
aaroncblack:
```suggestion
        type(C_PTR) :: cnode = C_NULL_PTR
```
```suggestion
        type(C_PTR) cnode => C_NULL_PTR
```

cyrush:
Compiler still not happy :-(

Are there example patterns we can follow for this, say in something like shroud?

cyrush:
@aaroncblack is this still undermining use for you? Any place we can look for an example of how to avoid the gfortran issue?

aaroncblack:
I'll close this.  I can work around it by creating a separate bool 'is_initialized' that I check, or move the code to C++, etc.

cyrush:
thanks @aaroncblack -- if it appears in other contexts we can revisit.

PR #775, opened 2021-06-03
Fix warnings

mathstuf:


cyrush (approved):
Thanks for these fixes!

cyrush:
I'll merge pending CI.

src/libs/relay/conduit_relay_io.cpp
```diff
@@ -670,7 +670,7 @@ load_merged(const std::string &path,
 void
 load_merged(const std::string &path,
             const std::string &protocol_,
-            const Node &options,
+            const Node &/*options*/,
```
cyrush: These will be used if hdf5 is in the mix.

We can use the CONDUIT_UNUSED() macro in the `#ifdef else` clause, similar to: 

https://github.com/LLNL/conduit/blob/91e04448207bb329f9dafe1c1110d4f9303d5cca/src/libs/relay/conduit_relay_io.cpp#L613

cyrush:
merged, thanks again!

PR #779, opened 2021-06-21
conduit_datatype: add missing query APIs for datatypes from C

mathstuf:


cyrush:
thanks for adding these! 
Let me know when you have added everything you need + I'll merge pending CI.

mathstuf:
This is everything I need. I tried to do the `index_t` bit, but eventually found that the C++ backing implementation doesn't exist yet.

PR #781, opened 2021-06-22
blueprint: expose the other mesh example generation APIs to C

mathstuf:
---
Cc: @cyrush


cyrush (approved):
thanks for exposing these!

PR #789, opened 2021-07-09
test updating to blt 0.4.0

cyrush:


src/cmake/Setup3rdParty.cmake
```diff
@@ -151,4 +153,25 @@ if(H5ZZFP_DIR)
...
+# Export BLT Targets when needed
+##################################
+
+# cmake < 3.15, we use BLT's mpi target and need to export
+# it for use downstream
+if( ${CMAKE_VERSION} VERSION_LESS "3.15.0" )
+    blt_list_append(TO BLT_TPL_DEPS_EXPORTS ELEMENTS mpi IF ENABLE_MPI)
+endif()
+
+foreach(dep ${CONDUIT_BLT_TPL_DEPS_EXPORTS})
```
kennyweiss: I think there's a typo here:
```suggestion
# cmake < 3.15, we use BLT's mpi target and need to export
# it for use downstream
if( ${CMAKE_VERSION} VERSION_LESS "3.15.0" )
    blt_list_append(TO CONDUIT_BLT_TPL_DEPS_EXPORTS ELEMENTS mpi IF ENABLE_MPI)
endif()

foreach(dep ${CONDUIT_BLT_TPL_DEPS_EXPORTS})
```

cyrush: Thanks, yes this was a mistake. I pushed a fix.

src/cmake/SetupBLT.cmake
```diff
@@ -23,6 +23,10 @@ if(NOT ENABLE_FOLDERS)
     set(ENABLE_FOLDERS TRUE CACHE STRING "")
 endif()
 
+################################################################
+# Make sure BLT exports its built in targets
```
xjrc, line 27: I think this should be [built-in](https://www.merriam-webster.com/dictionary/built-in) instead.

xjrc (approved):

src/cmake/SetupBLT.cmake
cyrush: - [ ] cyrus fix this

cyrush:
end game for this PR:

- [ ] The new BLT export logic should be conditional on the version of BLT used.
- [ ] Possibly not update to to BLT 0.4.0, but add a CI test that tests it with external BLT 0.4.0

cyrush:
Closing, since next update would be to much newer version of BLT.

PR #790, opened 2021-07-09
Generate sides and map fields

JustinPrivitera:
`generate_sides_and_map_fields` calls the existing `generate_sides`, and then maps fields correctly based on the original and newly generated topologies. It does not currently support vertex associated fields and volume dependent fields.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 #### Blueprint
 - Added the `blueprint::mesh::examples::polychain` example. It is an example of a polyhedral mesh. See Mesh Blueprint Examples docs (https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#polychain) for more details.
 - Fixed a bug that was causing the `conduit::blueprint::mesh::topology::unstructured::generate_*` functions to produce bad results for polyhedral input topologies with heterogeneous elements (e.g. tets and hexs).
-- Added `blueprint::mesh::topology::unstructured::map_fields_to_generated_sides`, a companion function to `generate_sides` that takes fields from the topology before `generate_sides` and maps it onto the new topology.
+- Added a new function signature for `blueprint::mesh::topology::unstructured::generated_sides`, which performs the same task as the original and also takes fields from the original topology and maps them onto the new topology.
 
 ## [0.7.2] - Released 2021-05-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 #### Blueprint
 - Added the `blueprint::mesh::examples::polychain` example. It is an example of a polyhedral mesh. See Mesh Blueprint Examples docs (https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#polychain) for more details.
 - Fixed a bug that was causing the `conduit::blueprint::mesh::topology::unstructured::generate_*` functions to produce bad results for polyhedral input topologies with heterogeneous elements (e.g. tets and hexs).
+- Added `blueprint::mesh::topology::unstructured::map_fields_to_generated_sides`, a companion function to `generate_sides` that takes fields from the topology before `generate_sides` and maps it onto the new topology.
 
 ## [0.7.2] - Released 2021-05-19
 
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -3015,6 +3015,10 @@ map_fields_specific(const Node &poly_mesh,
             {
                 CONDUIT_ERROR("Vertex associated fields are not supported.");
             }
+        }
+
+        if (field.has_child("volume_dependent"))
+        {
             if (field["volume_dependent"].as_string() != "false")
             {
                 CONDUIT_ERROR("Volume dependent fields are not supported.");
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -3114,7 +3114,6 @@ namespace detail
 
 void
 mesh::topology::unstructured::generate_sides(const conduit::Node &topo_src,
-                                             const conduit::Node &fields_src,
                                              conduit::Node &topo_dest,
                                              conduit::Node &coordset_dest,
                                              conduit::Node &fields_dest,
@@ -3124,6 +3123,7 @@ mesh::topology::unstructured::generate_sides(const conduit::Node &topo_src,
 {
     std::string field_prefix = "";
     std::vector<std::string> field_names;
+    const Node &fields_src = (*(topo_src.parent()->parent()))["fields"];
 
     // check for existence of field prefix
     if (options.has_child("field_prefix"))
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -2963,10 +2963,10 @@ mesh::topology::unstructured::generate_sides(const Node &topo,
 //-----------------------------------------------------------------------------
 template<typename T, typename U> 
 void 
-map_field(Node &field_out, 
-          const Node &field, 
-          int num_shapes, 
-          const U *tri_to_poly)
+map_field_to_generated_sides(Node &field_out, 
+                             const Node &field, 
+                             int num_shapes, 
+                             const U *tri_to_poly)
 {
     T* values_array = field_out["values"].value();
     const T* poly_field_data = field["values"].value();
@@ -2979,10 +2979,10 @@ map_field(Node &field_out,
 
 template<typename T>
 void 
-map_fields_specific(const Node &poly_mesh,
-                    const Node &d2smap,
-                    Node &side_mesh,
-                    std::string topology)
+map_fields_to_generated_sides_template(const Node &poly_mesh,
+                                       const Node &d2smap,
+                                       Node &side_mesh,
+                                       std::string topology)
 {
     NodeConstIterator fields_itr = poly_mesh["fields"].children();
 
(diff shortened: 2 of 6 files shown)


JustinPrivitera:
One quick fix is already on the way regarding error-throwing in the function titled `map_fields_specific`

src/libs/blueprint/conduit_blueprint_mesh.hpp
```diff
@@ -351,6 +351,17 @@ namespace topology
                                                   conduit::Node &s2dmap,
                                                   conduit::Node &d2smap);
 
+        //---------------------------------------------------------------------
+        void CONDUIT_BLUEPRINT_API map_fields(const conduit::Node &poly_mesh,
```
cyrush: since this only works for mapping to sides, we should aim for a more descriptive name, how about:

`map_fields_to_generated_sides`

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -3421,6 +3421,170 @@ adjset_uniform(Node &res)
     }
 }
 
+//-----------------------------------------------------------------------------
+void 
+polychain(const index_t length, // how long the chain ought to be
+          Node &res)
```
cyrush: It looks like this might not be a fresh branch? (It also has the polychain code)?

I suggest starting a new branch fresh from develop, and replaying the commits related to mapping.

src/tests/blueprint/t_blueprint_mesh_generate_map_fields.cpp
```diff
@@ -0,0 +1,152 @@
...
+        else if (i < num_tets_in_hex + num_tets_in_triprism)
+        {
+            EXPECT_EQ(id_values[i], 1);
+        }
+        else
+        {
+            EXPECT_EQ(id_values[i], 2);
+        }
+    }
+}
```
cyrush, line 430: I think it would be good to also add separate tests for the the `map_fields_to_generated_sides` methods as well. So they can serve as another example and so test if the calling conventions are different than the all in one.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2960,6 +2960,171 @@ mesh::topology::unstructured::generate_sides(const Node &topo,
...
 
+//-----------------------------------------------------------------------------
+template<typename T, typename U> 
+void 
+map_field(Node &field_out, 
+          const Node &field, 
+          int num_shapes, 
+          const U *tri_to_poly)
+{
+    T* values_array = field_out["values"].value();
```
cyrush: one thing I forgot to mention -- there could be multi component fields here. I can discuss with you in the future.

xjrc:
I have a few general questions about this capability:

1. Is the long-term goal to have a `map_fields` function that handles arbitrary source/destination topologies generated via `generate_*`, or will there be a separate `map_fields` function for each type?
2. What if the user wants to embed the topology for `generate_sides` in the same hierarchy as the source topology (this is the case for some of our clients)? This will cause inadvertent path overwrites in the current implementation.
3. Do we really want to provide a helper for composed operations like `generate_sides_and_map_fields`? This will add a lot of interface entries for what I see as minimal benefit.

cyrush:
@xjrc 
For 1:  

I am not sure about general map fields -- I don't know how we could identify the general case without context.

That's why I want the primary workhorse here to be: `map_fields_to_generated_sides`

If we have other specific cases, I think using specific names might be the path forward.
How does that sound?

For 2:
Looking at `map_fields_to_generated_sides` -- do we just need a way to create unique field names for they newly mapped fields? 

For 3:
I guess we can remove it, that does limit the public API change just in case we overlooked something.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2960,6 +2960,171 @@ mesh::topology::unstructured::generate_sides(const Node &topo,
     blueprint::o2mrelation::generate_offsets(d2smap, info);
 }
 
+//-----------------------------------------------------------------------------
+template<typename T, typename U> 
+void 
+map_field(Node &field_out,
```
cyrush: These internal funcs should also be `map_fields_to_generated_sides` or some derivative

xjrc:
> @xjrc
> For 1:
> 
> I am not sure about general map fields -- I don't know how we could identify the general case without context.
> 
> That's why I want the primary workhorse here to be: `map_fields_to_generated_sides`
> 
> If we have other specific cases, I think using specific names might be the path forward.
> How does that sound?

I'm not thrilled as I think this strategy will lead to a drastic increase in the number of public functions (i.e. NxM where N is the number of `generate_*` functions and M is the number of mapping types (e.g. fields, materials, species, etc.)), but I may be overestimating the factors. This function will be fairly limited in use and thus easy to deprecate if we decide on a different naming/presentation scheme in the future, so I won't raise too much of a fuss about this for now.

> For 2:
> Looking at `map_fields_to_generated_sides` -- do we just need a way to create unique field names for they newly mapped fields?

Yes, that's correct. I believe that you recently added some string formatting capabilities to Conduit; perhaps we could leverage some of that technology?
 
> For 3:
> I guess we can remove it, that does limit the public API change just in case we overlooked something.

I think it helps to be conservative, particularly early on. If we find a lot of people want to do exactly this operation, then we can consider adding it as a public function.

cyrush:
@xjrc you have a good point --- having a corresponding map functions for fields, matsets, etc for all of the generate public API will be hard to manage. 

But for the cases where the generated topology changes number of elements or the number of vertices change -- we do need to have something that helps use map the associated data. 

Fields are the first thing you notice with poly stuff, but adjsets could come into play for other generate cases. 

Should field mapping be an option to generate sides?  That might make that method itself more complex. 


Regardless, we need this exact code in both VisIt and Ascent, so I really want to find the right balance to get it into blueprint.

xjrc:
> @xjrc you have a good point --- having a corresponding map functions for fields, matsets, etc for all of the generate public API will be hard to manage.
> 
> But for the cases where the generated topology changes number of elements or the number of vertices change -- we do need to have something that helps use map the associated data.
> 
> Fields are the first thing you notice with poly stuff, but adjsets could come into play for other generate cases.
> 
> Should field mapping be an option to generate sides? That might make that method itself more complex.
> 
> Regardless, we need this exact code in both VisIt and Ascent, so I really want to find the right balance to get it into blueprint.

@cyrush: I like the idea of embedding the option into `generate_sides` directly, but it will require some careful consideration in order to handle it sanely (not too many arguments) and meaningfully (arguments are clear enough to use). I'll post some ideas I have for this tomorrow morning.

cyrush:
@xjrc 
For public API: how about a new `generate_sides` signature that looks like the following:
```
   //---------------------------------------------------------------------
        void CONDUIT_BLUEPRINT_API generate_sides(const conduit::Node &n,
                                                  conduit::Node &topo_dest,
                                                  conduit::Node &coordset_dest,
                                                  conduit::Node &fields_dest,
                                                  conduit::Node &s2dmap,
                                                  conduit::Node &d2smap,
                                                  const std::string &field_name_prefix="");
```

cyrush:
example usage:
```
generate_sides(mesh["topologies/topo"],
                          sides["topologies/sides_topo"],
                          sides["coordsets/sides_points"],
                          sides["fields"],
                          s2dmap, d2smap,
                          "my_sides_");
```

cyrush:
other option we may want:

a list of fields to map (vs mapping all of them)

xjrc:
Would packing some of these options into a node make sense? I feel like we do this for other functions. For example:

```
   //---------------------------------------------------------------------
        void CONDUIT_BLUEPRINT_API generate_sides(const conduit::Node &n,
                                                  conduit::Node &topo_dest,
                                                  conduit::Node &coordset_dest,
                                                  conduit::Node &fields_dest,
                                                  conduit::Node &s2dmap,
                                                  conduit::Node &d2smap,
                                                  const conduit::Node &options);
```

Where `options` is something like:

```
Node options;
options["field_prefix"].set("sides_");
options["field_names"].append().set("a");
options["field_names"].append().set("b");
options.print();
// field_prefix: "sides_"
// field_names:
// - a
// - b
```

xjrc:
I realize this route has a few major drawbacks, i.e.:

1. How does a user know which options are supported?
2. Even if a user knows an option key, how will they figure out the corresponding value (e.g. "field_names" as a node list)?

These have to be balanced against the huge benefit of the interface not changing if new options are added. Given that the user will often have to check the documentation for information about these functions anyway, my current leaning is toward specification via the option `Node`.

cyrush:
I like using a node for the options. 

That makes it easier to wrap into other languages and gives us some flexibility.

When things start to get more complicated. A Node is always the answer :-)

cyrush:
I also think that given field names will need to be a list, that pushes us over the line to use a node.  Will make the interface much easier.

JustinPrivitera:
I still need to fix the rest of the tests; but this is most of the way there. However, I did change the function parameters slightly; to this:
```
void CONDUIT_BLUEPRINT_API generate_sides(const conduit::Node &topo_src,
                                          const conduit::Node &fields_src,
                                          conduit::Node &topo_dest,
                                          conduit::Node &coordset_dest,
                                          conduit::Node &fields_dest,
                                          conduit::Node &s2dmap,
                                          conduit::Node &d2smap,
                                          const conduit::Node &options);
```
The parameter `n` has been changed to `topo_src`, since it's where the topology comes from, and I added a new parameter `fields_src`, since we need to have access to the original fields so we can copy from them.

Is there a better or cleaner way of accomplishing this? Maybe just pass `n` (not the topology but the entire node) and then a string `topology` that can be used to get the chosen topology? What are your thoughts? @cyrush @xjrc

cyrush:
even in the case we were passing just topology -- we still needed to access the coordset (but it is not passed!)

How does this work?

The trick is, you can always assume the passed topology is part of a valid blueprint mesh. 
That means, if there are fields they will always exist in the same tree. You can look two levels up from the passed topology, and you will have the root of the entire mesh. From here you can get the coordset, or the fields.

JustinPrivitera:
> The trick is, you can always assume the passed topology is part of a valid blueprint mesh.
> That means, if there are fields they will always exist in the same tree. You can look two levels up from the passed topology, and you will have the root of the entire mesh. From here you can get the coordset, or the fields.

Is this code we're looking for to make that happen?
```
Node coordset;
bputils::find_reference_node(topo, "coordset", coordset);
```

Edit: the code we want is the `parent()` function.

JustinPrivitera:
I've made the requested changes; does this look good to merge? @cyrush @xjrc

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2960,6 +2960,260 @@ mesh::topology::unstructured::generate_sides(const Node &topo,
...
+        {
+            CONDUIT_ERROR("field_names must be a string or a list of strings.");
+        }
+    }
+
+    // check that the discovered field names exist in the target fields
+    for (uint64 i = 0; i < field_names.size(); i ++)
+    {
+        bool found = false;
+        NodeConstIterator itr = fields_src.children();
```
cyrush: you can simplify with `fields_src.has_child(field_names[i])`, then you don't need to write your own loop

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2960,6 +2960,260 @@ mesh::topology::unstructured::generate_sides(const Node &topo,
...
+            if (cld_name == field_names[i])
+            {
+                found = true;
+            }
+        }
+        if (!found)
+        {
+            CONDUIT_ERROR("field " + field_names[i] + " not found in target.");
+        }
+    }
```
cyrush, line 3186: +1 for good error messages

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 #### Blueprint
 - Added the `blueprint::mesh::examples::polychain` example. It is an example of a polyhedral mesh. See Mesh Blueprint Examples docs (https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#polychain) for more details.
 - Fixed a bug that was causing the `conduit::blueprint::mesh::topology::unstructured::generate_*` functions to produce bad results for polyhedral input topologies with heterogeneous elements (e.g. tets and hexs).
+- Added `blueprint::mesh::topology::unstructured::map_fields_to_generated_sides`, a companion function to `generate_sides` that takes fields from the topology before `generate_sides` and maps it onto the new topology.
```
cyrush: adjust this changelog entry to reflect new generate sides signature

src/libs/blueprint/conduit_blueprint_mesh.hpp
```diff
@@ -351,6 +351,15 @@ namespace topology
                                                   conduit::Node &s2dmap,
                                                   conduit::Node &d2smap);
 
+        //---------------------------------------------------------------------
```
cyrush, line 354: add a note here (the header) that this variant of the function call will also map the fields 
also add the details about what can be passed via the options node.

CHANGELOG.md
JustinPrivitera: right, thanks for the reminder

src/libs/blueprint/conduit_blueprint_mesh.hpp
JustinPrivitera: ok

JustinPrivitera:
I will add tests now for a couple other exceptions

cyrush:
thanks @JustinPrivitera !

PR #792, opened 2021-07-13
Fix check for threads target

white238:
This changes the if check to the existence of the target `Threads::Threads` not treating it like a constant, variable, or a string.

This unfortunately is not the source of the current problem I think, this was just causing `find_package(Threads)` to always be called.  I guess something could have this variable defined but that would be crazy and I just noticed this while poking around.

### DIFF
--- a/src/config/conduit_setup_deps.cmake
+++ b/src/config/conduit_setup_deps.cmake
@@ -10,7 +10,7 @@ include(CMakeFindDependencyMacro)
 if(UNIX AND NOT APPLE)
 # we depend on Threads::Threads in our exported targets
 # so we need to bootstrap that here
-    if(NOT Threads::Threads)
+    if(NOT TARGET Threads::Threads)
         find_package( Threads REQUIRED )
     endif()
 endif()


kennyweiss (approved):
Thanks @white238 !

cyrush (approved):

cyrush:
Thanks @white238 -- to be consistent, I think we need the target check here as well:

https://github.com/LLNL/conduit/blob/4b767a08213c7e8ec397c3222da7f2b1b05de871/src/config/conduit_setup_deps.cmake#L13

white238:
> 
> 
> Thanks @white238 -- to be consistent, I think we need the target check here as well:
> 
> https://github.com/LLNL/conduit/blob/4b767a08213c7e8ec397c3222da7f2b1b05de871/src/config/conduit_setup_deps.cmake#L13

... isn't that the same place?

There are two places you call `find_package(Threads ...)` and the SetupThirdParty one doesn't have the if.

https://github.com/LLNL/conduit/blob/ef19c50f000633f1a1566cd964eade37a6b6745f/src/cmake/Setup3rdParty.cmake#L18

cyrush:
Oh man, yes it is.  Sorry. 

Being parinoid: Also I checked the FindThreads module source back to CMake 3.14 & I verified that it does not add the `Threads::Threads` target  if it already exists. 

In theory, and find should be find to call ad nauseam -- but that is only in theory.

Thanks for the improved guard.

PR #793, opened 2021-07-13
polytess 3d and docs + changlog update

JustinPrivitera:
This branch adds the completed polytess 3d example, an update to the docs with a picture of the results, and an update to the changelog reflecting these additions.


src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -3585,7 +3585,273 @@ polychain(const index_t length, // how long the chain ought to be
...
 
+//-----------------------------------------------------------------------------
+void polytess_3d(conduit::index_t nlevels, Node &res)
+{
+    // Our goal here is to take the original polytess and extend it
+    // into 3 dimensions. The way we will accomplish this is by
+    // placing the original polytess into the z = 0 plane, placing a 
+    // copy of it into the z = 1 plane, and constructing "walls" between
+    // the top and bottom edges. Then we will specify polyhedra that use
+    // all the faces at our disposal.
```
cyrush: Great comment!

cyrush (approved):
Looks great to me, thanks for the detailed comments!

cyrush:
@xjrc, I think this is ready - lets us know if you have any suggestions

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -3585,7 +3585,273 @@ polychain(const index_t length, // how long the chain ought to be
     }
 }
 
+//-----------------------------------------------------------------------------
+void polytess_3d(conduit::index_t nlevels, Node &res)
```
xjrc: The interface split between `polytess` and `polytess_3d` seems to go against the established integrated 2D/3D interface convention of the other `examples` functions (e.g. `basic`, `braid`, etc.). Is there any way that `polytess_3d` integrated into `polytess` rather than being a separate function? I could envision using the `nz` variable seen in the other `examples` functions in a similar way here, e.g.:

```
// nz == 0/1 -> 2D
// nz == 2+ -> 3D (variable number of planes instead of just z=0, 1)
void polytess(index_t nlevels, index_t nz, Node &res);
```

xjrc (changes requested):
Looks good @JustinPrivitera, but I have a question about the interface that I think we should resolve prior to merging.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
JustinPrivitera: ok, so this would allow for users to create not just a single "sheet" of polytess, but stack sheets arbitrarily high?

xjrc: Yes, that's correct. To be honest, this is more of an aesthetic consideration rather than a practical one, but I do think having a greater level of interface uniformity is a good thing to aim for.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
JustinPrivitera: :)

src/libs/blueprint/python/conduit_blueprint_mesh_examples_python.cpp
```diff
@@ -630,6 +630,7 @@ PyBlueprint_mesh_examples_polytess(PyObject *, //self
     Node &node = *PyConduit_Node_Get_Node_Ptr(py_node);
 
     blueprint::mesh::examples::polytess(nlevels,
+                                        1,
```
JustinPrivitera, line 633: I'll need some guidance @cyrush  on how to get this working... it'd be nice if people using python weren't limited to 2 dimensions for their polytessalations.

JustinPrivitera:
This is implemented as discussed; please let me know if you have any questions @xjrc @cyrush and if there's anything that ought to be changed.
Right now the default behavior is that nz must be greater than 0; 1 places it in 2D, and 2 or more makes the 3D tower we discussed.

cyrush:
```
// python func
static PyObject * 
PyBlueprint_mesh_examples_polytess(PyObject *, //self
                                   PyObject *args,
                                   PyObject *kwargs)
{
    Py_ssize_t nlevels = 0;
    Py_ssize_t nz        = 0;
    PyObject   *py_node  = NULL;

    static const char *kwlist[] = {"nlevels", 
                                                    "nz",
                                   "dest",
                                   NULL};

    if (!PyArg_ParseTupleAndKeywords(args,
                                     kwargs,
                                     "nnO",
                                     const_cast<char**>(kwlist),
                                     &nlevels,
                                     &nz,
                                     &py_node))
    {
        return (NULL);
    }

    if(!PyConduit_Node_Check(py_node))
    {
        PyErr_SetString(PyExc_TypeError,
                        "'dest' argument must be a "
                        "conduit.Node instance");
        return NULL;
    }

    Node &node = *PyConduit_Node_Get_Node_Ptr(py_node);

    blueprint::mesh::examples::polytess(nlevels,
                                        nz,
                                        node);

    Py_RETURN_NONE;
}

```

should work for python, it does change the api so you may have to adjust existing test (if its there)

Ah we also need a try catch.

Let's merge and I can follow up with the Python updates.

xjrc (approved):
Thanks for adding `nz` and all of the new renders @JustinPrivitera; looks great!

cyrush (approved):

cyrush:
thanks @JustinPrivitera !

PR #795, opened 2021-07-19
conduit/c: use a type safe alias for `conduit_{datatype,node}`

### LINKED ISSUE #782: In the C API, `conduit_node` is `void` which allows for surprising conversions
Currently, the C API does:

```c
typedef void conduit_node;
```

which means the APIs are all really:

```c
void conduit_node_some_api(const void* node);
```

to the compiler. This means that:

```c++
std::vector<int> vec;
conduit_node_some_api(&vec);
```

typechecks and compiles just fine (and probably crashes or causes some other kinds of fireworks at runtime). Instead, it should be something like:

```c
struct _conduit_node;
typedef struct _conduit_node conduit_node;
```

so that there is a distinct type associated with the call and the compiler can at least complain that types don't match (or that the user has to explicitly do a C cast or `reinterpret_cast` in which case the fault is squarely in their code).

Cc: @cyrush @utkarshayachit

mathstuf:
Using `typedef void` meant that the APIs were all taking `void*` which
in turn means that any pointer is valid to pass to these types. Instead,
use a unique opaque type so that callers must either do the unsafe
casting themselves or use the `cpp_to_c` APIs to do the casting the
intended way.

The actual definition of the structs themselves are empty; they are just
used it as a way to smuggle the C++ pointer around under an opaque name.

Fixes: #782

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 
 #### General
 - Avoid compile issue with using `_Pragma()` with Python 3.8 on Windows
+- `conduit_node` and `conduit_datatype` in the C API are no longer aliases to `void` so that callers cannot pass just any pointer to the APIs.
 
 #### Blueprint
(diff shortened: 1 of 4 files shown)


mathstuf:
Cc: @cyrush @utkarshayachit

cyrush:
thanks @mathstuf!   Everything looks good -- can you add an entry to the CHANGELOG.md?

cyrush:
thanks @mathstuf !

PR #796, opened 2021-07-20
Simplify cmake api

mathstuf:
---
What's the preferred way to patch `civetweb`?

Cc: @utkarshayachit @cyrush

### DIFF
--- a/src/cmake/CMakeBasics.cmake
+++ b/src/cmake/CMakeBasics.cmake
@@ -245,56 +245,6 @@ endmacro()
 
 
 ###############################################################################
-# Provide macros to simplify adding compile and link flags to a target
-###############################################################################
-
-macro(add_target_compile_flags)
-    set(options)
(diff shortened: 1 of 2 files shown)


cyrush:
@mathstuf, we don't have a very robust procedure for patches.

When making changes, I add a comment with Conduit in the name so we can hunt them down, here is an example:

https://github.com/LLNL/conduit/blob/c161c5df3e30327b7e0ffe678fd1c0658de07f54/src/thirdparty_builtin/civetweb-0a95342/src/civetweb.c#L141

If you also add a patch file here:

https://github.com/LLNL/conduit/tree/develop/src/thirdparty_builtin/civetweb-0a95342

I think that would help improve things going forward.

cyrush:
In this case the `CMakelists.txt` file is ours, I thought you may have had another source level patch at play, sorry if my advice was confusing.

cyrush:
thanks!

PR #797, opened 2021-07-20
bugfix for fortran logical overread

### LINKED ISSUE #778: Memory overrun in fortran interface returning bool
Aaron Skinner and I ran into a stack memory overflow on the Fortran interface:
```
conduit_node_has_path(node, path)
```

The C function implementation is returning a 4 byte integer, but the Fortran interface has the return type declared as a
```
logical(c_bool)
````
which is 1 byte.

So, there is a 3 byte overrun.

cyrush:
resolves #778


cyrush:
FYI: @aaroncblack  I think this will resolve the issue you saw, can you give the changes a quick look?

aaroncblack:
That looks good, thanks.

cyrush:
@aaroncblack  thanks!

PR #806, opened 2021-07-27
improve mesh bp func sigs

### LINKED ISSUE #794: better func sig for generate_functions
From the header file:

https://github.com/LLNL/conduit/blob/d7f9ac7edef861f8b57a9e4abd579c50e5d9715e/src/libs/blueprint/conduit_blueprint_mesh.hpp#L319

It's hard to tell that `Node &n` is implied to be a valid blueprint topology. 

We should change these to `topo` in the header. (Note these are topo in the cpp file)

cyrush:
resolves #794

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -2275,7 +2275,7 @@ mesh::topology::length(const Node &topology)
 
 //-----------------------------------------------------------------------------
 bool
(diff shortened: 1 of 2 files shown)


xjrc (approved):
This is a good start, but we'll have to update the source file as well because it inherited the ambiguous input parameter name `n`.

cyrush:
Thanks, I'll take a pass at the source file as well.

cyrush:
i updated the cpp file as well

PR #807, opened 2021-07-31
generate sides: support for volume dependent fields

JustinPrivitera:


src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2963,43 +2963,269 @@ mesh::topology::unstructured::generate_sides(const Node &topo,
...
+    {
+        return 0.5f * (x1 * (y2 - y3) + x2 * (y3 - y1) + x3 * (y1 - y2));
+    }
+
+    float tetrahedron_volume(float x1, float y1, float z1,
+                             float x2, float y2, float z2,
+                             float x3, float y3, float z3,
+                             float x4, float y4, float z4)
+    {
+        CONDUIT_ERROR("not yet implemented");
```
JustinPrivitera: This remains to be done

JustinPrivitera:
What's here so far is the functionality for volume dependent fields. On Monday I'll add the `tetrahedron_volume()` function and comprehensive tests.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2963,43 +2963,269 @@ mesh::topology::unstructured::generate_sides(const Node &topo,
...
+            CONDUIT_ERROR("Higher dimensions are not supported.");
+        }
+
+        // next we calculate the volume of each polygon
+        (*volumes)["poly"].set(conduit::DataType::float64(num_orig_shapes));
+        float64 *poly_volumes = (*volumes)["poly"].value();
+
+        for (int i = 0; i < num_orig_shapes; i ++)
+        {
+            poly_volumes[i] = 0;
```
JustinPrivitera, line 3080: Can I assume with `set()` that everything will be set to 0, or should I do this for loop regardless?

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -3065,40 +3327,134 @@ namespace detail
...
+                    }
+
+                    volume_ratio = volumes["ratio"].value();
+
+                    // make volume into a field
+                    Node &volumes_field = fields_dest["volume"];
+                    volumes_field["topology"] = topo_name;
+                    volumes_field["association"] = "element";
+                    volumes_field["volume_dependent"] = "true";
+                    volumes["tri"].to_float64_array(volumes_field["values"]);
```
JustinPrivitera: This is what Cyrus wanted changed

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -3065,40 +3318,130 @@ namespace detail
...
+                    }
+                    else if (topo_dest["elements/connectivity"].dtype().is_int32())
+                    {
+                        volume_dependent<T, int32>(topo_dest,
+                                                   coordset_dest,
+                                                   dimensions,
+                                                   num_new_shapes,
+                                                   num_orig_shapes,
+                                                   tri_to_poly,
+                                                   volumes);
```
cyrush: I think you can pass `volumes_field["values"]` here

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2963,43 +2963,284 @@ mesh::topology::unstructured::generate_sides(const Node &topo,
...
+    volume_dependent_helper(const Node &topo_dest,
+                            const Node &coordset_dest,
+                            const int dimensions,
+                            const int num_new_shapes, // number of new triangles or tetrahedrons
+                            const int num_orig_shapes, // number of original polygons or polyhedra
+                            const T *tri_to_poly,
+                            Node &volumes) 
+    {
+        // first we calculate the volume of each triangle
+        volumes["tri"].set(conduit::DataType::float64(num_new_shapes));
```
cyrush: skip the `["tri"]` here

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -3008,6 +3259,13 @@ namespace detail
         
         const T *tri_to_poly = d2smap["values"].value();
 
+        Node &original_elements = fields_dest["original_element_ids"];
```
cyrush: we need to also add the prefix name here

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -3065,40 +3337,209 @@ namespace detail
                     }
                 }
 
+                // handle volume dependent fields
+                if (vol_dep)
+                {
+                    // make volume into a field
+                    Node &volumes_field = fields_dest["volume"];
```
cyrush: also add the prefix name here.

We only need to calc the volume once. Since this is in the loop over fields, as is it could happen multiple times unnecessarily.

To solve this, you can check if the field was already created on subsequent trips around the loop.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -3065,40 +3338,220 @@ namespace detail
                     }
                 }
 
+                // handle volume dependent fields
+                // if the field is volume dependent and we have not already calculated the volumes
+                if (vol_dep && !volumes_info.has_child("poly"))
```
cyrush: one more suggestion (sorry !) 

you don't need to pass in volumes_info, you can use a bool to check if you already created it during the iteration + if it doesn't exist, you can create it. If it does you can pull out the existing volume field ref. 



If it is already there, you can pull the volumes out

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2963,43 +2963,296 @@ mesh::topology::unstructured::generate_sides(const Node &topo,
...
+
+    float64 tetrahedron_volume(vec3 &a, vec3 &b, vec3 &c, vec3 &d)
+    {
+        return fabs((a - d).dot((b - d).cross(c - d))) / 6.0f;
+    }
+
+    // T is the type of 'tri_to_poly' values
+    // U is the type of connectivity values
+    // V is the type of coordset values
+    template<typename T, typename U, typename V>
```
xjrc, line 3013: I think that templates are a bit overkill here, and cause a lot of the caller code to be more complicated than it needs to be (e.g. via large case statements). In order to solve the general problem of type uncertainty in Conduit, I would recommend using Conduit's type coercion/transformation capabilities. You can find a good demonstration of this in [`blueprint::mesh::utils::topology::unstructured::generate_offsets`](https://github.com/LLNL/conduit/blob/develop/src/libs/blueprint/conduit_blueprint_mesh_utils.cpp#L1063), which uses a temporary dynamic node at a specific index and `to_int64` to coerce input data into a type that can be readily used and understood.

I realize this is quite high-level feedback; does this make sense? If not, I can create a PR off of your set of changes here to demonstrate what I mean more concretely.

cyrush: @xjrc  In this case we dispatch this way avoid extra intermediate copies.  We do this often in other places like Ascent (even if we don't use templates to accomplish it, we don't always coerce b/c we want to avoid touching memory. 

That said, for volume dep, the result kind is coerced to be a `float64`, b/c we are scaling it by a volume that is `float64`

For other mappings we do want to preserve the type. Templates help this (and aren't exposed in the interface)

JustinPrivitera: Looking at what you've written and at `generate_offsets`, I think this makes sense. I'll play with it and let you or Cyrus know if I run into any trouble.

xjrc: @cyrush: I focused a bit too much on coercion in my prior comment, but the general point I wanted to make was that Conduit's internal type handling mechanisms can perform type transformation in a cleaner way than templates and their corresponding required large case statements. A more applicable example to the type preservation case can be found [here](https://github.com/LLNL/conduit/blob/develop/src/libs/blueprint/conduit_blueprint_mesh_utils.cpp#L565); basically, `conduit::Node::to_data_type` allows for type determination/transformation with minimal intervention, and using it one data element at a time has a O(1) memory cost.

cyrush: @xjrc, I think I see what you are aiming for.  In this case -- we are doing a lot of indirection, we need to pull out either 3 or 4 indices and then apply those indices to pull out the coord values we need for the calc. The current implementation does zero copy (using all the original data). It would be substantially slower to use a node to marshal these to known types.  With the suggested pattern, I think it might also be more quite a bot more code as well? Maybe I missing something.

JustinPrivitera: @xjrc Looking at the example you provided, I don't quite understand what changes I should be making. In order to do type determination/transformation on one data element at a time, I need a way to index into the array where those data elements live, in which case it seems I'd need to know the type of the array, which means doing the excessive templates we're trying to avoid. Is there a way to index into a conduit node array without needing to know the type? There must be something I'm missing here.

xjrc: @cyrush: I do agree that the templated version is probably faster, but I don't think I'd go so far as to say substantially so (i.e. I believe the asymptotic behavior is the same). On the code amount point, the exact line differential is hard to call, but the three big conditional type cascades in the existing implementation lead me to believe that the template-free solution would be smaller in size, even if only slightly.

Fundamentally, I think the choice here boils down to a trade-off between code efficiency and readability, with templates coming ahead in the former category and dynamic typing in the latter. To drive home the readability point, consider any Conduit/Blueprint function `f` that operates on one or more arrays `n1, ... , nN`. To leverage the efficiency advantage of templates, we'll need to create a set of internal template helpers `f^i` where `i in [1, N-1]` where each helper contains a conditional type cascade to fix a subsequent type in the type list for the next template call, with `f^N` being the actual implementation.

We can observe the above pattern in @JustinPrivitera's implementation of `volume_dependent_helper`, which needs 3 template dispatch layers to handle its two index arrays and one geometric data array:

- `f^3 = volume_dependent_helper`: Three template arguments - `T` (map values), `U` (topology values), and `V` (coordinate values)
- `f^2 = volume_dependent`: Type cascade fixes `V` (coordinate values)
- `f^1 = map_fields_to_generated_sides`: Type cascade fixes `U` (topology values)
- `f = generate_sides`: Type cascade fixes `T` (map values)

If we start using this approach in other places, I think we'll find our type cascades quickly multiplying. I realize that we have a lot of them already, but I think we'll find that deciphering type errors will become increasingly difficult as we add more layers.

Let me know what you think; I realize that this is fairly abstract and I'm happy to hash this out further in a more analog format.

@JustinPrivitera: I think [this example](https://github.com/LLNL/conduit/blob/develop/src/libs/blueprint/conduit_blueprint_mesh_utils.cpp#L565) demonstrates the one-by-one indexing pattern much better. The general idea is to use a Node to proxy an individual element of the dynamic type and to use `conduit::Node::element_ptr` to access values at particular indices.

cyrush: thanks @xjrc  I think it make sense to ponder this a bit more. I do want to look a bit at the performance to get a handle on the differences. By doing set externals we aren't allocating and freeing the data, but there are still a lot of moving pieces (and there will be allocations associated with the data type changes, etc). The per element overhead is going to be high.

Also, it might be possible for us to make the node dynamic conversion pattern a little easier to wield.  I recognize it when I see it, but I am not sure it is obvious to other folks what is happening.  Templates are locally more readable, but the entire scheme is harder to understand -- the node conversion method is overall easier to grok with local steps being more complex.

cyrush: @xjrc checkout: https://github.com/LLNL/conduit/pull/819

Note -- First, I want to make sure I haven't made a typo that is skewing the numbers, but the element by element conversion appears to be very expensive compare to template dispatch: ( ~50 to 100x slower )

xjrc (approved):
I think we'll want to standardize our use of templates for array processing dispatch and improve our linear geometry capabilities (perhaps by using the template-enabled `glm`?), but this is good to come in as is. Thanks @cyrush for investigating the efficiency benefits of templates and @JustinPrivitera for your patience.

cyrush:
thanks @xjrc

cyrush:
thanks @JustinPrivitera !

PR #810, opened 2021-08-06
add I/O support for Silo geometry/fields

xjrc:
The changes in this pull request update Relay's Silo capabilities to enable basic Blueprint mesh I/O (i.e. geometry and fields). The full enumerated set of changes/features can be found below:

- [ ] Read Support (Silo to Blueprint)
  - [ ] Geometry
    - [ ] Implementation
    - [ ] Testing
  - [ ] Fields
    - [ ] Implementation
    - [ ] Testing
- [ ] Write Support (Blueprint to Silo)
  - [ ] Geometry
    - [ ] Implementation
    - [ ] Testing
  - [ ] Fields
    - [ ] Implementation
    - [ ] Testing


cyrush:
a few notes:

1. we should keep the old api (silo_write_mesh) in addition to the new api (silo::read_mesh) for at least one release and note that it will be deprecated. 

2. One of the confusing choices we made a long time ago were the semantics for load and save vs read and write. 
That said we still honor them. 

`read` and `write` allow us to read and write into an existing tree  or files without things being cleared. 

`load` and `save` imply that when you read you only get those results and when you write you overwrite.

There are good cases for both. 

Takeaway here: you want to implement `read` and `write`, and then `load` and `save` are just stricter variants that use the more general read and write functions.  For example, `load` can just reset the output node and then call `read`.

`save` can open the file in overwrite mode (vs append), and then call `write`

cyrush:
closing in favor of #844

Issue #813, opened 2021-08-10
HDF5 Offset bug


kevinhkhuynh:
Currently, offset and stride are integers in: https://github.com/LLNL/conduit/blob/develop/src/libs/relay/conduit_relay_io_hdf5.cpp

This is a problem for very large datasets whose number of elements are greater than INT_MAX. This can be fixed by switching from integers to hsize_t, which are unsigned long long.

cyrush:
resolved in #814

PR #814, opened 2021-08-10
switch to hsize_t for hdf5 offset and stride

kevinhkhuynh:
This is a small bug fix for issue #813.


cyrush:
thanks @kevinhkhuynh!  It looks like there is a python related issue, I'll smooth that out.

kevinhkhuynh:
> 
> 
> thanks @kevinhkhuynh! It looks like there is a python related issue, I'll smooth that out.

Thanks, Cyrus!

cyrush:
python issue was just that it was expecting an exception for neg numbers, we no longer need that check.

cyrush (approved):

cyrush:
thanks @kevinhkhuynh !

PR #815, opened 2021-08-10
cmake: resolve an issue finding main python lib for more corner cases

cyrush:


cyrush:
should resolve:

https://github.com/visit-dav/visit/issues/5704

(needs to be ported into VisIt's cmake to do so)

PR #819, opened 2021-08-12
add a unit test that demos and times a few diff dispatch options

cyrush:
does a simple float64 sum of input array using 3 different methods 

You can override the array size and the number of iterations  timed:

`./tests/conduit/t_conduit_node_type_dispatch {array_size} {num_iters}`

Example run (debug build on macOS with swelling battery)

`./tests/conduit/t_conduit_node_type_dispatch 1000000 10`

```
Number of elements: 1000000
Iterations: 10
on_the_fly_convert: 2.02642
array_convert:      0.218024
template_dispatch:  0.0253357
```

### DIFF
--- a/src/tests/conduit/CMakeLists.txt
+++ b/src/tests/conduit/CMakeLists.txt
@@ -36,6 +36,7 @@ set(BASIC_TESTS t_conduit_smoke
                 t_conduit_node_iterator
                 t_conduit_node_cpp_iterator
                 t_conduit_node_obj_names_with_slashes
+                t_conduit_node_type_dispatch
                 t_conduit_schema
                 t_conduit_error
                 t_conduit_log
--- a/src/tests/conduit/t_conduit_node_type_dispatch.cpp
+++ b/src/tests/conduit/t_conduit_node_type_dispatch.cpp
@@ -138,31 +138,41 @@ TEST(conduit_node, test_dispatch)
     std::cout << "Number of elements: " <<  ARRAY_SIZE << std::endl;
     std::cout << "Iterations: " << NUM_ITERS << std::endl;
 
+    float64 res1,res2,res3;
+
     Timer t1;
     for(index_t i=0;i<NUM_ITERS;i++)
     {
-        on_the_fly_convert(n);
+        res1 = on_the_fly_convert(n);
     }
     t1.elapsed();
 
     Timer t2;
     for(index_t i=0;i<NUM_ITERS;i++)
     {
-        array_convert(n);
+        res2 = array_convert(n);
     }
     t2.elapsed();
 
     Timer t3;
     for(index_t i=0;i<NUM_ITERS;i++)
     {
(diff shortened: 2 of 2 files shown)


xjrc:
Thanks for putting together a test case to get some concrete numbers. I'll run the same test using the release version on `rzgenie` and see what we get there.

xjrc:
Here are some approximately average results after multiple iterations for each node:

#### Login Node ####

```
Number of elements: 1000000
Iterations: 10
on_the_fly_convert: 0.745429
array_convert:      0.130413
template_dispatch:  0.01168
```

#### Batch Node ####

```
Number of elements: 1000000
Iterations: 10
on_the_fly_convert: 0.643126
array_convert:      0.091531
template_dispatch:  0.00946761
```

With that in mind, your original approximation of ~50-100x slower seems accurate even in release conditions.

xjrc:
For what it's worth, I attempted to optimize `to_float64` to improve the performance of the "on-the-fly" implementation (by using dispatch based on a `DataType`-indexed function table), but it didn't make much of a difference.

Supposing that these are close to the optimal implementations for each "bulk conversion" solution, it's clear that templates are the fastest and least memory intensive solution. That being said, I think there are two important considerations going forward:

1. Considering that a move to template-based solutions indicates a clear emphasis on the efficiency of our code, do we think this solution will help us in our efforts to enable future optimizations on the GPU? Put more particularly, do we think this solution will be "performance-portable"?
2. While the "on-the-fly" solution isn't exactly intuitive, I think the template-based solution will be even harder for future Conduit developers and developers of adapter codes (e.g. the code @terryhaut is working on) to pick up. Do we need to start writing developer documentation that covers concepts like these?

cyrush:
1) It helps. Templates alone aren't enough, but having bulk work described in this way would make it easier to adapt that work directly to the GPU if needed.

2) I think there are two considerations: what we do internally to implement capabilities and the suggestions that we share for other folks to follow.  I do think we should build up some guidance and examples.

cyrush:
The issue with the on_the_fly is the all the function call and init overhead happening at array entry. All of that is stripped away with a template dispatch.

xjrc:
> The issue with the on_the_fly is the all the function call and init overhead happening at array entry. All of that is stripped away with a template dispatch.

I've long been skeptical of "zero-cost abstractions" and thus am not surprised that 1 template call over the entire array is much faster than 1 `to_*` call per array element. The piece that's surprising to me is the magnitude; I might be compelled to argue the readability angle if we were only sacrificing 2-10x slowdowns, but I consider 60-80x to be beyond the pale for code like this. If it weren't for the memory cost, I'd probably argue for choosing the `array_convert` approach, but alas.

xjrc (approved):

cyrush:
Oh don't worry, there is a cost to `zero-cost`, wielding it extensively will consume your soul.

PR #823, opened 2021-08-16
use parmetis to generate a field that shows a suggested partition

cyrush:


src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1482,6 +1481,33 @@ mesh::domains(const conduit::Node &mesh)
     return std::vector<const conduit::Node *>(std::move(doms));
 }
 
+//-------------------------------------------------------------------------
+void
+mesh::non_empty_domains(conduit::Node &mesh,
```
cyrush: - [x] revisit this

src/libs/conduit/conduit_error.cpp
```diff
@@ -59,6 +59,7 @@ Error::Error(const std::string &msg,
  m_line(line),
  m_what("")
 {
+    std::cout << " SOMETHING IS WRONG! " << message() << std::endl;
```
cyrush: - [x] revert this

src/libs/conduit/conduit_utils.cpp
```diff
@@ -171,6 +171,7 @@ default_error_handler(const std::string &msg,
                       const std::string &file,
                       int line)
 {
+    std::cout << msg  << " " << file << " " << line << std::endl;
```
cyrush: - [x] revert this

src/libs/blueprint/conduit_blueprint_mpi_mesh_parmetis.cpp
```diff
@@ -231,9 +231,11 @@ void generate_global_element_and_vertex_ids(conduit::Node &mesh,
 
 //-----------------------------------------------------------------------------
 void generate_partition_field(conduit::Node &mesh,
+                              int npartitions,
                               MPI_Comm comm)
```
cyrush, line 234: To avoid confusion -- I think we should probably only support the signature that includes options (only provide one interface).

The reason why we prefer a node tree to handle options is that it makes multi-lang easier (the options logic is only implemented in C++)

src/tests/blueprint/t_blueprint_mpi_mesh_parmetis.cpp
```diff
@@ -0,0 +1,287 @@
...
+    MPI_Comm_size(MPI_COMM_WORLD, &par_size);
+    MPI_Comm_rank(MPI_COMM_WORLD, &par_rank);
+
+    const int npts = 10;
+
+    // test with a 2d poly example
+    Node mesh, side_mesh, info;
+
+    // create braid
+    conduit::blueprint::mesh::examples::braid("structured",
```
cyrush, line 122: This is really good for bootstrapping new tests. 

To be clear, long term -- our eventual goal here is to pass in the other mesh types  w/o doing a topology transform (uniform, structured, etc -- will be directly supported)

src/tests/blueprint/t_blueprint_mpi_mesh_parmetis.cpp
```diff
@@ -0,0 +1,287 @@
...
+        Node &side_topo = side_mesh_dom["topologies/topo"];
+        Node &side_fields = side_mesh_dom["fields"];
+
+        // we can't map vert assoced fields yet
+        Node opts;
+        opts["field_names"].append().set("global_element_ids");
+        opts["field_names"].append().set("parmetis_result");
+
+        // gen sides and save so we can look at this in visit.
+        blueprint::mesh::topology::unstructured::generate_sides(in_domain["topologies/topo"],
```
cyrush, line 264: For your info:

We are working to get `generate_sides` automatically used by the VisIt plugin for the poly case. 

This will make looking at these meshes much easier in the future!

cyrush (commented):
These look good! A few comments about broader context.

PR #824, opened 2021-08-17
demo data accessor approach

cyrush:
demos DataAccessor.

They are used much like data arrays (`float64_accessor` instead of `float64_array`  ), but they only provide very simple read only access to values. 

They allow us to access a Node's array elements as a particular type (regardless of the underlying type, using on the fly casting after elements are fetched)

Here is an example of how to use this:
```
float64_accessor vals(n);
float64 res = 0;
index_t nele = vals.number_of_elements();

for(index_t i=0; i < nele;i++)
{
    res += vals[i];
}
```

updated times from type dispatch test:

```
Number of elements: 100000000
Iterations: 1
on_the_fly_convert: 19.8219
array_convert:      2.33158
pointer_dispatch:   0.296983
array_dispatch:     1.02438
data_accessor:      1.42676
```

`pointer_dispatch` here is using raw pointer math -- which will always be the fastest, however it does not handle cases with non-standard strides. 

Both `array_dispatch` and `data_accessor` do handle custom strides + they have similar performance. 

I think `data_accessor` can  greatly simplify writing some algos - we should discuss.


cyrush:
@xjrc, lets discuss this soon.

cyrush:
since there is not an existing precedent for array wrapping a node, this might be strange:

```
float64_accessor vals(n);
```

We may only want to support:

```
float64_accessor vals = n.value();
float64 res = 0;
index_t nele = vals.number_of_elements();

for(index_t i=0; i < nele;i++)
{
    res += vals[i];
}
```

And we may want `as_zzz` helpers, not sure what name to use --  `Node::as_float64_accessor`  style?

In this case there won't be a distinction between `as_` and `to_`, so `Node::float64_accessor` style might be good.

cyrush:
For Node class interface we will go with `as_zzz_accessor`

src/tests/conduit/t_conduit_data_accessor.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include "conduit.hpp"
+
+#include <iostream>
+#include "gtest/gtest.h"
+
+using namespace conduit;
+
+//-----------------------------------------------------------------------------
+TEST(conduit_data_accessor, value)
+{
```
cyrush, line 20: this file tests all the cases, and shows how the accessors can be used

cyrush:
@agcapps @kanye-quest

I think the `DataAccessor` foundation is ready. 

As a gut check for me, can you take a look (perhaps focused on the tests) and see if you have any questions or concerns?

Long term goal is to think about magically swapping out template optimized DataArrays or DataAccessors for cases with trivial or native strides to gain back some performance compared to raw pointer cases. 

Basically, on object creation -- we can make a runtime choice to map to the optimized template imp.

All of that said, it really depends on how this impacts the interface.  What we have here is nice, don't want to stray to far from it to chase optimization.

agcapps:
Thanks, @cyrush . This looks like good style, unobtrusive and not awkward.

publixsubfan:
Some thoughts on the API:

- We might want to differentiate between `DataAccessor<T>` and `DataAccessor<const T>` in order to avoid the const_cast from a `const void*` pointer, akin to the `std::span<T>` vs `std::span<const T>` distinction.
Probably isn't that important right now since `DataAccessor` is read-only ATM, but:
- Like `std::span<T>`, maybe we should also support writes to the underlying array. That would help clean up some more boilerplate in the repartitioning code. Although type conversions might be more fraught in that direction; it would probably be good practice to guard or warn when constructing writable accessors that implicitly narrow.

cyrush:
got it thanks! For now, we are read only I think we are in good shape.

I am interested in the implications of trying to think about supporting writing, let's add a ticket for that?

agcapps (approved):
Thank you, Cyrus.  Good to see the "dogfooding" in here too.

PR #827, opened 2021-08-20
Support for vertex associated fields in generate sides

JustinPrivitera:
Generate sides should now support vertex associated fields in the following way: for every new point, it is assigned a field value equal to the averages of the points it is connected to in the new topology.


JustinPrivitera:
Thorough tests have yet to be written.

cyrush (approved):
Looks good, thanks!

Issue #830, opened 2021-08-24
cmake typo


cyrush:
reported here:

https://github.com/LLNL/conduit/blob/b8b6884d6a8ae4c243b938c50418b095b15ee0ab/src/cmake/thirdparty
/SetupPython.cmake#L28

cyrush:
resolved in #831

PR #831, opened 2021-08-25
a few cmake fixes

cyrush:
prefer python3 when `PYTHON_DIR` is given, and `PYTHON_EXECUTABLE` is not
typo in cmake message statement
fix issue in cmake hdf5 setup logic


cyrush:
resolves #830

PR #844, opened 2021-09-30
add I/O support for Silo geometry/fields

jameshcorbett:
This is a continuation of #810, since for some reason I couldn't push my commits to Joe's branch.

There shouldn't be any changes to the old silo code. However, my editor ate most of the trailing white space in that file, making the diff a little more complicated. 

The new write functionality (in particular, lines 2130-2862) is a slightly refactored copy of the existing functionality, just with a wrapper to handle multiple domains, some of the options we'll be supporting, and to write the DBmultimesh/DBmultimat/DBmultivar objects. That is just to say that I may not be able to answer questions about that code very well. But please ask!

- The code for writing matsets and fields should probably be replaced by the `to_silo` blueprint functions.
- One of the tests fails at the moment. It would be good to have more multidomain 'write' tests
- I noted some TODO items in the code. 
- It might be a good idea to go through the silo documentation for some of the major structs used in the code and check what kinds of fields are being ignored in the code. For instance, some Silo structs have 'row_major' fields that are ignored in the code at the moment.
- The existing test files should probably be replaced with files that don't require symlinks. The best way to do that may be to write a short C program. I think the symlinks can be removed while still staying withing the Overlink spec.
- There is no support for reading implicit coordinates, but I'm not even sure silo supports those

In general I think the path forward is to add more examples/tests and see where things break. That is what I have been doing. 

Please post comments and questions!


src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -1380,19 +1392,1777 @@ void silo_mesh_write(const Node &node,
...
+        return "polyhedral";
+    if (shapetype == DB_ZONETYPE_POLYGON)
+        return "polygonal";
+    CONDUIT_ERROR("Unsupported zone type " << shapetype);
+    return "";
+}
+
+// copy data and assign it to a Node
+template <typename T>
+void copy_and_assign(T *data, int data_length, conduit::Node &target) {
```
cyrush, line 1493: - [ ] Cyrus TODO -- look into this more. Plain old sets might work.

cyrush:
thanks @jameshcorbett ! This is a nice chunk of work. I will it up a bit later and work on the final changes to merge.

cyrush:
FYI: merging this set of work into a branch in conduit repo, where we will continue work

PR #850, opened 2021-10-26
add more silo blueprint i/o support

cyrush:
bulk of the work done by:
 jameshcorbett


cyrush:
TODO:

- [x] roundtrip for spiral fails (mesh bp verify fails):

Topo created is incorrect.

Verify result:

```
0: 
  coordsets: 
    valid: "true"
    topo: 
      type: 
        valid: "true"
      info: 
        - "mesh::coordset::explicit: 'type' has valid value 'explicit'"
        - "mesh::coordset::explicit: 'values' is an mcarray"
      values: 
        valid: "true"
      valid: "true"
  topologies: 
    valid: "false"
    topo: 
      type: 
        valid: "false"
        errors: 
          - "mesh::topology::type: has invalid value 'quad'"
      valid: "false"
      coordset: 
        valid: "true"
  valid: "false"
1: 
  coordsets: 
    valid: "true"
    topo: 
      type: 
        valid: "true"
      info: 
        - "mesh::coordset::explicit: 'type' has valid value 'explicit'"
        - "mesh::coordset::explicit: 'values' is an mcarray"
      values: 
        valid: "true"
      valid: "true"
  topologies: 
    valid: "false"
    topo: 
      type: 
        valid: "false"
        errors: 
          - "mesh::topology::type: has invalid value 'quad'"
      valid: "false"
      coordset: 
        valid: "true"
  valid: "false"
info: 
  - "mesh: is a multi domain mesh"
valid: "false"
```

more details from domain 0:

```
comparing domain 0
{input}

coordsets: 
  topo: 
    type: "rectilinear"
    values: 
      x: [0.0, 1.0]
      y: [0.0, 1.0]
topologies: 
  topo: 
    type: "rectilinear"
    coordset: "topo"

{loaded}

topologies: 
  topo: 
    type: "quad"
    coordset: "topo"
coordsets: 
  topo: 
    type: "explicit"
    values: 
      x: [1.0, 2.0]
      y: [0.0, 1.0]

/Users/harrison37/Work/github/llnl/conduit/src/tests/relay/t_relay_io_silo.cpp:215: Failure
Value of: l_curr.diff(s_curr, info)
  Actual: true
Expected: false
{diff}

children: 
  diff: 
    topologies: 
      children: 
        diff: 
          topo: 
            children: 
              diff: 
                type: 
                  errors: 
                    - "data_array::diff: data string mismatch (\"quad\" vs \"rectilinear\")"
                  valid: "false"
                coordset: 
                  valid: "true"
            valid: "false"
      valid: "false"
    coordsets: 
      children: 
        diff: 
          topo: 
            children: 
              diff: 
                type: 
                  errors: 
                    - "data_array::diff: data string mismatch (\"explicit\" vs \"rectilinear\")"
                  valid: "false"
                values: 
                  children: 
                    diff: 
                      x: 
                        value: [1.0, 1.0]
                        errors: 
                          - "data_array::diff: data item(s) mismatch; see 'value' section"
                        valid: "false"
                      y: 
                        value: [0.0, 0.0]
                        valid: "true"
                  valid: "false"
            valid: "false"
      valid: "false"
valid: "false"
```

JustinPrivitera:
- [x] copy read mesh method from blueprint to silo
 - [x] fix all todos
 - ~~[ ] overlink~~
 - [x] audit error throwing in write method, follow approach taken in read method using wrappers to clean up
 - [x] audit all helper methods to make them use wrappers
 - [x] lots of tests
 - [x] fill out optlists
 - [x] read multivar support
 - [x] "state"
 - [x] array vars
 - [x] make CI pass
 - [x] MPI tests

cyrush:
Screen Shot 2023-05-08 at 8 46 48 AM

JustinPrivitera:
@cyrush 
This is complete. It may be rough around the edges; I unfortunately do not have time right now to go back through every change and make sure each one looks good. However, I have been through the code many times over the course of this project.

Please let me know (when you have time, of course) what needs fixing and changing and we can iterate from there.

src/tests/blueprint/t_blueprint_mesh_examples.cpp
```diff
@@ -232,11 +235,10 @@ TEST(conduit_blueprint_mesh_examples, mesh_2d)
                 continue;
             }
 
-            relay::io::save(mesh,
-                            "braid_2d_" + name +  "_example.silo:mesh",
-                            "conduit_silo_mesh");
+            relay::io::silo::save_mesh(mesh, "braid_2d_" + name +  "_example.silo:mesh");
```
cyrush: Q: are we using `:mesh` here or should we remove that from the arg string?

src/tests/blueprint/t_blueprint_mesh_examples.cpp
```diff
@@ -367,11 +369,10 @@ TEST(conduit_blueprint_mesh_examples, mesh_3d)
                 continue;
             }
 
-            relay::io::save(mesh,
-                            "braid_3d_" + name +  "_example.silo:mesh",
-                            "conduit_silo_mesh");
+            relay::io::silo::save_mesh(mesh, "braid_3d_" + name +  "_example.silo:mesh");
```
cyrush: same q about `:mesh`

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -573,900 +420,4343 @@ conduit_wedge_connectivity_to_silo(Node &n_mesh_conn)
...
+        if (dtype.is_uint64())
         {
-            // TODO: check for explicit # of elems
-            int num_elems    = n_mesh_conn.dtype().number_of_elements() / 8;
-            shapetype[i] = DB_ZONETYPE_HEX;
-            shapesize[i] = 8;
-            shapecnt[i]  = num_elems;
-            total_num_elems  += num_elems;
-
+            silo_wedge_connectivity_to_conduit<uint64>(elements["connectivity"]);
```
cyrush, line 571: does this end up modifying the input mesh order?

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -573,900 +420,4343 @@ conduit_wedge_connectivity_to_silo(Node &n_mesh_conn)
...
+        origin["i"] = quadmesh_ptr->base_index[0];
+        if (ndims > 1) origin["i"] = quadmesh_ptr->base_index[1];
+        if (ndims > 2) origin["i"] = quadmesh_ptr->base_index[2];
+    }
 
-    n_mesh_info[topo_name]["num_pts"].set(num_pts);
+    copy_point_coords(quadmesh_ptr->datatype,
+                      quadmesh_ptr->coords,
+                      ndims,
+                      real_dims,
```
cyrush, line 706: do we need `real_dims` vs just `dims`, does it get updated somewhere that isn't obvious via viewing the diff?

cyrush (commented):
So far looking great, here is a a batch of a few quick questions

src/libs/relay/conduit_relay_io_silo.cpp
JustinPrivitera: In the collinear case (rectilinear grid), we want to use the dimensions the quadmesh provides directly. In the noncollinear case (structured grid), we are setting up an explicit coordset so the dimensions for x,y,z are the same, equal to the total number of nodes.

I create the dimensions for the structured case and then create a pointer (`real_dims`) to them so that if we hit the rectilinear case, I can make it point to the quadmesh dimensions directly.

See line 673.

There maybe is a better way to do this...

src/libs/relay/conduit_relay_io_silo.cpp
JustinPrivitera: wow this diff is so confused.

No this goes wedge by wedge and reorders the connectivity. The order of wedges stays the same. Is that what you were asking?

src/libs/relay/conduit_relay_io_silo.cpp
cyrush: that makes sense, I think I could not see b/c of how the diffs are presented.

src/libs/relay/conduit_relay_io_silo.cpp
cyrush: I was asking if the input mesh topology values are reordered. const node ref would not necessarily protect against that, and we don't want to mix up the inputs. 

For example (and a way to test): If you write twice do things get mixed up b/c the reorder happens twice. Does that make sense?

JustinPrivitera: > I was asking if the input mesh topology values are reordered

There shouldn't be any input mesh topology values... this code is for reading from silo. At this point, we are reading connectivity information from silo. We use node `set` to do that. If it is a wedge, we make changes to our copy.

Perhaps you are wondering about `conduit_wedge_connectivity_to_silo` instead? That is used on the writing side. Again, we make a copy using node `set` before using a function to reorder stuff.

src/tests/blueprint/t_blueprint_mesh_examples.cpp
JustinPrivitera: It just ends up in the file name: `braid_2d_lines_example.silo:mesh.cycle_000100.root`

I don't know what the behavior for 
```
            relay::io::save(mesh,
                            "braid_2d_" + name +  "_example.silo:mesh",
                            "conduit_silo_mesh");
```
was before, did that use mesh as an internal name or put it in the filename?

cyrush: prior did use mesh inside the file, but that doens't make sense with our new setup. 
Lets drop the substring `:mesh` from our tests.

src/libs/relay/conduit_relay_io_silo.cpp
cyrush: I was curious about the write side, sounds like we are good. Thanks for clarifying!

src/tests/blueprint/t_blueprint_mesh_examples.cpp
JustinPrivitera: ok

PR #856, opened 2021-10-29
Feature/il tabular blueprint

Laganellac:
This PR includes enhancements to the blueprint library to support a tabular data blueprint ("table") and the functionality of "flattening" blueprint meshes into blueprint tables in serial or parallel. The relay library has been enhanced to accept a "csv" protocol which will read/write blueprint tables to/from csv files (or a collection of csv files). Sphinx documentation has been added for blueprint table and language bindings have been updated to include the new functions.

The work in this PR depends on some of the work done in PR #838 so it is currently branched off of that branch.


cyrush:
@Laganellac  the redecomp branch is now in develop, so you can shift to target that as the base branch. 
If you hit any pot holes, let me know can I can try to help

Laganellac:
Hey @cyrush , when I try to target develop with this PR it includes a bunch of the commits from the il_blueprint_redecompose branch.

Are you saying I should target the il_blueprint_redecompose branch? When I do that it cleans up the PR a bit.

I could also try rebasing this branch on top of develop if that would make things easier.

Laganellac:
@cyrush I think I've cleaned things up, sorry for the confusion earlier.

Looks like there are a few CI errors that I've got to clear up still.

src/libs/relay/conduit_relay_io_csv.cpp
```diff
@@ -0,0 +1,663 @@
...
+    index_t retval = 0;
+    if(values.number_of_children() > 0)
+    {
+        const Node &ref = values[0];
+        const index_t nc = ref.number_of_children();
+        retval = nc > 0
+            ? ref[0].dtype().number_of_elements()
+            : ref.dtype().number_of_elements();
+    }
+    return retval;
```
cyrush, line 68: should this be a property method of the table blueprint?

src/libs/relay/conduit_relay_io_csv.cpp
```diff
@@ -0,0 +1,663 @@
...
+
+//-----------------------------------------------------------------------------
+// Q: Should this go in conduit utilities?
+static void
+trim(std::string &str, const char *t = whitespace)
+{
+    // ltrim, rtrim
+    str.erase(0, str.find_first_not_of(t));
+    str.erase(str.find_last_not_of(t) + 1);
+}
```
cyrush: +1 for this as `conduit::utils::trim_string`

cyrush (approved):
Looks great! Thanks for good docs + tests. Are there other aspects that you want to tackle before we get the initial capability merged into develop?

Laganellac:
Hey @cyrush, this should be ready to go.

I can add the two changes suggested above if you'd like.

src/libs/relay/conduit_relay_io_csv.cpp
Laganellac: Yea this would make sense.

I was probably hesitant to do this because the table blueprint supports multiple tables in one node (like a multi domain mesh), and this would be something that only works on a single table. The mesh blueprint supports a ton of operations like that so I don't see why a table shouldn't.

cyrush:
thanks @Laganellac!  
lets do the `utils::trim_string`change now and then add a ticket to think more about the nrows as a property.  No rush on that I want ponder a bit more from the context you provided.

cyrush:
@Laganellac - there are a few conflicts now, the ones in `src/libs/blueprint/conduit_blueprint.cpp` are straightforward to resolve, but the others require some though.  Would it be possible for you to take a pass at them?

Laganellac:
@cyrush Sure I can take a look at the conflicts and add the trim_string function

Laganellac:
Hey @cyrush, I've made a pass at the merge conflict and added the trim_string function (+ a small test).

One of the big conflicts is that the iterate_elements code was moved to a different file, I've applied the new data_accessor changes to iterate_elements in its new home.

The other conflict was due to a bugfix I made in partition to properly apply the new domain id to all output domains; there is a new call to adjset code in this same block. Currently applying new domain id then calling the adjset code; I will need to do some manual testing and then rebaseline the partition test.

cyrush:
@Laganellac thanks! Let me know when you feel the baselines are in good shape and we will merge.

Laganellac:
Hey @cyrush, I think we're in good shape here.

Turns out we don't need to rebaseline - I had assumed the new adjset stuff would alter the test output but that does not seem to be the case. I also ran this against the visit test that I've been writing for the new blueprint export functionality, everything looking good there too.

cyrush:
Great - thanks for checking everything!

cyrush:
thanks @Laganellac  and @BradWhitlock !

Issue #842, opened 2021-09-17
Braid triangle mesh should not accept three mesh dims?


mclarsen:
You can ask for a explicit triangle mesh `tris` and specify the dims as `20,20,20` but the resulting mesh only has a 2d coordinate system.

```
tate: 
  time: 3.1415
  cycle: 100
coordsets: 
  coords: 
    type: "explicit"
    values: 
      x: [-10.0, -8.94736842105263, -7.89473684210526, ..., 8.94736842105263, 10.0]
      y: [-10.0, -10.0, -10.0, ..., 10.0, 10.0]
```

mclarsen:
I assume the same thing happens for `quads`.

cyrush:
Worked on locking this down to `npts_z == 0`, throwing an exception if folks pass anything else for a 2D only topo.

In conduit's tests -- there are several places were we pass `npts_z == 1`. 

With indirection in our tests, it's not trivial to track them all down -- but i am working on it. 

Since they were in many tests, this usage might be prevalent in the wild as well.  

Instead of an error, may use `CONDUIT_INFO` as a stop gap, switch to `CONDUIT_ERROR` in a future release

cyrush:
resolved in #868

PR #868, opened 2021-11-08
add more input guards mesh bp braid

cyrush:
audit uses of braid for 2d only shape types with `npts_z != 0`
add plumbing to allow a strict mode (currently off)


cyrush:
will resolve #842

cyrush:
issues a info message when `npts_z !=0` case is detected.

Added ticket #869 to tackle converting to error in future release.

PR #871, opened 2021-11-09
Missing includes for gcc 11

CharlesGueunet:
add some missing limits include for numeric_limits use


cyrush:
thanks @CharlesGueunet  -- appreciate these fixes. 

I'll try to add gcc 11 to our CI in the near future to catch these in the future

PR #878, opened 2021-11-30
to_polytopal

### LINKED ISSUE #745: change `topology::unstructured::to_polygonal` to `to_polytopal`?
The new name would more accurately describe what the function does, but it would require updating some downstream libraries. This decision will be important as it will also impact #642 and the final names used for the functions in this PR.

Tagging: @cyrush, @nselliott

cyrush:
resolves #745 

- adds non mpi and mpi to_polytopal 
- adds mpi comm argument to mpi to_poly methods
- adds some input validate to the mpi to_poly


nselliott (approved):
Looks good, thanks.

PR #894, opened 2021-12-21
Account for shared vertices in parmetis field generator

publixsubfan:
* Adds a user-specified option for an adjset name, which is used to determine global vertex ids of shared nodes between domains
* Fixes a bug identified in the re-partitioner where adjsets for the new domains would sometimes be missing.


publixsubfan:
## Partition without adjsets:
This is a partitioning from 4 domains -> 5 domains, run with 4 ranks.
image
And the `global_vertex_ids` field shows discontinuities at the domain boundaries, indicating that vertices aren't being shared across domains:
image


## Partition with adjsets:
image
The `global_vertex_ids` field no longer shows discontinuities at the domain boundaries.
image

cyrush (approved):
Looks good!

cyrush:
thanks @kanye-quest !

Issue #909, opened 2022-01-25
Windows Python 3.8 (+?) DLL Resolution


cyrush:
starting with 3.8, python on windows changed dll resolution strategy for modules. 

Prior to this, we were able to provide the path to Conduit DLLs via `PATH` env var and Python would resolve them on start up.

For Python 3.8, this behavior was disabled, here are some docs and discussions related to this change:

https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew

https://docs.python.org/3/library/os.html#os.add_dll_directory

https://bugs.python.org/issue43173

cyrush:
This is the cause of the symptoms in: #695

cyrush:
resolved in #910

Issue #893, opened 2021-12-21
conda forge build issues with 0.8.0 release


cyrush:
linux builds: python 3.8 and 3.9 builds do something different with python path:

```
CMake Error at cmake/thirdparty/SetupPython.cmake:164 (MESSAGE):
  Failed to find main library using
  PYTHON_EXECUTABLE=/home/conda/feedstock_root/build_artifacts/conduit_1640064348711/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/bin/python
Call Stack (most recent call first):
  cmake/Setup3rdParty.cmake:77 (include)
  CMakeLists.txt:83 (include)
```

Python 3.8 build issue on windows only:
```
conduit_python.cpp
%SRC_DIR%\src\libs\conduit\python\conduit_python.cpp(3184): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
%SRC_DIR%\src\libs\conduit\python\conduit_python.cpp(3184): error C2440: 'initializing': cannot convert from 'const char [20]' to 'int'
%SRC_DIR%\src\libs\conduit\python\conduit_python.cpp(3184): note: There is no context in which this conversion is possible
%SRC_DIR%\src\libs\conduit\python\conduit_python.cpp(8085): error C2027: use of undefined type 'module_state'
%SRC_DIR%\src\libs\conduit\python\conduit_python.cpp(8034): note: see declaration of 'module_state'
%SRC_DIR%\src\libs\conduit\python\conduit_python.cpp(8088): error C2027: use of undefined type 'module_state'
%SRC_DIR%\src\libs\conduit\python\conduit_python.cpp(8034): note: see declaration of 'module_state'
%SRC_DIR%\src\libs\conduit\python\conduit_python.cpp(8098): error C2065: 'PyConduit_DataType_TYPE': undeclared identifier
%SRC_DIR%\src\libs\conduit\python\conduit_python.cpp(8103): error C2065: 'PyConduit_Schema_TYPE': undeclared identifier
%SRC_DIR%\src\libs\conduit\python\conduit_python.cpp(8108): error C2065: 'PyConduit_Generator_TYPE': undeclared identifier
%SRC_DIR%\src\libs\conduit\python\conduit_python.cpp(8113): error C2065: 'PyConduit_NodeIterator_TYPE': undeclared identifier
[...]

Issue #695, opened 2021-02-05
PATH issues for python unit tests on some windows systems


cyrush:
Had issues running python tests on my windows system, I think this could use a TO_NATIVE path treatment for cases where msys is in play:

https://github.com/LLNL/conduit/blob/6dbbe7bf718203fa75e0a5b9bb29689da5101233/src/cmake/SetupTests.cmake#L176

(Tests run if I copy dlls in, so pretty certain its a PATH issue)

cyrush:
resolved in #910

PR #910, opened 2022-01-25
windows updates for newer python versions

cyrush:


cyrush:
resolves #909 #893 #695

Issue #913, opened 2022-01-26
src/libs/blueprint/conduit_blueprint_mpi_mesh_parmetis.cpp need to #include <algorithm>?


lee218llnl:
I am trying to build conduit-0.8.1 in spack and got this error:

```
/tmp/lee218/spack-stage/spack-stage-conduit-0.8.1-bmqqtmgvjptitcw3gj3xie6jcfnaj6au/spack-src/src/libs/blueprint/conduit_blueprint_mpi_mesh_parmetis.cpp:451:18: error: 'sort' is not a member of 'std'
             std::sort(recv_groups.begin(), recv_groups.end());
```

Should this source file ```#include <algorithm>```?

cyrush:
@lee218llnl -- thanks for the report -- what compiler are you using? 

For a quick fix, we can add a patch to the spack package.

lee218llnl:
$ gcc --version
gcc (GCC) 8.5.0 20210514 (Red Hat 8.5.0-4)

lee218llnl:
I should add that this is for TOSS 4, so you can test on corona if you have access

cyrush:
thanks, I'll take a look!

lee218llnl:
I created a patch in my personal copy of spack, so no rush on this. I could submit a PR to spack if you like, but I'm not sure which release versions of conduit this would apply to.

cyrush:
Great -  the patch would apply to `0.8.0` and `0.8.1` (that feature is new starting in `0.8.0`. If you can add it to spack that would be great, or I can do it later this week.

lee218llnl:
It looks like 0.8.0 doesn't use std::sort, so I just added 0.8.1 to the spack package along with a patch that just includes the algorithm header.

lee218llnl:
do you need me to submit a PR to conduit itself?

cyrush:
that would be great :-)

PR #916, opened 2022-01-27
heal parmetis ci testing

cyrush:
I was a bit surprised we didn't catch https://github.com/LLNL/conduit/issues/913,

turns out when we updated ci cases, the wiring for parmetis testing was lost.


cyrush:
parmetis testing is back online :-)

PR #917, opened 2022-01-28
Add feature to map back fields post-partitioning

publixsubfan:
This PR adds a function `conduit::blueprint::mesh::partition_map_back` (and MPI-aware equivalent) which uses the `original_element_ids` field to map a zone-centered field from the repartitioned mesh back onto the original mesh.
- This is implemented in the `Partitioner` class so we can reuse the field slicing/merging functions implemented there.
- A `Partitioner::get_global_domids()` function is used to ensure that globally-unique domain IDs are generated consistently for a pre-partition mesh.


cyrush:
Awesome work! 

For the main interface `partition_map_back` - can we switch to use a conduit node to provide the field names?

(This allows us to expose the args uniformly in C++, python, fortran,etc)

Also, are their any caveats about over writing the fields in original mesh?

Would this be more composable if we had something like:

```
partition_map_back(const Node&part_mesh, const Node &org_mesh, const Node &opts, Node &fields)
```

Then we could then choose where to put the fields:

```
partition_map_back(part_mesh, org_mesh, opts, orig_mesh["fields"]);

vs

partition_map_back(part_mesh, org_mesh, opts, some_other_node);

```

Not sure if this is feasible or useful, thoughts?

publixsubfan:
I agree, the second interface seems much nicer.

One thought: how does this play with multiple domains on a single processor? Would we pass in something like `orig_mesh[0]["fields"]` for the `fields` argument, and do some node hopping to get to the other domain targets?

cyrush:
Hmm, I agree that's a complication.  In other cases we are able to do it domain by domain (like with `generate_sides`), but that won't work in this case. 


Need to think about the best approach.

publixsubfan:
I've added partial support for mapping back vertex-centered fields. The issue surrounding values on shared nodes is still TODO; the `original_vertex_ids` field needs to capture that a single post-partition node may have been merged from nodes on multiple pre-partition domains.

Effectively, we need to store a one-to-many relation from nodes to original vertex IDs and domains -- thoughts on how we can do this?

PR #926, opened 2022-03-02
Example for arbitrary-sized storage of vertex and element fields.

agcapps:
This PR adds an example that shows how to use arbitrary-sized arrays for vertex and element fields in structured mesh data.  This allows the zero-copy use of arrays that store ghost zones with Conduit.


agcapps:
I guess I could have written "over-sized" instead of "arbitrary-sized," since the arrays storing field data do have lower bounds on their sizes.

rhornung67 (approved):
Thank you @agcapps

cyrush:
Looks good so far!

I tried to think of a more specific name than `arbitrary` -- how about: `strided_structured`

cyrush (commented):
One suggestion: Can we add an ASCII Art diagram to the example?
That context is super helpful for examples like this.

cyrush (approved):

agcapps:
@cyrush , once I went to the trouble of logging out and logging back in, I was able to merge this PR and delete the branch.

cyrush:
thanks!

PR #936, opened 2022-03-15
add test to demo multi topo load balancing and map back

cyrush:


src/libs/blueprint/conduit_blueprint_mesh_examples_related_boundary.cpp
```diff
@@ -707,7 +713,7 @@ void related_boundary(index_t base_grid_ele_i,
     mesh["domain1/adjsets/main_adjset/topology"] = "main";
     Node &d1_adj_groups = mesh["domain1/adjsets/main_adjset/groups"];
 
-    d1_adj_groups["group_0_1_2/neighbors"].set({1,2});
+    d1_adj_groups["group_0_1_2/neighbors"].set({0,2});
     // one point is shared between all three
```
cyrush, line 717: thanks for solving this issue @kanye-quest !

cyrush:
@kanye-quest I think this is now ready, it demos the steps we are interested in. Do you think we should include anything else? 

Thanks again for catching the adjset issue.

PR #937, opened 2022-03-16
improve pip install logic

cyrush:


cyrush:
the cpp py fort issue with python 2.7 static case is related to python modules load failing to hdf5 shared libs at runtime. 🤯

PR #938, opened 2022-03-17
Fix partitioner handling of adjsets for multi-topology domains

publixsubfan:
Currently, the partitioner assumes that all adjsets are associated with the topology being partitioned. When partitioning a related topology, we need to skip over any adjsets that aren't associated with the topology.

Related to #936


cyrush (approved):
thanks for this fix!

PR #946, opened 2022-03-28
add input and result output for to_poly 2d unit test cases

cyrush:


cyrush:
@agcapps this adds output files to the basic to 2D poly tests

PR #949, opened 2022-03-29
Add topology reindexing support

publixsubfan:
Adds a `conduit::blueprint::mesh::utils::topology::reindex_coords()` function to remap a related topology to a main coordset post-partitioning.

This function requires passing global vertex ID fields for both the main and the related topology in order to determine the correct relation of coordset indices post-partitioning, since both post-partition topologies might have different associated coordsets.


cyrush (approved):

cyrush:
looks good, thanks!

agcapps (commented):
Thank you, @kanye-quest

PR #951, opened 2022-03-30
Prevent scalars from including component names in CSV output.

BradWhitlock:
I made a 1 line change into the CSV writing code in relay so scalars do not get an unwanted component name appended. This change fixed an issue reported in VisIt and did not break any tests here when I ran the test suite locally.


cyrush (approved):
thanks @BradWhitlock !

BradWhitlock:
Sure! Sorry that was not caught earlier.

Issue #950, opened 2022-03-30
Add 3D mesh to t_blueprint_mpi_mesh_polytopal


nselliott:
A simple 2-domain 3D mesh should be added to do the same things that are being done in 2D in this test.

PR #952, opened 2022-03-30
Add simple 3D AMR mesh to blueprint polytopal test

nselliott:
For issue #950

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
@@ -1160,6 +1160,8 @@ void to_polyhedral(const Node &n,
 {
     dest.reset();
 
+    Node temp;
+
     index_t par_rank = relay::mpi::rank(comm);
 
     NodeConstIterator itr = n.children();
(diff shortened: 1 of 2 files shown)


src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
```diff
@@ -1252,9 +1254,10 @@ void to_polyhedral(const Node &n,
                 const Node& group = grp_itr.next();
                 if (group.has_child("neighbors") && group.has_child("windows"))
                 {
-                    int64_array neighbors = group["neighbors"].as_int64_array();
-
-                    index_t nbr_id = neighbors[1];
+                    temp.reset();
```
cyrush, line 1257: FYI: There is a new pattern you can use to simplify logic like this, you can use:

int64_accessor vals = group["neighbors"].value();

`vals`  will give you access as an int64 regardless of underlying type, and it will handle the offset, striding.

cyrush (approved):
Looks good, thanks for adding this!

I sent a FYI comment about new Conduit API that we can used to simplify generic access to arrays.

agcapps (approved):
Thank you, @nselliott .

src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
agcapps: Thanks for the comment, @cyrush !  this will be handy for some of my code as well.

nselliott: I'm going to leave this as it is, as I was trying to access everything as `index_t` rather than `int64`, but it had been missed in a few places in this file; this tip will be helpful for future coding.

nselliott:
I don't have merging privileges--please merge when appropriate.

src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
cyrush: You can also use `index_t_accessor`.

Regardless - find to leave as is, I think we will do a large pass a pass in the future to simplify cases like this. 

Thanks again!

PR #954, opened 2022-04-08
add polystar example

cyrush:
adds mesh blueprint polystar example, that generates a mesh with a polyhedral star pattern that demonstrates  hanging vertices in a topology. Includes various fields that count element to vertex and vertex to element relationships. Also generates corner and side meshes from the main topology

main topo:
Screen Shot 2022-04-08 at 4 22 15 PM

sides mesh:
polystar_sides

corners mesh:

polystar_corners

(note: looks like this is also demos a non-idea case for centroid creation?


cyrush:
@agcapps this example should help frame discussion about hanging nodes and `generate_corners` support.

agcapps (approved):

PR #968, opened 2022-05-11
Add a 1D basic example

agcapps:
Add basic1D() to generate a 1D example.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
@@ -130,6 +130,21 @@ void basic_init_example_element_scalar_field(index_t nele_x,
 
 //---------------------------------------------------------------------------//
 bool
+braid_1d_allowed_shape_type(const std::string& mesh_type)
+{
+    if ( mesh_type == "structured"  ||
+         mesh_type == "uniform")
+    {
+        return true;
+    }
+    else
+    {
+        return false;
+    }
+}
+
+//---------------------------------------------------------------------------//
+bool
 braid_2d_only_shape_type(const std::string& mesh_type)
 {
     if ( mesh_type == "tris"  ||
@@ -2245,18 +2260,22 @@ basic(const std::string &mesh_type,
     // NOTE(JRC): The basic mesh example only supports simple, homogenous
     // element types that can be spanned by zone-centered fields.
     const std::string mesh_types[] = {
+        "1Duniform", "1Dstructured",
         "uniform", "rectilinear", "structured",
         "tris", "quads", "polygons",
         "tets", "hexs", "polyhedra"};
     const std::string braid_types[] = {
+        "uniform", "structured",
         "uniform", "rectilinear", "structured",
         "tris", "quads", "quads_poly",
         "tets", "hexs", "hexs_poly"};
     const index_t mesh_types_dims[] = {
+        1, 1,
         2, 2, 2,
         2, 2, 2,
         3, 3, 3};
     const index_t mesh_types_subelems_per_elem[] = {
+        1, 1,
         1, 1, 1,
         2, 1, 1,
         6, 1, 1};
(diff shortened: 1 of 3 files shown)


src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2270,6 +2327,61 @@ basic(const std::string &mesh_type,
 }
 
 
+//---------------------------------------------------------------------------//
+void
+basic1D(const std::string &mesh_type,
+        index_t npts_x, // number of points in x
+        Node &res)
```
agcapps: This is `basic()`, specialized for 1D mesh types and checks.  Should it should be wrapped into `basic()`?

agcapps:
Need to add support for structured meshes

agcapps:
Need to test construction and saving

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
cyrush: I think wrapping it in basic (case where npts_y and npts_z == 0) is a good approach.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2270,6 +2327,61 @@ basic(const std::string &mesh_type,
...
+    }
+
+    const bool npts_x_ok = npts_x > 1;
+
+    // don't let de-morgan get you ...
+    if( ! (npts_x_ok) )
+    {
+        // error, not enough points to create the topo
+        CONDUIT_ERROR("blueprint::mesh::examples::basic1D requires npts_x > 1" << std::endl <<
+                      << std::endl <<
```
cyrush: I think we have a double `<<` here causing the compile problems we see in CI.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -266,9 +302,11 @@ void braid_init_example_point_vector_field(index_t npts_x,
         dx = 20.0  / float64(npts_x - 1);
     }
 
-    if(npts_x > 1)
+    // AGC is this right?  Note change to variable being tested
+    // and used in divisor.
+    if(npts_y > 1)
     {
-        dy = 20.0  / float64(npts_x - 1);
+        dy = 20.0  / float64(npts_y - 1);
```
cyrush, line 309: Yes, looks like you found a bug. Good eye!

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2718,6 +2802,10 @@ basic(const std::string &mesh_type,
     {
         // error, not enough points to create the topo
         CONDUIT_ERROR("blueprint::mesh::examples::basic requires: " << std::endl <<
+                      "For 1D only topologies"
+                      " ( mesh_type={\"1Duniform\" or \"1Dstructured\"} )"
+                      " npts_x > 1 and npts_y == 0 and npts_z == 0"
+                      << std::endl <<
```
cyrush, line 2833: Can we just support with `mesh_type="uniform"` with 1D appropriate args, instead of a new type `1Duniform`?

cyrush (commented):
One q about the new mesh_type option for `basic`.

Also, can you add an entry into the changelog?

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
agcapps: I would like to.  I did it this way because `mesh_types_dims[mesh_type_index]` specifies the dimensionality for `mesh_types[mesh_type_index]`.  If we were to do away with `1Duniform` as a mesh type, I would make a new variable `int dims = mesh_types_dims[mesh_type_index]` and then set `dims = 1` if `npts_y == 0 && npts_x == 0`.

cyrush: I think it's worth doing that refactor.

cyrush:
We are green, merging. Thanks @agcapps !

PR #969, opened 2022-05-19
Change name of map in matsets generated index to "materials"

nselliott:
The automatic generation of the blueprint index used "material_map" but VisIt expects "materials".

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -1810,7 +1810,7 @@ mesh::generate_index_for_single_domain(const Node &mesh,
                 while(mats_itr.has_next())
                 {
                     mats_itr.next();
-                    idx_matset["material_map"][mats_itr.name()] = mats_itr.index();
+                    idx_matset["materials"][mats_itr.name()] = mats_itr.index();
                 }
             }
             else // surprise!


cyrush:
thanks @nselliott -- can you send me the details about the VisIt error?

The VisIt plugin has logic to handle `materials` and `material_map`, so it sounds like there is a bug there. 

We are presenting a `material_map` here vs a list of material names. 

Maybe one work around could be to create both `material_map`  and `materials` -- but that depends on the nature of the bug in the VisIt plugin.

nselliott:
This is what I found in a `visit -debug 5` log file.  I'll send you the file.

```
   "matsets":
    {
      "valid": "false",
      "mats":
      {
        "topology":
        {
          "valid": "true"
        },
        "materials":
        {
          "valid": "false"
        },
        "errors":
        [
          "mesh::matset::index: missing child 'materials'"
        ],
        "path":
        {
          "valid": "true"
        },
        "valid": "false"
      }
    }, 
```

This was specifically in a case where there aren't material names, so it creates numerical strings "001", "002", etc. to label the materials.

cyrush:
this might actually be a bug in blueprint index verify, let me look into it a bit more.

cyrush:
This is an issue with a VisIt install that is using an older version of conduit. 
The index verifies with `material_map` when we use a newer conduit. 

We need to think about a way to support proper case, while still bridging for production version of VisIt

cyrush:
@nselliott 

VisIt 3.3.0 will release using a newer conduit.

We have a beta on the RZ, can you check and see if it can read the non modified file?

`/usr/gapps/visit/bin/visit -v 3.3`

nselliott:
Closing this, as I have confirmed that VisIt 3.3 works without needing this change.

cyrush:
Awesome, thanks for confirming!

PR #970, opened 2022-06-03
cmake: use the commands that get the git information directly

mathstuf:
Asking `git describe` to `--match` nothing is just a long-winded way of
asking for the hash. Instead, ask for the hash and dirty status
directly.

Note that this is not ideal anyways as CMake can have stale information
as the build is not regenerated if the working tree changes at all.

---
Note that the better solution is to instead grab the git information at *build* time. This makes sure that this information is always up-to-date with any build. I've done it before [here](https://github.com/Kitware/sprokit/blob/master/src/sprokit/CMakeLists.txt#L8) but there is significant other machinery below it.


cyrush:
Thanks for the improvement!

PR #972, opened 2022-06-07
Fix missing c api impls

mathstuf:


cyrush:
@mathstuf thanks for fixing these!
Would it be possible to add simple tests to make sure they stay fixed?

They would go inside one of these two files:

https://github.com/LLNL/conduit/blob/develop/src/tests/conduit/c/t_c_conduit_node.cpp
https://github.com/LLNL/conduit/blob/develop/src/tests/conduit/c/t_c_conduit_node_set.cpp

mathstuf:
I'm not sure how to use the `data_ptr` and `element_ptr` functions. I found these while testing Catalyst's ABI where we export Conduit's C API and I want to detect when anything changes in an ABI-breaking way.

mathstuf:
Whoops, still a typo here.

cyrush:
looks good, thanks again!

Issue #971, opened 2022-06-07
Polyhedral/Polygonal support for stream-based mesh blueprint - how to best do this?


mennodeij:
Good morning,

I'm working on adding support for mixed cell streams to the VTK/ParaView implementation of Conduit/MeshBlueprint for in-situ processing. A [merge request](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/9240) is pending at the moment , but before this is merged I wanted to check here if my approach is valid when using `element_counts` or `offsets`. Two examples are given below for these two approaches.

My question is, other polyhedral cell examples have `sub_elements` for the faces. Below I have not done that, and the whole cell is defined in the stream with the VTK approach, i.e. `nfaces, npoints_face1, pt_0, pt_1, ... ptN, npoints_face2, pt_0, pt_1, ..., ptM, ...`

Is this a good approach, or would you prefer the approach with sub_elements? 

Polyhedral example for three polyhedra with `element_counts`:
```
coordsets:
  coords:
    type: "explicit"
    values:
      x: [0.0, 1.0, 1.0, 0.0, 0.0, 1.0, 1.0, 0.0, 0.5, 0.5]
      y: [0.0, 0.0, 1.0, 1.0, 0.0, 0.0, 1.0, 0.0, 0.5, 0.5]
      z: [0.0, 0.0, 0.0, 0.0, 1.0, 1.0, 1.0, 1.0, -1.0, 2.0]
topologies:
  mesh:
    type: "unstructured"
    coordset: "coords"
    elements:
      element_types:
        polyhedral:
          stream_id: 0
          shape: "polyhedral"
      element_index:
        stream_ids: 0
        element_counts: 3
      stream: [5, 4, 4, 5, 6, 7, 3, 4, 5, 9, 3, 5, 6, 9, 3, 6, 7, 9, 3, 7, 4, 9, 5, 4, 0, 1, 2, 3, 3, 0, 1, 8, 3, 1, 2, 8, 3, 2, 3, 8, 3, 3, 0, 8, 6, 4, 0, 1, 4, 4, 4, 1, 2, 6, 5, 4, 2, 3, 7, 6, 4, 3, 0, 4, 7, 4, 0, 1, 2, 3, 4, 4, 5, 6, 7]
```

2) Same example but with offsets:
```
coordsets:
  coords:
    type: "explicit"
    values:
[...]

PR #977, opened 2022-06-16
Mixed protocol initial implementation ready for review

mennodeij:
As previously discussed in #971 - this is an initial implementation with examples for a mixed element definition using the `o2mrelation` for the connectivity.

I'm looking forward to having this reviewed and making changes if needed.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -469,13 +469,13 @@ bool verify_shapes_node(const Node &node, const Node &shape_map, Node &info)
   const std::string protocol = "mesh::topology::unstructured";
   bool res = true;
   res &= verify_integer_field(protocol, node, info);
-  
-  DataAccessor<int32> shapes(node.data_ptr(), node.dtype());
+
+  const int32_accessor shapes = node.value();
 
   std::vector<int32> range;
   for(auto &child : shape_map.children())
   {
-    range.emplace_back(child.as_int32());
+    range.emplace_back(child.to_int32());
   }
 
   for(index_t i = 0; i < shapes.number_of_elements(); ++i)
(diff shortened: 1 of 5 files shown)


cyrush:
@mennodeij  this looks great! Thanks for adding these and updating the verify methods.

If we can add the example `braid` and `vel` fields to both of these new examples then we can merge. 

Adding these vertex associated fields should be easy b/c the number of vertices are the same as other braid examples.
```
    Node &fields = res["fields"];

    braid_init_example_point_scalar_field(npts_x,
                                                                 npts_y,
                                                                 npts_z,
                                                                 fields["braid"]);

    braid_init_example_point_vector_field(npts_x,
                                                                 npts_y,
                                                                 npts_z,
                                                                 fields["vel"]);
```

Adding an element associated field, such as the `radial` example, is a bit more complex -- since the number of elements and the new pattern with respect to the original field / elements is non trivial. We can leave out `radial` for now.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -464,6 +464,83 @@ bool verify_reference_field(const std::string &protocol,
...
+  }
+
+  if (res)
+  {
+    for(auto &child : shape_map.children())
+    {
+
+      log::info(info, protocol, "cells found for shape "
+        + child.name() + " ("
+        + std::to_string(child.as_int32())
```
cyrush: you can use `to_int32()` and it will support a broader set of types

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -464,6 +464,83 @@ bool verify_reference_field(const std::string &protocol,
     return res;
 }
 
+bool verify_shapes_node(const Node &node, const Node &shape_map, Node &info)
+{
+  const std::string protocol = "mesh::topology::unstructured";
+  bool res = true;
+  res &= verify_integer_field(protocol, node, info);
+  
+  DataAccessor<int32> shapes(node.data_ptr(), node.dtype());
```
cyrush: FYI: You can simplify this pattern using:
`int32_accessor shapes = node.value();`

mennodeij: Done in 52049eb

src/libs/blueprint/conduit_blueprint_mesh.cpp
mennodeij: Done in 52049eb

mennodeij:
<strike>I'm looking at the mixed and mixed_2d examples and found some issues with the actual cell definitions - please hold off the merge until that is fixed.

mennodeij:
The mixed cell definitions are now fixed.

cyrush:
Merged. Thanks for incorporating feedback & thanks again for implementing this!

PR #980, opened 2022-06-20
Add documentation for mixed shapes. Fix one example.

mennodeij:
Add documentation on the new way to specify mixed elements, see #977 

(note: I saved with removal of trailing whitespace, so there are a lot of apparent changes, but the relevant changes are at line 257, 384 and and 1231).

Also, I noticed that in the mixed_2d example the index calculation was not updated everywhere yet. The results are the same, but the code is now consistent with the rest of the examples.


cyrush (approved):
Looks great, thanks for adding this!

PR #985, opened 2022-07-19
all_gather handle case the input node is compact

nicolemarsaglia:
all_gather was setting the recv_node schema to the compact node that is created if the input node is not compact. If the input node was compact it would break in odd and unexpected ways. 

fixes PR [https://github.com/Alpine-DAV/ascent/pull/839](https://github.com/Alpine-DAV/ascent/pull/839)

### DIFF
--- a/src/libs/relay/conduit_relay_mpi.cpp
+++ b/src/libs/relay/conduit_relay_mpi.cpp
@@ -990,25 +990,27 @@ all_gather(Node &send_node,
     const void *snd_ptr  = send_node.contiguous_data_ptr();
     index_t     snd_size = send_node.total_bytes_compact();
 
-    int mpi_size = mpi::size(mpi_comm);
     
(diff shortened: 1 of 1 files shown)


cyrush:
Looking over, I think the right answer for the `recv_node` might always be:

```
recv_node.list_of(s_snd_compact,
                            mpi_size);
```

it always needs a list of `mpi_size` copies of the compact schema version of the node that is sent (that is what s_snd_compact should be in all cases)

cyrush:
```
recv_node.list_of(n_snd_compact.schema(),mpi_size)
```

will be bogus if we didn't actually populate `n_snd_compact` (which would be the compact case :-))

cyrush:
thanks @nicolemarsaglia !

nicolemarsaglia:
No prob. Thanks for the suggestion! @cyrush

PR #987, opened 2022-07-28
expose to_json() conduit method to the c api

lucas-givord:
Hello,

in order to be able to convert a conduit node to a string, I need to expose the `to_json` conduit method to the c api. First, Is there a particular reason why this method is not already exposed in the c api ?

I propose in my commit a version of how we can expose it, otherwise we can also expose this method like this (in `conduit_node_c.cpp`) :
```c
void conduit_node_to_json(conduit_node *cnode, conduit_node *output_node)
{
    std::string str = cpp_node(cnode)->to_json();
    cpp_node(output_node)->set_string(str);
}
```

what did you prefer?

Thanks,
Lucas

### DIFF
--- a/src/libs/conduit/c/conduit_node.h
+++ b/src/libs/conduit/c/conduit_node.h
@@ -211,6 +211,10 @@ CONDUIT_API void conduit_node_load(conduit_node *cnode,
                                    const char* protocol);
 
 //-----------------------------------------------------------------------------
+// the caller must free the result
+CONDUIT_API char* conduit_node_to_json(const conduit_node *cnode);
+
+//-----------------------------------------------------------------------------
 //-----------------------------------------------------------------------------
 // Conduit Node "Set" Methods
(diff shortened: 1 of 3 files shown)


cyrush:
@lucas-givord-kitware  - Thanks!

Your proposed change looks good and I will merge post CI.

There is no underlying reason why it is omitted.  Most of the C-API was focused on coupling to Fortran, so there are gaps. 

Your change leaves open the option to add method that supports the full `to_json` signature in the future, which could be important. 

It would also be good to add `to_yaml`, `to_string`, `to_summary_string`, etc -- I'll created https://github.com/LLNL/conduit/issues/988 to track this.

lucas-givord:
Thank you for your quick reply!

It will indeed be very useful to have these methods to the c api.
I also added a unit test in my last commit.

cyrush:
merged, thanks for adding the unit test!

PR #1005, opened 2022-08-16
add any tag any recv variants of mpi recv methods

### LINKED ISSUE #990: Allow recv_using_schema to work with MPI_ANY_SOURCE and MPI_ANY_TAG
This requests a small change in `conduit::relay::mpi::recv_using_schema` to make it work properly when the source is `MPI_ANY_SOURCE`.

This currently doesn't work because the `MPI_Recv` doesn't necessarily catch the message found by the preceding `MPI_Probe`, because it doesn't specifically identify the message it wanted (and allocated memory for).  To fix this, replace
```
mpi_error = MPI_Recv(n_buffer.data_ptr(),
                     buffer_size,
                     MPI_BYTE,
                     src,
                     tag,
                     comm,
                     &status);
```
with
```
mpi_error = MPI_Recv(n_buffer.data_ptr(),
                     buffer_size,
                     MPI_BYTE,
                     status.MPI_SOURCE,
                     tag,
                     comm,
                     &status);
```

Better yet, it can also support `MPI_ANY_TAG` as well, with
```
mpi_error = MPI_Recv(n_buffer.data_ptr(),
                     buffer_size,
                     MPI_BYTE,
                     status.MPI_SOURCE,
                     status.MPI_TAG,
                     comm,
                     &status);
```

cyrush:
This should resolve #990

@gunney1 -- can you take a look and see if these variants will work for your use case?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 #### Blueprint
 - Added `blueprint::mesh::paint_adjset`, which paints fields that encode adjacency set counts and ordering details. 
 
+#### Relay
+- Added any source, any tag variants of mpi receive functions: `recv`, `recv_using_schema`, and `irecv`.
+
 ### Changed
 
 #### Relay
--- a/src/libs/relay/conduit_relay_mpi.cpp
+++ b/src/libs/relay/conduit_relay_mpi.cpp
@@ -363,6 +363,56 @@ recv_using_schema(Node &node, int src, int tag, MPI_Comm comm)
 }
 
 //---------------------------------------------------------------------------//
+// any source, any tag variant
+int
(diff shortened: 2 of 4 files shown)


cyrush:
@gunney1 I am going to merge so we have these new methods. 
If there are other cases we need to address, lets create a new ticket.

gunney1:
@cyrush I was wondering why a new variant instead of modifying the existing versions to work properly when the user specifies MPI_ANY_SOURCE and/or MPI_ANY_TAG.  If I'm reading it right, the new variant seems to support only MPI_ANY_SOURCE **and** MPI_ANY_TAG.  So I can't say "any source, but only this tag" or any tag, but only this source."  I think my original proposal should work, unless there's something I'm not aware of.

cyrush:
@gunney1 -- I see, yes we should change the existing versions so MPI_ANY_SOURCE and MPI_ANY_TAG work. I'll get those changes in.

PR #1018, opened 2022-09-29
1D to Strip

agcapps:
Adds a verify() function and strip-conversion routines


agcapps:
This is at the discussion stage.  In particular, the results need to be tested against other libraries that expect strip meshes, and we need to examine assumptions.

agcapps:
Currently, the `oneD::verify` function expects a structured mesh, as do the conversion routines.  `verify`, at least, should support uniform meshes (specifying a dx and an origin).  The conversion routines should support uniform meshes if needed by users.  In addition, we should probably also support conversion between 1D structured and 1D uniform and back.

cyrush:
Instead of a new verify case, can we identity existing verified meshes with a subset of properties for this case?

cyrush:
Suggested methods for strip creation:

Suggested top level method:
```
blueprint::mesh::generate_strip(conduit::Node& mesh,
      const std::string& src_topo_name,
      const std::string& dst_topo_name);
```


Suggested topology specific method:

```
blueprint::mesh::topology::generate_strip(const conduit::Node &topo,
       conduit::Node &topo_dest,
       conduit::Node &coords_dest)
```

cyrush:
Q: For the result strip mesh, does that have to be uniform/rectilinear/structured for the carter use case, or could it be unstructured for any case we could care about?

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2053,6 +2210,280 @@ mesh::generate_index_for_single_domain(const Node &mesh,
...
+mesh::is_convertible_to_one_dimension(const Node &mesh,
+                                      Node &info)
+{
+    const std::string protocol = "convertible_to_one_dimension";
+    bool res = true;
+    info.reset();
+
+    // Each coordset c in mesh["coordsets"] must have two children of c["values"]
+    bool consistent_dim = is_dimension_consistent(mesh);
+    if (!consistent_dim)
```
cyrush: We will be interested in a specific topology, so I thinl we can check the coordset linked with that topology vs all possible coordsets.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2053,6 +2210,280 @@ mesh::generate_index_for_single_domain(const Node &mesh,
...
+    {
+        log::error(info, protocol, "inconsistent dimensionality in the mesh coordsets");
+        res = false;
+    }
+
+    index_t mesh_dim = max_dimension(mesh);
```
cyrush: We can use: `conduit::blueprint::mesh::utils::coordset::dims` for the specific coordset

agcapps:
> 

I don't know for sure if Carter wants uniform/rectilinear/structured.  I will try to track down a test case and model off that.

src/libs/blueprint/conduit_blueprint_mesh.cpp
agcapps: That will probably make the code more specific and constrained (in this case, both good things).

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2053,6 +2148,73 @@ mesh::generate_index_for_single_domain(const Node &mesh,
...
+    mesh::topology::generate_strip(src_topo, dst_topo);
+    mesh::coordset::generate_strip(src_coordset, dst_coordset);
+
+    mesh["topologies"][dst_topo_name] = dst_topo;
+    mesh["coordsets"][dst_topo_name] = dst_coordset;
+
+
+    // Flag this as a 1D strip mesh
+    mesh["private_nonblueprint/carterflags/oneD_strip"] = 1;
+}
```
cyrush, line 2183: Should we put this flag in the conduit carter interface instead of here?

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -5828,6 +6007,40 @@ mesh::field::index::verify(const Node &field_idx,
...
+void
+mesh::field::oneD::strip_to_oneD(conduit::Node &field,
+                                 conduit::Node &info)
+{
+    // verify?  or live dangerously?
+    // Convert field
+    // Convert nodal fields
+}
+
+
```
cyrush: - [x] remove these before merge

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1043,6 +1043,101 @@ convert_coordset_to_explicit(const std::string &base_type,
...
+                             conduit::Node & to,
+                             const std::string & key,
+                             const std::string & new_name,
+                             const std::string & new_sibling)
+{
+    to[key].set(from[key]);
+    conduit::Node & n = to[key];
+    const std::string from_name = n.child(0).name();
+    n.rename_child(from_name, new_name);
+    n[new_sibling].set(n[new_name]);
```
cyrush: Despite the name - I am having a hard time parsing this. 

Do we need to pass from & to and the child key?
Why not just pass those nodes like:
`copy_node_rename_first_child(src[key],dest[key], new_name, new_sibling)`

I also don't understand the context of `new_sibling`? 

I see the rename, but then `n[new_sibling]` is a copy of the first child after it was renamed?

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1043,6 +1043,101 @@ convert_coordset_to_explicit(const std::string &base_type,
...
+                               conduit::Node &dest)
+{
+    dest.reset();
+    std::string coord_type = coordset["type"].as_string();
+    dest["type"].set(coord_type);
+
+    if (coord_type == "uniform")
+    {
+        copy_node_rename_first_child(coordset, dest, "dims", "j", "i");
+        dest["dims/i"].set(1);
```
cyrush:
```
copy_node_rename_first_child(coordset, dest, "dims", "j", "i");
```
This will copy `coordset["dims"]` into `dest["dims"]`.
Then rename `dest["dims"].child(0)` to be named `j` or `dest["dims/j"]`
Then copy `dest["dims/j"]` so it is also `dest["dims/i"]`
And then you reset:
`dest["dims/i"].set(1);`

Why have the sibling copy if we are going to reset it in these cases?

cyrush (commented):
New methods are on the right track 👍 

I a few questions about implementation.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -1374,6 +1374,11 @@ braid_structured(index_t npts_x,
                                  npts_z,
                                  res["coordsets/coords"]);
 
+    if (npts_y < 1 && npts_z < 1)
+    {
+        res["coordsets/coords/type"] = "rectilinear";
+    }
+
```
cyrush: Not sure I understand this change.

braid_structured aims to create implicit logical topology with on an explicit coordset.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1084,12 +1084,12 @@ convert_oneD_coordset_to_strip(const conduit::Node &coordset,
     if (coord_type == "uniform")
     {
         copy_node_rename_first_child(coordset, dest, "dims", "j", "i");
-        dest["dims/i"].set(1);
+        dest["dims/i"] = 1;
```
cyrush, line 1057: I am still confused about this, is:

```
copy_node_rename_first_child(coordset, dest, "dims", "j", "i");
dest["dims/i"] = 1;
```

equiv to:

```
dest["dims/j"] = coordset["dims/i"];
dest["dims/i"] = 1;
```

?

If so, the helper is obfuscating this, I suggest we just do the direct operations. 

Also, for the result coords for the strip mesh, is there a reason we want the first logical coord to be flat vs instead of the second?

Why not use:
```
dest["dims/j"] = coordset["dims/i"];
dest["dims/j"] = 1;
```

It seems strange to swap the coords instead of just extending?

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1101,6 +1101,7 @@ convert_oneD_coordset_to_strip(const conduit::Node &coordset,
     else
     {
         copy_node_rename_first_child(coordset, dest, "values", "y", "x");
+        dest["values/x"].reset();
```
cyrush: `reset` isn't needed b/c set will take care of it if needed.

cyrush (commented):
@agcapps a few more questions.  Let me know if you want to meet to discuss.

src/libs/blueprint/conduit_blueprint_mesh.cpp
agcapps: The helper is probably due to my inexperience.  I was trying to do:

```
dest["dims/j"] = coordset["dims/i"];
dest["dims/i"] = 1;   // being sure that dest["dims/i"] has the same data type as dest["dims/j"]
```

If the extra step isn't needed, I'll happily do it that way.

There's a library that expects the strip mesh to have y as the independent variable, and x to be [0, 1].

src/libs/blueprint/conduit_blueprint_mesh.cpp
agcapps: Before I added the call to `reset()`, I was seeing `dest["values/x"]` hold the same values and dimension as `dest["values/y"]`, except elements 0 and 1 were 0.0 and 1.0.  Now much has changed, and after trying it without the call to `reset()` I see that the expected two values are there.

Removed the call to `reset()`.

src/libs/blueprint/conduit_blueprint_mesh.cpp
agcapps: Removed the helper function; using the pattern you described above.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
agcapps: woops.  I mixed up rectilinear and structured.  Just trying to keep the amount of allocation to a minimum.

src/libs/blueprint/conduit_blueprint_mesh.cpp
agcapps: Well, it should go somewhere.  I'll remove it.

agcapps:
@cyrush , any more thoughts on this PR?

cyrush:
Yes, I have been pondering.

We can chat at our meeting today.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1043,6 +1043,74 @@ convert_coordset_to_explicit(const std::string &base_type,
...
+        {
+            dest["spacing/dy"] = coordset["spacing/dx"];
+            dest["spacing/dx"] = 1.;
+        }
+    }
+    else
+    {
+        dest["values/y"] = coordset["values/x"];
+        // dest["values/x"].reset();
+        dest["values/x"].set(DataType::float64(2));
```
cyrush, line 1076: coordset["values/x"].to_float64_array(dest["values/y"])

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1043,6 +1043,74 @@ convert_coordset_to_explicit(const std::string &base_type,
...
+
+        if (coordset.has_child("spacing"))
+        {
+            dest["spacing/dy"] = coordset["spacing/dx"];
+            dest["spacing/dx"] = 1.;
+        }
+    }
+    else
+    {
+        dest["values/y"] = coordset["values/x"];
```
cyrush:
```suggestion
coordset["values/x"].to_float64_array(dest["values/y"]);
```

cyrush (approved):
Look good,thanks for the iteration on this.

PR #1020, opened 2022-09-30
Use int64_accessor instead of creating a temp Node and using to_int64().

agcapps:
Performance bugfix: previous code instantiated a Node, set it, and used to_int64().  On one particular test mesh, using the int64_accessor cuts time in `conduit::blueprint::mesh::utils::topology::unstructured::points()` by a factor of about 4.


cyrush (approved):
+1

PR #1025, opened 2022-10-24
add Node move and swap methods

cyrush:
Idea from discussion with @gunney1

Adds `Node::move(Node &)` and `Node::swap(Node &)` methods, that help build trees efficiently from other nodes. 

Includes c, fortran, and python bindings.


cyrush:
@gunney1 and @agcapps can you give this a quick review?

agcapps (approved):
Thanks for this, @cyrush

PR #1027, opened 2022-10-25
Fixes to helper functions (correctness and performance)

agcapps:


src/libs/blueprint/conduit_blueprint_o2mrelation.cpp
```diff
@@ -263,16 +263,30 @@ bool generate_offsets(conduit::Node &n,
...
     {
+        // int64_accessor sizes = n["sizes"].as_int64_accessor();
+        // n["offsets"].reset();
+        // n["offsets"].set(DataType::int64(sizes.number_of_elements()));
+        // int64_array offsets_vals = n["offsets"].value();
+
+        // for (index_t o = 0; o < sizes.number_of_elements(); ++o)
+        // {
+        //    offsets_vals[o] = (o > 0) ? offsets_vals[0-1] + sizes[o-1] : 0;
+        // }
```
agcapps: This fills `n["offsets"]` with nonsense low integers.  @cyrush , what am I doing wrong here?

src/libs/blueprint/conduit_blueprint_o2mrelation.cpp
```diff
@@ -263,16 +263,30 @@ bool generate_offsets(conduit::Node &n,
...
         conduit::Node &sizes = n["sizes"];
         conduit::Node &offsets = n["offsets"];
 
         conduit::Node temp;
         std::vector<int64> offset_array(sizes.dtype().number_of_elements());
         for(index_t o = 0; o < (index_t)offset_array.size(); o++)
         {
-            temp.set_external(conduit::DataType(sizes.dtype().id(), 1),
-                const_cast<void*>(sizes.element_ptr(o)));
```
agcapps, line 274: The correctness error was that the offset was `o` and should be `o-1`.

src/libs/blueprint/conduit_blueprint_o2mrelation.cpp
cyrush: `offsets_vals[0-1]` instead intended offsets_vals[`o-1`]  ?

maybe we should use `o_idx` instead `o`

agcapps: AArgh.  Yes, that was the problem.  Using the better name helps.

src/libs/blueprint/conduit_blueprint_o2mrelation.cpp
```diff
@@ -263,21 +263,15 @@ bool generate_offsets(conduit::Node &n,
     }
     else
     {
-        conduit::Node &sizes = n["sizes"];
-        conduit::Node &offsets = n["offsets"];
+        int64_accessor sizes = n["sizes"].as_int64_accessor();
```
cyrush: lets move to `index_t` for these instead of `int64`

src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
```diff
@@ -2014,21 +2014,19 @@ topology::unstructured::generate_offsets(const Node &n,
...
 
         std::vector<index_t> shape_array;
         index_t ei = 0;
         index_t es = 0;
-        while(ei < topo_elem_size.dtype().number_of_elements())
+        while(ei < topo_elem_size.number_of_elements())
         {
-            const Node index_node(int_dtype,
-                const_cast<void*>(topo_elem_size.element_ptr(ei)), true);
             shape_array.push_back(es);
```
cyrush: we could do a counting pass to size the shape array output and avoid using vector w/ `push_back`, what do you think?

src/tests/blueprint/t_blueprint_o2mrelation_examples.cpp
```diff
@@ -187,7 +190,7 @@ TEST(conduit_blueprint_o2mrelation_examples, o2mrelation_generate_offsets)
     n.remove("offsets");
     EXPECT_TRUE(blueprint::o2mrelation::generate_offsets(n, info));
     EXPECT_TRUE(blueprint::o2mrelation::verify(n,info));
-    EXPECT_FALSE(ref.diff(n, info));
+    EXPECT_FALSE(ref.diff(n, info, epsilon, relaxint));
```
agcapps, line 193: This was returning true because `n` had sizes of type uint32, not index_t == int64.  To quiet the failures, I set relaxint==true.

src/tests/blueprint/t_blueprint_mesh_generate.cpp
```diff
@@ -1404,6 +1404,8 @@ TEST(conduit_blueprint_generate_unstructured, generate_sides)
             EXPECT_TRUE(o2mrelation::verify(map_node, info));
             for(const std::string &o2m_path : O2M_PATHS)
             {
+                // Do we need the type checks?  If sizes are always generated
+                // as index_t, there will sometimes be type mismatches.
                 EXPECT_TRUE(map_node.has_child(o2m_path));
                 EXPECT_EQ(map_node[o2m_path].dtype().id(), grid_conn.dtype().id());
```
agcapps: This is failing now that o2m sizes get generated as index_t.  Keep, or comment out?

cyrush: I don't think we need the data type to match moving forward.

src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
```diff
@@ -2014,37 +2014,35 @@ topology::unstructured::generate_offsets(const Node &n,
...
+        index_t elem_count = topo_elem_size.number_of_elements();
+        std::vector<index_t> shape_array(elem_count, 0);
         index_t ei = 0;
         index_t es = 0;
-        while(ei < topo_elem_size.dtype().number_of_elements())
+        while(ei < elem_count)
         {
-            const Node index_node(int_dtype,
-                const_cast<void*>(topo_elem_size.element_ptr(ei)), true);
-            shape_array.push_back(es);
```
agcapps, line 2030: Got rid of the calls to  `push_back()`

agcapps:
@cyrush , any more comments?

src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
```diff
@@ -2014,37 +2014,35 @@ topology::unstructured::generate_offsets(const Node &n,
...
+        index_t_accessor topo_elem_size = n["elements/sizes"].as_index_t_accessor();
+        index_t_accessor topo_subelem_size = n["subelements/sizes"].as_index_t_accessor();
 
         Node elem_node;
         Node subelem_node;
 
-        std::vector<index_t> shape_array;
+        index_t elem_count = topo_elem_size.number_of_elements();
+        std::vector<index_t> shape_array(elem_count, 0);
         index_t ei = 0;
```
cyrush: We can avoid the std::vector all together (even with the std::vector there is still index_t type conversoin, so best to just use index_t):

```suggestion
        dest_elem_off.set(DataType::index_t(elem_count);
        index_t_array shape_array  = dest_elem_off.value();
        index_t ei = 0;
```

cyrush (approved):
Looks good one final suggestion (avoid std::vector alloc all together)

cyrush:
@agcapps we also need a similar change for the sub elements

cyrush:
@agcapps  refactoring to fix current dependence on side effects is a bit of a rabbit hole, we can tackle that in a follow on PR. 

Let's get your changes in -- Can you "Squash Merge" -- thanks!

PR #1034, opened 2022-11-04
Generated one-D fields

agcapps:
When generating a 2D strip topology, copy the element fields from the old to the new.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -1080,36 +1080,6 @@ convert_oneD_coordset_to_strip(const conduit::Node &coordset,
 }
 
 
-//-------------------------------------------------------------------------
-void
-convert_oneD_topo_to_strip(const conduit::Node &topo,
-                           const std::string &csname,
-                           conduit::Node &dest)
-{
-    dest.reset();
-    dest["type"] = topo["type"].as_string();
-    dest["coordset"] = csname;
-
-    if (topo.has_child("elements"))
-    {
-        const Node & topoelts = topo["elements"];
-        Node & destelts = dest["elements"];
-
-        if (topoelts.has_child("origin"))
-        {
-            destelts["origin/i"] = 0.;
-            destelts["origin/j"] = topoelts["origin/i"];
-        }
-
-        if (topoelts.has_child("dims"))
-        {
-            destelts["dims/i"] = 1;
-            destelts["dims/j"] = topoelts["dims/i"];
-        }
-    }
-}
-
-
 // TODO(JRC): For all of the following topology conversion functions, it's
 // possible if the user validates the topology in isolation that it can be
 // good and yet the conversion will fail due to an invalid reference coordset.
@@ -2191,6 +2161,8 @@ mesh::generate_strip(const conduit::Node& topo,
                      conduit::Node& fields_dest,
                      const conduit::Node& options)
 {
+    const std::string topo_name = topo.name();
+    const std::string topo_dest_name = topo_dest.name();
     std::string field_prefix = "";
     std::vector<std::string> field_names;
     const Node& fields_src = (*(topo.parent()->parent()))["fields"];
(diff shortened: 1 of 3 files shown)


src/tests/blueprint/t_blueprint_mesh_examples.cpp
```diff
@@ -1292,57 +1292,41 @@ TEST(conduit_blueprint_mesh_examples, polystar)
 TEST(conduit_blueprint_mesh_examples, oneDtostrip)
 {
     {
-        std::cout << "starting rectilinear" << std::endl;
```
agcapps, line 1295: Clean up construction debris

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1071,11 +1071,11 @@ convert_oneD_coordset_to_strip(const conduit::Node &coordset,
     }
     else
     {
-        coordset["values/x"].to_float64_array(dest["values/y"]);
         dest["values/x"].set(DataType::float64(2));
         double *x_vals = dest["values/x"].value();
         x_vals[0] = 0.;
         x_vals[1] = 1.;
+        coordset["values/x"].to_float64_array(dest["values/y"]);
```
agcapps, line 1078: Fix ordering (discussed in previous PR, but got left out).  See also line 1091.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -5903,6 +5904,60 @@ mesh::field::verify(const Node &field,
...
+            }
+        }
+    }
+
+    NodeConstIterator newfields_it = newfields.children();
+    while (newfields_it.has_next())
+    {
+        const Node& newfield = newfields_it.next();
+        fields[newfields_it.name()] = newfield;
+    }
```
agcapps, line 6064: I used the pattern
- iterate over source Node's children, generating a new from qualifying Nodes
- add the generated Nodes back to the target, which happens to be the source
because the source and target are the same nodes (`mesh["fields"]`).

If there's a better pattern (like hypothetical `Node::MergeFrom(const Node& other)`), then we can use that.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -5903,6 +5904,60 @@ mesh::field::verify(const Node &field,
...
+    while (fields_it.has_next())
+    {
+        const Node& field = fields_it.next();
+
+        if (field["topology"].as_string() == toponame)
+        {
+            // TODO Something useful with grid functions, when needed by users.
+            if (field.has_child("association"))
+            {
+                if (field["association"].as_string() == "element")
```
agcapps, line 6044: If we want, we can put the body of this function into a helper function.

agcapps:
@cyrush , after merging in my previous PR #1018 I realized that it didn't do anything to the fields.  This PR adds fields (somewhat important).

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -5903,6 +5904,60 @@ mesh::field::verify(const Node &field,
...
+                    if (field.has_child("volume_dependent"))
+                        newfield["volume_dependent"] = field["volume_dependent"].as_string();
+                    if (field.has_child("matset"))
+                    {
+                        newfield["matset"] = field["matset"].as_string();
+                        newfield["matset_values"].set_external(field["matset_values"]);
+                    }
+                    if (field.has_child("values"))
+                    {
+                        newfield["values"].set_external(field["values"]);
```
agcapps: zero-copy when we can (also matset_values)

cyrush:
For the case we need to propagate fields,  the interface should follow the pattern used in generate_sides:

```
mesh::topology::unstructured::generate_sides(const conduit::Node &topo_src,
                                                                            conduit::Node &topo_dest,
                                                                            conduit::Node &coordset_dest,
                                                                           conduit::Node &fields_dest)
```
https://github.com/LLNL/conduit/blob/74b2fdf9dc40d5088fbab0d54aa07dbad598ec8c/src/libs/blueprint/conduit_blueprint_mesh.hpp#L560

cyrush:
You can identify all the fields from the src_topo, then make choices about which ones to propagate.

cyrush:
Here is where that logic begins in gen_sides:

https://github.com/LLNL/conduit/blob/74b2fdf9dc40d5088fbab0d54aa07dbad598ec8c/src/libs/blueprint/conduit_blueprint_mesh.cpp#L4927

cyrush:
generate_sides also takes options that allow you to say which fields and provide a field prefix (in case the new fields/topo live in the same tree). We should also follow that.

cyrush:
One more thing: We may have an issue with invaliding the iterator like we discussed on Friday.

agcapps:
> One more thing: We may have an issue with invaliding the iterator like we discussed on Friday.

We're doing the fields by making a list of strings (the source names), then building each new derived field.  This may avoid invalidating iterators.

cyrush (approved):

PR #1037, opened 2022-11-07
generate_offsets without side effects

cyrush:
* add generate_offsets_inline to serve the case where we want to create them in the tree
* refactor the polyhedral case for generate_offsets to clearly identify the cases where we are breaking const rules
* simplify generate_offsets logic using accessors
* refactored `mesh::topology::logical_dims to use direct calc instead of relying on generate offsets

* fixed const-cast related crime for element and subelement offset gen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 #### Relay
 - Added any source, any tag variants of mpi receive functions: `recv`, `recv_using_schema`, and `irecv`.
 - Added subpath support for `relay::io::{save,load,save_merged,load_merged}` for basic protocols (json, yaml, etc).
+- Added `generate_offsets_inline(Node &)` for cases where we want topology offsets created in existing tree.
 
 ### Changed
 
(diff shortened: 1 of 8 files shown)


src/tests/blueprint/t_blueprint_mesh_generate.cpp
```diff
@@ -1269,6 +1269,10 @@ TEST(conduit_blueprint_generate_unstructured, generate_faces)
 
         Node info;
         EXPECT_TRUE(mesh::topology::unstructured::verify(face_topo, info));
+        
+        face_topo.print();
+        info.print();
+
```
cyrush: - [x] remove this

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 #### Relay
 - Added any source, any tag variants of mpi receive functions: `recv`, `recv_using_schema`, and `irecv`.
 - Added subpath support for `relay::io::{save,load,save_merged,load_merged}` for basic protocols (json, yaml, etc).
+- Added `generate_offsets_inline(Node &)` for cases where we want topology offsets created in existing tree.
```
agcapps: Shouldn't this be under Blueprint?

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 #### Relay
 - Added any source, any tag variants of mpi receive functions: `recv`, `recv_using_schema`, and `irecv`.
 - Added subpath support for `relay::io::{save,load,save_merged,load_merged}` for basic protocols (json, yaml, etc).
+- Added `generate_offsets_inline(Node &)` for cases where we want topology offsets created in an existing tree.
```
agcapps: Shouldn't this change be under Blueprint?

cyrush:
@agcapps  I believe I have remediated the const-cast crime :-)

agcapps:
Thanks for this, @cyrush .  I went over the PR a few times and it looks reasonable.

agcapps (approved):

PR #1038, opened 2022-11-08
Generate 1D strip matsets

agcapps:
Similar to fields, new matsets must be generated that point to the new topologies.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -2144,11 +2144,15 @@ mesh::generate_strip(conduit::Node &mesh,
     const Node& src_topo = mesh["topologies"][src_topo_name];
     const Node& src_coordset = mesh["coordsets"][src_topo["coordset"].as_string()];
     Node dst_topo, dst_coordset;
+    std::map<std::string, std::string> matset_names;
 
     mesh::coordset::generate_strip(src_coordset, dst_coordset);
     mesh::topology::generate_strip(src_topo, dst_topo_name, dst_topo);
-    mesh::field::generate_strip(mesh["fields"], src_topo_name, dst_topo_name);
-    mesh::matset::generate_strip(mesh["matsets"], src_topo_name, dst_topo_name);
+    mesh::field::generate_strip(mesh["fields"], src_topo_name, dst_topo_name, matset_names);
+    if (matset_names.size() > 0)
+    {
+        mesh::matset::generate_strip(mesh["matsets"], dst_topo_name, matset_names);
+    }
 
     mesh["topologies"][dst_topo_name] = dst_topo;
     mesh["coordsets"][dst_topo_name] = dst_coordset;
@@ -2164,8 +2168,9 @@ mesh::generate_strip(const conduit::Node& topo,
 {
     const std::string topo_name = topo.name();
     const std::string topo_dest_name = topo_dest.name();
-    std::string field_prefix = "";
+    std::string field_prefix = "", matset_prefix = "";
     std::vector<std::string> field_names;
+    std::map<std::string, std::string> matset_names;
     const Node& fields_src = (*(topo.parent()->parent()))["fields"];
     const Node& coordset_src = (*(topo.parent()->parent()))["coordsets/" + topo["coordset"].as_string()];
 
@@ -2175,6 +2180,7 @@ mesh::generate_strip(const conduit::Node& topo,
         if (options["field_prefix"].dtype().is_string())
         {
             field_prefix = options["field_prefix"].as_string();
+            matset_prefix = field_prefix;
         }
         else
         {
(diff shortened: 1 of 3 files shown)


src/libs/blueprint/conduit_blueprint_mesh.hpp
```diff
@@ -642,6 +650,11 @@ namespace topology
 //-----------------------------------------------------------------------------
 namespace matset
 {
+    //-------------------------------------------------------------------------
+    void CONDUIT_BLUEPRINT_API generate_strip(conduit::Node& fields,
```
cyrush: This should be a private helper b/c it really can't be used outside the context of other generate_strip methods.

agcapps:
Need to add tests

agcapps:
Putting up this PR because it turns out we need matsets right away.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2255,6 +2256,97 @@ mesh::generate_strip(const conduit::Node& topo,
...
+            matset_prefix = options["matset_prefix"].as_string();
+        }
+        else
+        {
+            CONDUIT_ERROR("matset_prefix must be a string.");
+        }
+    }
+
+    // check for target matset names
+    if (options.has_child("matset_names"))
```
cyrush: Note: matset names could be inferred from the `field_names` options (or all fields).
I expect thats how folks would think to use it ("I want these fields, make it so")

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2255,6 +2256,97 @@ mesh::generate_strip(const conduit::Node& topo,
...
+    {
+        if (!matsets_src.has_child(matset_names[i]))
+        {
+            CONDUIT_ERROR("matset " + matset_names[i] + " not found in target.");
+        }
+    }
+
+    // generate new coordset, topology, and fields
+    mesh::generate_strip(topo, topo_dest, coords_dest, fields_dest, options);
+
```
cyrush: For any new field that has a matset, we need to rewire:

`fields/{name}/matset = newmatsetname`

src/libs/blueprint/conduit_blueprint_mesh.hpp
cyrush: Further review: don't think we need this this guy?
(Not being called)

Sorry if I am commenting while things are evolving.

agcapps:
Matsets refer to a topology.  All fields refer to a topology, and some fields refer to a matset.  

Here's my question, @cyrush :  Will we ever have a matset that does not get referred to by a field?  Do we care about those matsets?  It's much easier if we can ignore "orphan" matsets that refer to the source topo but aren't referred to by any field.  In that case, as you suggested on line 2287 of conduit_blueprint_mesh.cpp, we can just use options/field_prefix to name the new matsets.

I guess we could use options/field_prefix to name all the new matsets that get created for the new topology.  Thoughts?

cyrush:
For the intended use case, I believe folks will care first and foremost about mapping fields ("I need these fields")
and a standalone matset (no related fields) would only pop up for testing. 

To give full control, let's provide a `matsets_prefix` option that acts like `fields_prefix`. (that is optional, but when specifi is used for any new matsets created)

agcapps:
Another question:  A common case (that we are showing in the example) is for the user to create the new topo and all dependent fields, etc. in the same mesh.  In this case, unless the user specifies a fields_prefix, there will be overwritten fields.  Do we care?  Should we make a default fields_prefix?

I mean, perhaps the user meant to overwrite the field, though I really doubt that.

cyrush:
I think overwriting fields (and other components) will be unintended in most all cases, and could be a source of pain.

We should error if folks try to override any field, or matset.

We could also guard the topology or coordset, but those checks will require different implementation.

agcapps:
We need to exercise this on a test.  This opens up a huge can of worms.  Only one of the meshes produced by the functions in conduit_blueprint_examples.cpp has a matset, and that one isn't linked to a field.  

We could add matsets to basic(); I think that would definitely be scope creep.

We could add a top level (public) function to conduit_blueprint_mesh_examples.hpp/cpp, named something like `add_mixed_field()`, that only operates on rectilinear or structured meshes.

And then there is the fact that we have several different forms of matset.  I would propose to generate the form that the user needs, then later add the others.

cyrush:
We do need test data to exercise this -- but we don't need a fully fledged example.

Since we are doing a simple 1D dataset as input, I recommend we create a matset and field.
Easiest style matset and corresponding field data. 

We effectively copy the matset so I am not worried about testing all the combos of the matset internals.

agcapps:
> We do need test data to exercise this -- but we don't need a fully fledged example.

Agreed!

> Since we are doing a simple 1D dataset as input, I recommend we create a matset and field. Easiest style matset and corresponding field data.

OK, so I'll build the matset in the test.

> We effectively copy the matset so I am not worried about testing all the combos of the matset internals.

Yes, I'll build the kind of matset that we're interested in.

cyrush:
Good plan.

src/tests/blueprint/t_blueprint_mesh_examples.cpp
```diff
@@ -1288,12 +1288,46 @@ TEST(conduit_blueprint_mesh_examples, polystar)
...
+
+    float64 dval = 1. / nele;
+
+    for (int i = 0; i < nele; ++i)
+    {
+        f1ms001vals[i] = 1. - i*dval;
+        f1vals[i] = 1. - i * dval;
+        ms1vfracs[i] = 1.0;
+    }
+}
```
cyrush, line 1321: looks good!

cyrush (approved):

PR #1047, opened 2022-11-29
initial caliper annotations api

cyrush:
resolves #1045


cyrush:
@agcapps initial pass at functional caliper setup w/ conduit annotation macros. 

we can expand as needed to wrap more caliper functionality, but I think the basics (region marking, function can scope marking) will take us a long way. 

For our tests we can use `conduit::annotation::initialize()` to capture info, host codes will likley have their own caliper setup.

Let me know what you think. I hope to merge soon and we can start using asap.

src/libs/conduit/conduit_utils.hpp
```diff
@@ -21,12 +21,16 @@
...
 //-----------------------------------------------------------------------------
 // -- conduit includes --
 //-----------------------------------------------------------------------------
 #include "conduit_core.hpp"
 
+//-----------------------------------------------------------------------------
+#if defined(CONDUIT_USE_CALIPER)
+#include <caliper/cali.h>
+#endif
+
```
agcapps: Is conduit_utils.hpp a user-facing header file?  Will this let caliper leak out into a host code through Conduit?

cyrush: that one I need to remove, good catch!

cyrush:
- [x] add caliper cmake exports and config.mk logic

agcapps:
Thanks for this, Cyrus.  It looks good to me.  From my beginner's knowledge of Caliper, it does look like mark begin, end, function, and scope are reasonable things to expose.

agcapps (commented):
Thanks, Cyrus.

PR #1052, opened 2022-12-12
Pinning conduit to output files compatible with hdf5@1.8

rblake-llnl:
Added extra checks to conduit to always output files compatible with hdf5@1.8

### DIFF
--- a/src/libs/relay/conduit_relay_io_hdf5.cpp
+++ b/src/libs/relay/conduit_relay_io_hdf5.cpp
@@ -2950,6 +2950,17 @@ create_hdf5_file_access_plist()
                                  << "property list " << h5_fa_props);
 
     }
+    else if (major_num == 1 && ((minor_num == 10 && release_num >= 2) || minor_num > 10))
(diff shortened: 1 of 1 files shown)


cyrush:
@rblake-llnl  Last week, I did hear that some folks successfully used libver to pin files to 1.8 with a newer hdf5 build (1.12) and the silo files they created were readable by VisIt.

So that's promising news. 

Before I merge - I want to double check VisIt using 1.8 can read conduit flavored hdf5 created with 1.10, 1.12, and 1.13 b/c @markcmiller86 suggested this might depend on what HDF5 calls are used.  I'll get this PR in after that investigation if all looks good. 

I may also make the libver an option, but that can be a follow on.

cyrush:
@rblake-llnl  I tested builds with 1.10 and 1.12 with this on and I was able to read generated the files in VisIt fine. 

I am merging and I will do a follow up PR to add an option to control this just in case other folks don't want default of 1.8.

Thanks!

PR #1053, opened 2022-12-15
Create coordset using computed extent

agcapps:
Create coordset using computed extent, after padding user's requested data extent.

Unlike other types of mesh, the strided_structured mesh has padding on the sides.  This fix makes the coordset to the size computed by adding any padding to the logical size requested by the user.


src/libs/blueprint/conduit_blueprint_mesh_examples.hpp
```diff
@@ -72,14 +74,104 @@ namespace examples
...
+    /// one element in each dimension than arrays that store fields over
+    /// elements.  In strided structured meshes, the array dimensions of vertex
+    /// and element fields are decoupled.  In this example, vertex fields will
+    /// use a 7x6 array, not an 8x7 array as required by other Blueprint meshes.
+    /// Since there are 3x2 mesh elements, there will be 4x3 vertex field
+    /// values.  To match the element fields' padding as closely as possible,
+    /// the vertex field values are located at origin (2, 2).
+    ///
+    /// The diagram below shows the mesh.  Note that the origin is at the top
+    /// left.  For elements, a space indicates a padding element and a `d`
```
cyrush: Thought: 
The logical origin could be different with respect to how it is embedded in space. Not sure if saying top left will confuse folks?

agcapps: Perhaps rephrase to "Note that the diagram origin is at the top left."  What do you think?  I need something to anchor all the coordinates in the previous paragraph, not to mention orient the vertices that have no data (the blanks).

cyrush: That sounds good!

cyrush:
For debugging and so we can verify things match expectations, I added a point topology that uses the underlying coordinates w/o striding.  Since the field values also match the cardinality of this coordset, we can also plot the fields on this topology as well. 

Here is an example:
Screen Shot 2022-12-19 at 11 59 07 AM


Here the meshes plot are our intended strided results, and the points are the original points.

cyrush:
I want to make sure we have the correct logic for the `pad` cases.

Screen Shot 2022-12-19 at 12 11 13 PM
Screen Shot 2022-12-19 at 12 11 49 PM

cyrush:
@agcapps  I think the out of the box case is good but I still have questions about pad case.
I expect we will have to iterate more both in conduit and VisIt. 
I am merging so we can get the current improvements in the next Conduit release.

PR #1060, opened 2023-01-01
Fix for older Apple Clang - LLVM version 8.1.0 (clang-802.0.42)

BradWhitlock:
I ran into a compilation problem using clang on an old Mac. This commit fixes that compilation problem.


cyrush:
@BradWhitlock Q: should we still aim to get this fix in?

cyrush:
Closing since this is for an older set of compilers and very old mac.

PR #1072, opened 2023-01-30
Fix a triangle in TOPO_PYRAMID_EMBEDDING.

BradWhitlock:
I was suing the embedding shapes to produce a list of unique edge ids for an element and I found that the algorithm made an extra edge for pyramids. The 2nd triangle produced to split the pyramid base overlapped the 1st triangle. It seems that the 2 triangles should instead cover the base quad shape. I changed the pyramid so it does this and my other algorithm started making the right number of unique edges.


cyrush:
thanks @BradWhitlock !

Issue #908, opened 2022-01-24
O2M and ND Relation Map Interface


cyrush:
Strawman for new index-based interface for NDRelations and O2MRelations:

```
nd_rel = NDRel(shape)
nd_rel = NDRel(shape, offsets)
nd_rel = NDRel(shape, strides)
nd_rel = NDRel(shape, offsets, strides)

nd_rel.shape(0) ==> shape[0]
nd_rel.offset(0) ==> offsets[0]
nl_rel.stride(0) ==> strides[0]
nd_rel.map(i,j,k, ...) ==> index used with values arrays

// full walk (3d):
for( i = 0; i < nd_rel.shape(0); i++)
  for( j = 0; j < nd_rel.shape(1); j++)
     for (k = 0; k < nd_rel.shape(2); k++)
           nd_rel.map(i,j,k) # applied to values arrays

o2m_rel = O2MRel(sizes)
o2m_rel = O2MRel(sizes, offsets )
o2m_rel = O2MRel(sizes, indices )
o2m_rel = O2MRel(sizes, offsets, indices)

o2m_rel.size() ==> number of "ones"
o2m_rel.size(i) ==> number of "many" for entry `i`
o2m_rel.offset(i) ==> start index of "many" for entry `i`
o2m_rel.map(i,j) ==>  applies( indices[offsets[i]] + j ), to calc  index used with values arrays

// full walk:
for( i = 0; i < o2m_rel.size(); i++)
   for (j = 0; j < o2m_rel.size(i); j++)
         o2m_rel.map(i,j) # applied to values arrays
```

For O2M - this style can be GPU friendly, vs the existing O2MIterator.

For ND - we expect an ND index based interface will be desired over an iterator.

cyrush:
related to https://github.com/LLNL/conduit/issues/74

cyrush:
Updated strawman:

```
nd_idx = NDIndex(const Node &)
nd_idx = NDIndex(shape)
nd_idx = NDIndex(shape, offsets)
nd_idx = NDIndex(shape, strides)
nd_idx = NDIndex(shape, offsets, strides)

nd_idx.shape(0) ==> shape[0]
nd_idx.offset(0) ==> offsets[0]
nd_idx.stride(0) ==> strides[0]
nd_idx.index(i,j,k, ...) ==> index used with values arrays

// full walk (3d):
for( i = 0; i < nd_idx.shape(0); i++)
  for( j = 0; j < nd_idx.shape(1); j++)
[...]

PR #1077, opened 2023-02-07
O2MIndex

agcapps:
Feature: add a lightweight class to compute the flatindex to elements of O2M relations.
Feature: add another lightweight class to compute the flatindex to elements in an ND array.

Addresses #908


src/libs/blueprint/conduit_blueprint_o2mrelation_index.cpp
```diff
@@ -0,0 +1,220 @@
...
+//---------------------------------------------------------------------------//
+O2MIndex::O2MIndex(const Node *node)
+: m_sizes_node(NULL),
+  m_indices_node(NULL),
+  m_offsets_node(NULL)
+{
+    std::vector<std::string> paths = conduit::blueprint::o2mrelation::data_paths(*node);
+
+    if(node->has_child("sizes"))
+        m_sizes_node = node->fetch_ptr("sizes");
```
cyrush: Q: Should we build the accessors in the constructor, so they aren't built on the fly during the lookup calls?

cyrush: https://github.com/LLNL/conduit/pull/1080 this might help.

agcapps: Done

agcapps:
@cyrush , what's with all the pip setup failures?  Yes, there's a log I can read, but I don't understand what they mean.

cyrush:
There is an error hiding in the warnings:
```
2023-03-17T23:08:39.8664824Z   /home/runner/work/conduit/conduit/src/libs/blueprint/conduit_blueprint_mesh_utils.cpp: At global scope:
2023-03-17T23:08:39.8667432Z   /home/runner/work/conduit/conduit/src/libs/blueprint/conduit_blueprint_mesh_utils.cpp:573:1: error: declaration of ‘void conduit::blueprint::mesh::utils::NDIndex::swap(conduit::blueprint::mesh::utils::NDIndex&)’ has a different exception specifier
2023-03-17T23:08:39.8668527Z     573 | NDIndex::swap(NDIndex& idx)
2023-03-17T23:08:39.8669704Z         | ^~~~~~~
2023-03-17T23:08:39.8671361Z   In file included from /home/runner/work/conduit/conduit/src/libs/blueprint/conduit_blueprint_mesh_utils.cpp:30:
2023-03-17T23:08:39.8674384Z   /home/runner/work/conduit/conduit/src/libs/blueprint/conduit_blueprint_mesh_utils.hpp:349:10: note: from previous declaration ‘void conduit::blueprint::mesh::utils::NDIndex::swap(conduit::blueprint::mesh::utils::NDIndex&) throw ()’
2023-03-17T23:08:39.8675170Z     349 |     void swap(NDIndex & idx) throw();
2023-03-17T23:08:39.8676224Z         |          ^~~~
```

We should also look into resolving the `deprecated-copy` warnings to improve signal to noise.

agcapps:
> There is an error hiding in the warnings:
> 
> ```
> 2023-03-17T23:08:39.8664824Z   /home/runner/work/conduit/conduit/src/libs/blueprint/conduit_blueprint_mesh_utils.cpp: At global scope:
> 2023-03-17T23:08:39.8667432Z   /home/runner/work/conduit/conduit/src/libs/blueprint/conduit_blueprint_mesh_utils.cpp:573:1: error: declaration ... has a different exception specifier
> 2023-03-17T23:08:39.8668527Z     573 | NDIndex::swap(NDIndex& idx)
> 2023-03-17T23:08:39.8669704Z         | ^~~~~~~
> 2023-03-17T23:08:39.8675170Z     349 |     void swap(NDIndex & idx) throw();
> 2023-03-17T23:08:39.8676224Z         |          ^~~~
> ```
> 
> We should also look into resolving the `deprecated-copy` warnings to improve signal to noise.

Ah.  thank you.

src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
```diff
@@ -342,6 +342,326 @@ find_domain_id(const Node &node)
     return domain_id;
 }
 
+
+//-----------------------------------------------------------------------------
+NDIndex::NDIndex(const Node& idx) :
+    NDIndex(&idx)
```
cyrush, line 348: Q about design of NDIndex -- What cases do we need to allocate nodes for?

Can re-rework this to rely on fully external data?

We want these to be as light weight as possible.

src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
```diff
@@ -342,6 +342,116 @@ find_domain_id(const Node &node)
...
+    }
+    return *this;
+}
+
+void NDIndex::info(Node& res) const
+{
+    index_t dim = shape();
+    res["shape"].set(DataType::index_t(dim));
+    res["offset"].set(DataType::index_t(dim));
+    res["stride"].set(DataType::index_t(dim));
```
cyrush, line 422: looks like it would be nice for use to off a Node::set via data_accessor :-)

src/libs/blueprint/conduit_blueprint_o2mrelation_index.cpp
cyrush: Following our recent changes to NDIndex - how about we avoid the node pointers here as well and only use the accessors in O2MIndex as well?

src/libs/blueprint/conduit_blueprint_mesh_utils.hpp
```diff
@@ -215,6 +215,205 @@ CONDUIT_BLUEPRINT_API const Node * find_reference_node(const Node &node, const s
...
+    //-----------------------------------------------------------------------------
+    template<typename T, typename... Ts>
+    index_t     index(T idx, Ts... idxs) const;
+    template<typename T>
+    index_t     index(T idx) const;
+
+    /// With default argument, returns the number of ranks or dimensions for
+    /// this NDIndex.  With dim >= 0, returns the extent of this NDIndex
+    /// for dimension dim.
+    index_t     shape(index_t dim = -1) const;
```
cyrush: instead of magic number here, can we just add a dims() method?

agcapps: Sure. We had discussed shape(-1) returning the extent.  Dims() is fine.  I think I like rank() better, though.

cyrush: Yes, in hindsight it just looks too magical. Calling equiv: `.shape()` seems better, but still magical. 

rank would match fortran vernacular, correct? 

Does RAJA have a term for dims / rank/ etc?

Are there other sources of inspiration? 

What does Axom use? 

Note sure if numpy helps b/c folks just take len(v.shape) -- I don't think there is explicit method to ask this?

cyrush: Actually: numpy has:

`v.ndims`

cyrush:
Looks good thanks!

I though if a few minor things (file naming) moving the ndindex to its own file since its independent of the mesh blueprint.
I didn't articulate these a head of time, and I don't want to keep heaping on more, so I will make these adjustments after this PR.

cyrush (approved):

PR #1078, opened 2023-02-09
Feature/whitlock/new topology metadata

BradWhitlock:
This PR adds a new implementation of conduit::blueprint::mesh::utils::TopologyMetadata. The old one was preserved in conduit::blueprint::mesh::utils::reference::TopologyMetadata so we have it. The new version is used in the code now. The new class differs from the previous implementation in the following ways:

 * It uses recursion and template functions to make sure we can access
   connectivity data using either native pointers or and index_t_accessor
   so no Conduit casts are needed.
 * Associations are mostly created only when asked for (or indirectly needed)
   and this is done largely after the cascade has been performed. This is
   a huge source of complexity but it can let us compute fewer maps, saving 
   time and memory.
 * Associations are stored contiguously in memory or are implicit.
 * Access to associations is made using a "vector_view" which returns a window
   into the bulk association data so the caller can use the data without
   having to copy it out. Local association data is returned as a "range_vector"
   so the local ranges can be represented implicitly while still looking
   like a vector.
* It uses several for loops and sorts that can be parallelized. These loops and sorts
  are decorated with parallel execution policies (also new)(sort not yet implemented).
* The implementation is hidden in the .cpp file and there are some new methods for
  getting data. This allows some methods to be lazily evaluated.

Other changes:
* The TopologyMetadata classes were moved into conduit_blueprint_mesh_topology_metadata.{cpp,hpp}
* generate_xxx functions limit the maps that will be generated
* generate_xxx functions have the TopologyMetadata class convert data into a provided
  node, eliminating an extra copy operation.
* A new CI test was added for TopologyMetadata


BradWhitlock:
Here are some timings for TopologyMetadata with different mesh types on TOSS3 compiled with Intel 19. Both the reference and new implementation are timed. The new implementation is timed twice - once creating all maps and then only creating a subset of maps.

conduit_topmd_hex
conduit_topmd_tet
conduit_topmd_hexpoly

BradWhitlock:
Looks like I need to make some changes to satisfy the Windows compiler.

rhornung67:
@BradWhitlock these look like very impressive speedups. Excellent work!

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.hpp
```diff
@@ -0,0 +1,613 @@
...
+
+ Example:
+ L(3,2) = {0,1,2,3,4,5},{6,7,8,9,10,11},...
+
+               6 faces 
+           ----------------
+           |              |
+ L(2,3) = {0}{0}{0}{0}{0}{0}{1}{1}{1}{1}{1}{1}...
+
+ */
```
cyrush, line 219: These new comments are amazing, thank you for adding these!

cyrush (approved):
This looks great!

Thank you for this large set of improvements, use of caliper,  and also keeping the old implementation for reference.

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.hpp
```diff
@@ -0,0 +1,613 @@
...
+            cascade as well as all associations.
+
+     @param topology     The input topology node.
+     @param coordset     The input coordset associated with the topology.
+     */
+    TopologyMetadata(const conduit::Node &topology, const conduit::Node &coordset);
+
+    //-----------------------------------------------------------------------
+    /**
+     @brief Constructor for the NewTopologyMetadata class. This constructor
```
agcapps:
```suggestion
     @brief Constructor for the TopologyMetadata class. This constructor
```
If you write `NewTopologyMetadata`, I would be looking for the class to be called that.

BradWhitlock: Oops. I originally wrote this class outside Conduit with that name. I'll fix it.

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.hpp
```diff
@@ -0,0 +1,613 @@
...
+            produced will have length 0.
+     @return The topology length.
+     */
+    index_t get_topology_length(size_t dim) const;
+
+    //-----------------------------------------------------------------------
+    /**
+     @brief Get a global association for a particular entity_id in the entity_dim
+            dimension. This lets us ask a question like: which face ids are in
+            element 5? Or, which faces contain edge 12?
```
agcapps, line 290: I really appreciate these comments.

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
```diff
@@ -0,0 +1,4111 @@
...
+#include "conduit_utils.hpp"
+
+//#define DEBUG_PRINT
+
+#define EA_INDEX(E,A) ((E)*(MAX_ENTITY_DIMS)+(A))
+
+// This macro is used to highlight some code in the new implementation that
+// is geared at reproducing unintuitive behavior present in the reference
+// implementation.
+#define REPRODUCE_REFERENCE
```
agcapps, line 37: Much appreciated.

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
```diff
@@ -0,0 +1,4111 @@
...
+     @brief Returns the number of entities for L(e,a) so we can iterate through them.
+     @param e The entity dimension
+     @param a The association dimension
+     @return The number of entities in L(e,a).
+     */
+    index_t get_local_association_entity_range(index_t e, index_t a) const;
+
+    //-----------------------------------------------------------------------
+    /**
+     @brief Takes the input topology and reuses it has the highest topology.
```
agcapps: reuses?  ensures?

BradWhitlock: I'll fix: "Takes the input topology and uses it as the highest topology."

agcapps (approved):
thank you again, @BradWhitlock !

cyrush:
thanks @BradWhitlock !

BradWhitlock:
Thanks to @cyrush and @agcapps for reviewing!

PR #1084, opened 2023-03-03
conduit: provide an "ABI hash" for exposed C APIs

mathstuf:
Conduit is vendored or bundled into numerous packages from Kitware's Catalyst (mangled) to various simulations (not always mangled). In order to facilitate interoperability between at least mangled versions, offer a function to compute an "ABI hash" of the types which are directly accessible from the C API. If these are not the same, the safe thing to do is not allow the conversion to work in order to avoid provoking invalid accesses when blindly casting a pointer from a mangled instance to a non-mangled instance (and vice versa).

The idea is that any vendoring that involves mangling may provide functions along the lines of:

```c
// Forward declare the structures involved.
struct conduit_node;
struct mangled_conduit_node;

// Declare the unmangled function as "weak" so that if it exists, we
// call the real one while also not forcing it to exist.
extern conduit_uint64 conduit_abi_hash() __attribute__((weak));

// Possibly provide `const` overloads of these APIs as well.
conduit_node* mangled_to_unmangled(mangled_conduit_node* m) {
    conduit_uint64 mangled_hash = mangled_conduit_abi_hash();
    if (conduit_abi_hash) {
        return conduit_abi_hash() == mangled_hash ? (conduit_node*)m : NULL;
    }
    return NULL;
}

mangled_conduit_node* unmangled_to_mangled(conduit_node* u) {
    conduit_uint64 mangled_hash = mangled_conduit_abi_hash();
    if (conduit_abi_hash) {
        return conduit_abi_hash() == mangled_hash ? (mangled_conduit_node*)u : NULL;
    }
    return NULL;
}
```

so that the mangled and unmangled worlds may be bridged. Note that these functions must be provided in some "safe" space without seeing the mangling APIs to avoid affecting usage of the unmangled in the same TU.

---
Cc: @cyrush


cyrush:
@mathstuf FYI: I'll need some time to ponder, but I will be in meetings most of next week. I'll try to learn about this as soon as I have cycles. Thanks!

mathstuf:
I'm writing up tests in Catalyst to show how this is intended to work for more context here.

mathstuf:
Note that docs do need written up. It is *not* recommended to make these "smuggled" nodes children of "native" nodes (or vice versa) as while the memory layout for a given pointer may be compatible, there is no guarantee that memory allocation routines are compatible between the implementations, so freeing a node on the "wrong" side is not sensible. I think this means that "smuggled" nodes should be limited to "basic" data types and path traversal to read-only uses (except to set further "basic" data type values).

There is no place to store such a "smuggled" bit as smuggling one node means that its entire hierarchy is "smuggled" since one can traverse anywhere using `conduit_node_parent` and the various child node APIs.

mathstuf:
Implementation of the cast functions: https://gitlab.kitware.com/ben.boeckel/catalyst/-/commit/1889d6d73fc3a28ff553ad865d63c82a58440ea8

Test code: https://gitlab.kitware.com/ben.boeckel/catalyst/-/commit/2b5780b4ef06559aaba95f956a9eb554985166fb (incomplete; TODO items will be a more meaningful test other than "casting works" by actually using nodes from each side subject to the alloc/dealloc rules).

mathstuf:
Alternative approach where Catalyst optionally uses an external Conduit (while still providing a stable ABI). With this PR, we could embed the ABI hash into the `catalyst_impl` struct and refuse to call code with a mismatching Conduit ABI (though one may already be SoL by just loading the library, we can at least not make things worse and hopefully indicate that *something* is wrong).

https://gitlab.kitware.com/paraview/catalyst/-/merge_requests/82

mathstuf:
Closing; I don't think we need this.

cyrush:
Thanks!

PR #1086, opened 2023-03-08
fix diff rules for char8_strs

### LINKED ISSUE #1076: DataArray<T>::diff string comparison error
The diff method does not appear to compare strings properly if a string has been overwritten into a node that already had a longer string. I assume that it saw that it did not have to reallocate and it stores "quad\0" into the existing buffer. Then diff() subtracts 1 off the end of the char buffer length but that does not reflect the length of the string. So, we end up comparing "quad" vs "quad\0on" and the comparison fails.

1. Can the code encounter non-null terminated strings? Is that why it's even passing a length to std::string?
2. If we change to the typical std::string(const char*) constructor (ignoring any lengths) it works - but my strings are null terminated.

```
node["elements/shape"] = "polygon";
// some other logic
node["elements/shape"] = "quad";
```

Then I called some code in a test to compare my in-memory value to a baseline read by relay.
```
conduit::relay::io::load(filename, "yaml", baseline);
bool equal = !baseline.diff(n, info, tolerance, true);
```

cyrush:
@BradWhitlock  this should resolve #1076

- updates diff rules for strings to respect null term (still avoids realloc when a smaller string is set) 
- fixes diff_compatible to use a substring match (in the proper direction) instead of a full string match

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,13 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 #### Relay
 - Fixed MPI baton logic error in `conduit::relay::io::blueprint::save_mesh()` that caused MPI tasks to serialize writes to files in the N domains to M files case. Fix restores parallel write performance. This bug did not impact cases where where N domains were written to N files (N to N) or N domains were written to a single file (N to 1).
 
+### Changed
+#### General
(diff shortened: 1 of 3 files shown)


cyrush:
@BradWhitlock if you have a chance to check this out let me know, if not no worries.

cyrush:
To be clear, this should resolve #1076 (the string diff issue)

PR #1092, opened 2023-03-20
BitwidthMapping: only set `bitwidth_prefer_long_long` without C++11

### LINKED ISSUE #1090: macOS [u]int64 mismatches
In updating Catalyst's embedded Conduit from 0.8.4 to 0.8.6, [this error](https://gitlab.kitware.com/paraview/catalyst/-/jobs/7972350#L290-L296) showed up:

```
/opt/glr/paraview/catalyst-ci/thirdparty/conduit/conduit/conduit_data_type.cpp:39:1: error: static_assert failed due to requirement 'std::is_same<long, long long>::value' "error: conduit_int64 != std::int64_t"
static_assert(std::is_same<conduit_int64, std::int64_t>::value,
^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/glr/paraview/catalyst-ci/thirdparty/conduit/conduit/conduit_data_type.cpp:52:1: error: static_assert failed due to requirement 'std::is_same<unsigned long, unsigned long long>::value' "error: conduit_uint64 != std::uint64_t"
static_assert(std::is_same<conduit_uint64, std::uint64_t>::value,
^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I don't see anything interesting in the diff between 0.8.4 and 0.8.6 though…

MR updating Conduit is [here](https://gitlab.kitware.com/paraview/catalyst/-/merge_requests/82/diffs).

mathstuf:
CMake policy CMP0126[1], introduced in CMake 3.21, changes behavior for `set(CACHE)` so that local variable definitions are no longer removed when defining a value for the cache. This is part of a series of policy changes related to how cache and local variables interacted and making them more separate and easier to reason about (namely that changing cache variables doesn't affect local variables). However, in connection with the rule that local variables always win, this policy change means that code which conditionally does a cache setting must defer setting some default local value until it knows the cache will not be used.

Fix `bitwidth_prefer_long_long` to be set locally only if it is not meant to be set via the cache so that the default `FALSE` setting does not unconditionally get used.

[1] https://cmake.org/cmake/help/latest/policy/CMP0126.html

Fixes: #1090

### DIFF
--- a/src/libs/conduit/cmake/BitwidthMapping.cmake
+++ b/src/libs/conduit/cmake/BitwidthMapping.cmake
@@ -787,7 +787,6 @@ endmacro()
 # over `long`.
 #-----------------------------------------------------------------------------
 
-set(bitwidth_prefer_long_long FALSE)
 if(CONDUIT_USE_CXX11)
     # note: this call assumes this file included from libs/conduit/CMakeLists.txt
     try_compile( bitwidth_prefer_long_long   # result var
@@ -796,6 +795,8 @@ if(CONDUIT_USE_CXX11)
                  OUTPUT_VARIABLE bitwidth_prefer_long_long_out
                  CXX_STANDARD 11
                  CXX_STANDARD_REQUIRED TRUE)
+else()
+    set(bitwidth_prefer_long_long FALSE)
 endif()
 
 


cyrush:
Thanks for helping solve this, I expect we may hit more of these mysteries. 

But maybe not,  the surprise for me is that  `try_compile` is using a `CACHE` set, instead of a standard set.  I would not have assumed that.

cyrush (approved):

mathstuf:
> But maybe not, the surprise for me is that try_compile is using a CACHE set, instead of a standard set.

The idea is to only run the actual `try_compile` if it hasn't already run. If it were a local variable, it would have to test its result on every single CMake run.

cyrush:
@mathstuf thanks, I merged this fix.

PR #1094, opened 2023-03-21
adjust 2d cyl rz guidance and add rz examples

cyrush:


cyrush:
@agcapps can you give this a quick review?

agcapps (approved):
It looks good, Cyrus.

PR #1102, opened 2023-04-03
Python:  ABI3 compatibility changes

ChristosT:
This pull request enables [ABI3](https://docs.python.org/3/c-api/stable.html) compatibility for Python 3.11+  for the conduit python wrappings.

ABI3 compatibility allows to build a single library/wheel which can be used across multiple Python 3.x versions.

The  option is by default off . It can be enabled if the detected python version is >= 3.11 and the cmake option `CONDUIT_PYTHON_USE_LIMITED_API` is set to `ON`.

Python 3.11 was set as the minimum  version since the [`Py_Buffer`](https://docs.python.org/3/c-api/buffer.html)    object which is used by `PyConduit_Node_set_external` and `PyConduit_Node_set`  became part of the stable ABI only since 3.11. 

The changes are mainly of three types:
 - Replace  Python C-API calls with equivalent from the limited API e.g. `PyFloat_AS_DOUBLE` -> `PyFloat_AsDouble`
 - Make static `PyTypeObject` objects heap allocated. In the limited API `PyTypeObject`  objects are opaque pointers. They are created based on a `PyType_Spec` object using `PyType_FromModuleAndSpec` during the `init` phase of the module. e.g. https://github.com/LLNL/conduit/commit/60dd34e0a0ac09a047f110c4125389790a4fd6b6
 - Access `PyType` members through accessors instead of directly since this is also considered a opaque pointer e.g. 0ab6602edcc5986b05378f70a8d99d1ae46d99e9
 
Python tests pass locally on linux with the option enabled and using python 3.11.
Creating a wheel and testing against [abi3audit](https://github.com/trailofbits/abi3audit)  reports to no abi violations .

cc: @cyrush  @mathstuf

### DIFF
--- a/src/libs/conduit/python/conduit_python.cpp
+++ b/src/libs/conduit/python/conduit_python.cpp
@@ -201,6 +201,7 @@ struct module_state {
     PyObject *error;
     #ifdef Py_LIMITED_API
     PyTypeObject* PyConduit_DataType_TYPE;
+    PyTypeObject* PyConduit_Generator_TYPE;
     #endif
 };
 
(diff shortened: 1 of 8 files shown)


src/libs/conduit/python/conduit_python.cpp
```diff
@@ -139,9 +140,16 @@ static PyObject *
...
                                     int string_len)
 {
+    #ifdef Py_LIMITED_API
+    return PyUnicode_Decode(unicode_buffer,
+                             string_len,
+                             "utf-32",
+                             "strict");
+    #else
     return PyUnicode_FromKindAndData(PyUnicode_4BYTE_KIND,
                                      unicode_buffer,
```
cyrush, line 150: Thanks for fixing this case

cyrush (approved):
Everything looks good, thanks so much for these changes!

cyrush:
@ChristosT Follow up topic: 

I would like to learn your process for building a wheel.

I want to setup cibuildwheel so we can easily provide wheels for pypi.

ChristosT:
> Everything looks good, thanks so much for these changes!

Thank you for the review it was fun working on this!

> @ChristosT Follow up topic:
> 
> I would like to learn your process for building a wheel.
> 
> I want to setup cibuildwheel so we can easily provide wheels for pypi.

For this test I just created the wheel locally to test it against abi3audit. I don't have experience working with wheels but I know that we are using the https://github.com/pypa/manylinux containers on linux and  have separate ci jobs for each python version we support on windows and macos. cibuildwheel looks very nice I wasn't aware of them.

cyrush:
Thanks, I'll keep you posted when I start the pypi wheels adventure.

cyrush:
Thanks again for this -- merging.

I will add a Python 3.11 case to our CI as a follow up.

ChristosT:
> Thanks again for this -- merging.
> 
> I will add a Python 3.11 case to our CI as a follow up.

Sounds great ! Let me know if you encounter any issues with the new bits of code in the process .

Issue #1100, opened 2023-03-28
Error installing conduit via uberenv


dyokelson:
Hello, I tried the quick start for conduit on a few different machines and ran the following steps:
```
git clone --recursive https://github.com/llnl/conduit.git
cd conduit
python3 scripts/uberenv/uberenv.py --install --prefix="build"
```
One machine is skylake-gold with gcc/10.2, the other is broadwell with gcc/8.5. I have cmake/3.26 in my path. 
On both I got this same error posted below during the install conduit-develop phase. Please let me know if you have any ideas for a fix or if you need additional information. Thanks!

```
==> Installing conduit-develop-7f52qjlnxehehsvgxkzkroi5jl6hczoh
==> No binary for conduit-develop-7f52qjlnxehehsvgxkzkroi5jl6hczoh found: installing from source
==> No patches needed for conduit
==> conduit: Executing phase: 'hostconfig'
==> conduit: Executing phase: 'cmake'
==> conduit: Executing phase: 'build'
==> Error: FileNotFoundError: [Errno 2] No such file or directory: '/storage/users/dewiy/conduit/build/spack/opt/spack/linux-rhel8-broadwell/gcc-8.5.0/conduit-develop-7f52qjlnxehehsvgxkzkroi5jl6hczoh/.spack/test/cmake_bin_path.txt'

/storage/users/dewiy/conduit/build/spack/var/spack/repos/builtin/packages/conduit/package.py:279, in cmake_bin:
        276        """(Hack) Set/get cmake dependency path."""
        277        filepath = join_path(self.install_test_root, "cmake_bin_path.txt")
        278        if set:
  >>    279            with open(filepath, "w") as out_file:
        280                cmake_bin = join_path(self.spec["cmake"].prefix.bin, "cmake")
        281                out_file.write("{0}\n".format(cmake_bin))
        282        else:

See build log for details:
  /tmp/dewiy/spack-stage/spack-stage-conduit-develop-7f52qjlnxehehsvgxkzkroi5jl6hczoh/spack-build-out.txt

[ERROR: failure of spack install/dev-build]
```

dyokelson:
Also, this directory does not exist '/storage/users/dewiy/conduit/build/spack/opt/spack/linux-rhel8-broadwell/gcc-8.5.0/conduit-develop-7f52qjlnxehehsvgxkzkroi5jl6hczoh/'

cyrush:
[...]

PR #1108, opened 2023-04-14
ci updates and changelog update

cyrush:
- add python 3.11 case to azure
- add uberenv install case (tests quick recipe install from docs)
- add changelog entry for Python 3 Stable ABI enhancement


cyrush:
Need to solve a spack mystery to enable 3.11, which will take more time.

As is, the resolves #1100 which would be best to have in ASAP.

I will handle 3.11 in follow up PR.

PR #1113, opened 2023-05-08
Feature/whitlock/adjset verifier tool

### LINKED ISSUE #1097: adjset validator tool
we have utils to paint adjsets on to meshes for visual debugging - but we should also consider creating a tool that validates adjsets across domains and identifies any possible issues.

BradWhitlock:
This PR, for #1097, includes a new `conduit_adjset_validate` tool that can read a .root file containing multiple domains and it will scan through the adjsets it finds and it will note any problems that it finds (vertices or elements that do not exist in a neighbor). Errors are printed to the console. For vertex association adjsets, a "3D" file containing the offending points is written too so they can be viewed in VisIt.

There are a number of other enhancements:
* `utils::PointQuery` - new class that can be used to ask "does domain x contain point y". There is a parallel version too.
* `utils::kdtree` - new class that can accelerate point lookups in PointQuery when the number of points is larger
* `utils::MatchQuery` - new class that can be used to ask "does domain x contain element y". There is a parallel version too.
* `utils::TopologyBuilder` - new class that can simplify making a new topology from an existing topology (it helps renumber coordinates referenced by the topo).
* `topology::search` - Search for elements from one topo in a second topo.
* `generate_corners` will now apply a point query to filter out any vertices that do not really exist in neighbor domains when making a new adjset.
* `generate_derived_entities` (internal function used in `generate_faces`, etc) will apply a match query to make sure elements it references have a counterpart in a remote domain, if using adjsets.
* `adjset::validate` function validates adjset functions using a combination of PointQuery and MatchQuery.


BradWhitlock:
Ugh. What's with Python in CI? https://readthedocs.org/projects/llnl-conduit/builds/20566873/

cyrush:
@BradWhitlock  Read the docs env changed / broke last week. 

Develop has a fix, try to merge develop into your branch that should do the trick.

BradWhitlock:
Ok. Thanks, Cyrus.

cyrush (approved):
Great work, thanks for working through this complex issue!

My only suggested tweak is to provide a blueprint mesh output option for the debug output point mesh. We can discuss as a follow on.

Is there anything else you want to get in? If not lets merge.

BradWhitlock:
Hi Cyrus, thanks for reviewing. Good point about the pointmesh in blueprint format. That would make more sense.

PR #1120, opened 2023-05-13
Added destination_ranks property to field selection.

BradWhitlock:
This PR adds a `destination_ranks` property to field selections for the partitioner. This is optional. When present, we can use the property to indicate a preference for how the partitioner maps partitioned domains to MPI ranks. This comes in handy when we make multiple calls to partition()  (once for volume mesh and again for boundaries) and we need the different topologies to end up on the same MPI rank.

```
sel1["type"] = "field";
sel1["domain_id"] = "any";
sel1["field"] = "partitionField";
sel1["topology"] = "main";
// Domain 0 goes to MPI rank 0, 1->1, 2->2, 3->3.
sel1["destination_ranks"].set(std::vector<int>{0,1,2,3});
```


cyrush (approved):
Looks good!

Issue #1091, opened 2023-03-20
switch coordinate order for 2D cylindrical blueprint


cyrush:
While `RZ` is the name lovingly given to 2D cylindrical meshes, for most cases in the wild when plotting RZ data folks expect:

`Z` shown the horizontal coordinate
`R` shown as the vertical coordinate

This way when you revolve to 3D, the revolve happens around the horizontal axis.

While we can support `ZR` and `RZ` (For example, VisIt does support both)  there are confusing implications related to the implicit field association ordering for structured grids.

To avoid leading people down this confusing path, we should change the examples to present `Z` first, then `R` for 
2D RZ meshes. 

In addition, the VisIt Blueprint plugin (starting at 3.3.3) will always assume `Z==horiz` and `R==vert`,

So we should aligning Conduit docs.

cyrush:
Most of these have been changed, but the first instance in the docs still needs a fix.

cyrush:
last case resolved in #1122

cyrush:
resolved in #1122

Issue #1017, opened 2022-09-22
'make install' equivalent on Windows doesn't install '__init__.py' files.


biagas:
I compiled v 0.8.4 on Windows with Visual Studio 2017. When I built the 'install' target, none of the '__init__.py' files were installed.

cyrush:
More info from recent experience:

```
Output from the install command:
2>copying C:/A_Visit/ThirdParty/conduit-v0.8.7-build/libs/conduit/python/conduit_python_py_setup_build_install\lib\conduit\__init__.py -> C:A_Visit\ForRegression\visit-deps\windowsbuild\MSVC2017\conduit\0.8.7\python-modules\conduit
```

The backslash from C:\ was stripped off in the destination path.

This is a string escape issue. 

In this case:
`CMAKE_INSTALL_PREFIX=C:/A_Visit/ForRegression/visit-deps/windowsbuild/MSVC2017/conduit/0.8.7`

cyrush:
resolved in #1122

PR #1122, opened 2023-05-17
version string and docs updates for 0.8.8

cyrush:


rhornung67 (commented):
We don't need no stiking spaces!

cyrush:
Resolves #1091
Resolves #1124

cyrush:
resolves #1017

Issue #1110, opened 2023-04-25
Make Relay support 3D HDF5 hyperslabs


agcapps:
Currently, Relay sets the offsets and strides [using only the first element](https://github.com/LLNL/conduit/blob/636c024ffe2a04ce0b68a818f4d65cd355de96bb/src/libs/relay/conduit_relay_io_hdf5.cpp#L2732) of whatever the user passed in.  We should support N-D hyperslabs, that is, whatever the user passes in.  Additionally, we should support element counts in N-D, not just "number of elements."

agcapps:
In a little greater detail, I propose to allow the user to specify offset, stride, and size as length-n arrays.  The code that prepares to call `H5Sselect_hyperslab()` would properly use these arrays instead of hardcoding length-1 arrays.  The challenge is how to do this without allocation.

agcapps:
```
Node opts, n;
int offset[3] = {1, 0, 0};
int count[3] = {2, 2, 4};
opts["offset"].set(offset);
opts["count"].set(count);
int output[16];
n.set(output);

// HDF5 object at path "thefile.hdf5:myobj" has size (4, 3, 4), 
// that is, four slabs with three rows of four elements
// The following should read the middle two slabs, the first two rows, all four elements:

io::hdf5_read("thefile.hdf5:myobj", opts, n);
```

This assumes that the "shape" of in-memory hyperslab will always be linear.  `opts` always apply to the HDF5 hyperslab: when passed to `hdf5_read()`, the source; when passed to `hdf5_write()`, the destination.  I can imagine use cases for specifying the in-memory hyperslab beyond linear, but specifying the HDF5 hyperslab is a good start and needed as a feature right now.

agcapps:
PR #1129 started out as read/write but I reduced its scope to just reading.  I'll add writing in a subsequent PR.

PR #1129, opened 2023-05-26
Add ability to read N-D arrays from HDF5 files

agcapps:
Conduit currently reads and writes 1D arrays in HDF5 files.  There are 2D and 3D arrays that a user code needs to access.  This PR adds the ability to read ~~and write~~ those 2D and 3D arrays.

In more detail, Conduit uses HDF5 calls with 1D size, offset, and stride to read and write 1D arrays.  These parameters come through an "opts" Node passed to hdf5_read and write with keys "size", "offset", and "stride".  Currently, the user can set those keys to integers, appropriate to 1D arrays.  This PR lets users specify arrays of integers, not just scalars.

The HDF5 API allows us to specify a "hyperslab" for both HDF5 file and memory.  This PR allows Conduit to use the values specified in the "opts" Node to build the HDF5 file hyperslab: the source for hdf5_read ~~or the destination for hdf5_write~~.  The memory is a compact linear array.

A follow-up PR will add the ability to write N-D arrays.


agcapps:
Implements part of issue #1110 .

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -682,11 +724,17 @@ conduit_dtype_to_hdf5_dtype_cleanup(hid_t hdf5_dtype_id,
 //-----------------------------------------------------------------------------
 DataType
 hdf5_dtype_to_conduit_dtype(hid_t hdf5_dtype_id,
-                            index_t num_elems,
+                            hsize_t * num_elems_array,
```
cyrush, line 739: can we keep the old signature and have it plumbed into this new general one?

src/tests/relay/t_relay_io_hdf5.cpp
```diff
@@ -33,6 +33,43 @@ check_h5_open_ids()
 }
 
 
+//-----------------------------------------------------------------------------
+// helper to create an HDF5 dataset
+herr_t
+create_hdf5_dataset(std::string fname, std::string path, int rank, int const * dims, 
+    hid_t mem_type, hid_t file_type, void * to_write)
```
cyrush, line 40: How about: `create_hdf5_nd_dataset()` as a more descriptive name for this helper?

cyrush (commented):
Looks good! 

I added some minor suggestions via inline comments and I have two larger suggestions:

1) 

For the multi-d case, we should accept options:

```
sizes, offsets, strides
```

These are consistent with the way we describe how to access NDArrays for the more complex structured grid cases.

With your refactor, we should be able to easily also accept the 1D case args (size, offset, stride) but plumb to the general multi-d internals.

2) 

hsize_t will not always be defined as unsigned long long on all platforms.  In some cases it may be uint64_t. 

I think we should use Conduit's uint64 here instead of unsigned long long, b/c the bitwidth intent is clear. 

In any case we may need reinterpret casts to grab a hsize_t pointer from the conduit node's pointers. 

Another option is to always allocate hsize_t arrays, and then copy and cast each value in.  In that way we will never have an issue if `hsize_t` is something wildly unexpected.

Here is a partial example:

```
index_t_accessor sizes_vals = node["give/me/sizes"].value()
hsize_t *hdf5_sizes = new[rank];
/// .. and offsets and strides
for(int i=0;i <rank,i++)
{
  hdf5_sizes[i] = (hsize_t) sizes_vals[i];
  /// .. and offsets and strides
}
delete [] hdf5_sizes;
/// .. and offsets and strides
```

src/libs/relay/conduit_relay_io_hdf5.cpp
agcapps: Good idea.

src/tests/relay/t_relay_io_hdf5.cpp
agcapps: Done

src/libs/relay/conduit_relay_io_hdf5.cpp
agcapps: Done.

agcapps:
[2023.06.02 Conduit HDF5 partial call tree.pdf](https://github.com/LLNL/conduit/files/12052824/2023.06.02.Conduit.HDF5.partial.call.tree.pdf)

Here is a partial call tree I made of this git branch.  The top level functions are `hdf5_write` and `hdf5_read`.  They take an argument `const Node &opts` where a user will specify options such as `sizes, offsets, strides`.  The fact that the options are const is an obstacle: I would like to remove const.  This is because I would like to call the routine `fill_dataset_opts` one time to validate and fill out sizes, offsets, strides, and any other such object, and *cache* the validated, filled-out options within opts.  This is because the routines that use the options (color coded differently from the rest of the routines) are not grouped close together.

Alternately, I could call `fill_dataset_opts` whenever I need to validate and fill out the options.  This might remove the need to cache the results back into `opts`, and perhaps remove the strong urge to remove the `const` off the `opts` parameter.

@cyrush , what are your thoughts?

cyrush:
@agcapps the opts won't be large, so how about making a copy near the top of the call stack, normalize that copy, and then let it pass as const thorough the rest of the tree?

agcapps:
> @agcapps the opts won't be large, so how about making a copy near the top of the call stack, normalize that copy, and then let it pass as const thorough the rest of the tree?

That might work.  The issue I see with that is that stride, offset, and size are specific to one dataset/leaf and don't make sense applied to the tree.  Indeed I don't exactly know what we should do if there is more than one data set in the HDF5 file.

Thanks for the reply---I'll keep thinking about this.

cyrush:
@agcapps  got it, I still think some combo of just in time creation of the opts will be important. 

For this use case, I expect most folks will be using the options for a known leaf at a specific path, vs loading a tree or something that like that. In fact we should think about the proper error messages (not sure what was plumbed for the 1D case)

agcapps:
Now I need to properly check HDF5 object for compatibility with conduit leaf.

agcapps:
@cyrush , please review.  As you requested above, users can now use "sizes", "offsets", and "strides", or the singular form.  And we store what the user gives us, making copies of type `hsize_t` arrays to pass to the HDF5 and deleting them afterward.  I also added an example and some sphinx documentation that quotes the example and its output.

src/tests/docs/t_conduit_docs_relay_io_hdf5_examples.cpp
```diff
@@ -121,3 +121,95 @@ TEST(conduit_docs, relay_io_example_hdf5_interface_2)
...
+    // Two rows, two columns; total of four elements.
+    int constexpr rrowlen = 2;
+    int constexpr rcollen = 2;
+    int constexpr reltcount = rcollen * rrowlen;
+    int p_sizes[rank]{ rcollen, rrowlen };
+    // offset to row 0, column 1
+    int p_offsets[rank]{ 0, 1 };
+    // read every row, every other column
+    int p_strides[rank]{ 1, 2 };
+    // Store pointers to these parameters in the read_opts Node
```
cyrush, line 194: The row vs column vs order surprised me in the example (input params vs expected output. )

My brain making bad (flipped) assumptions, b/c In structured meshes, the offsets and strides are tied to the coordinate axes, so first axis is "x", second is "y", etc.  Can you can fully control which is fast varying relative to the coord system. 

How can we best document this?

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
+#### Relay
+- Added ability to read N-D HDF5 arrays into linear memory arrays.
+
```
cyrush, line 13:
```suggestion
- Added ability to read N-dimensional hyperslabs from HDF5 leaf arrays into linear memory arrays.
```

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
+#### Relay
+- Added ability to read N-D HDF5 arrays into linear memory arrays.
```
cyrush:
```suggestion
- Added ability to read N-dimensional hyperslabs from HDF5 leaf arrays into linear memory arrays.
```

cyrush (approved):
Looks good, Q about how we clearly document what is fast varying with respect to the implied HDF5 ND-layout.

src/tests/docs/t_conduit_docs_relay_io_hdf5_examples.cpp
agcapps: HDF5 says "fastest varying rightmost", like a C or C++ nested array.  I could write a sentence or two about that in the Sphinx and Doxygen docs.

agcapps:
I took a stab at clarifying HDF5 argument array order.  If you think this covers it, I'll go ahead and merge.

PR #1134, opened 2023-06-28
Update docs for issue 1133

### LINKED ISSUE #1133: small error in the documentation about the mixed topologies
Hi,

After some confusion of my part and discussion with @mennodeij. It looks like there is a small error in the doc about mixed topologies here : https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#mixed-topologies-with-shapes-shape-map

we specify  `polygon` instead of `polygonal` for the `shape_map` here:

`shape_map/polygon: (numeric entry for polygon, e.g. 7 - VTK_POLYGON)`

mennodeij:
Fixes https://github.com/LLNL/conduit/issues/1133


cyrush:
Thanks for this fix!

PR #1080, opened 2023-02-17
allow default constructed DataAccessors

cyrush:

PR #1138, opened 2023-07-14
Clean up templating in generate sides code

### LINKED ISSUE #1105: Return to generate_sides and use data accessors to reduce excessive templating
My work on `conduit::blueprint::mesh::topology::unstructured::generate_sides` led to excessive templating which can be much simplified now that we have data accessors.

JustinPrivitera:
Closes #1105

Now that we have data accessors and data arrays, a lot of the templating in the generate sides code was able to be removed/simplified.

Also adds the assignment operator to the data accessor.

The tests all pass.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -4990,28 +4990,26 @@ namespace detail
         return fabs((a - d).dot((b - d).cross(c - d))) / 6.0f;
     }
 
-    // U is the type of connectivity values
-    // V is the type of coordset values
-    template<typename U, typename V>
-    void
+    // determines the type of the coordinate values and calls
     // we want access to the new topology so we can calculate the areas
     // of the new triangles/volumes of the new tetrahedra
-    volume_dependent_helper(const Node &topo_dest,
-                            const Node &coordset_dest,
-                            const int dimensions,
-                            const int new_num_shapes, // number of new triangles or tetrahedrons
-                            const int num_orig_shapes, // number of original polygons or polyhedra
-                            int_accessor tri_to_poly,
-                            Node &volumes_info,
-                            Node &volumes_field_values)
+    void
+    volume_dependent(const Node &topo_dest,
+                     const Node &coordset_dest,
+                     const int dimensions,
+                     const int new_num_shapes, // number of new triangles or tetrahedrons
+                     const int num_orig_shapes, // number of original polygons or polyhedra
+                     int_accessor tri_to_poly,
+                     Node &volumes_info,
+                     Node &volumes_field_values)
     {
         // first we calculate the volume of each triangle
         volumes_field_values.set(conduit::DataType::float64(new_num_shapes));
-        float64 *tri_volumes = volumes_field_values.value();
+        float64_array tri_volumes = volumes_field_values.value();
 
-        const U *connec = topo_dest["elements/connectivity"].value();
-        const V *coords_x = coordset_dest["values/x"].value();
-        const V *coords_y = coordset_dest["values/y"].value();
+        int_accessor connec = topo_dest["elements/connectivity"].value();
+        float_accessor coords_x = coordset_dest["values/x"].value();
+        float_accessor coords_y = coordset_dest["values/y"].value();
 
         if (dimensions == 2)
         {
@@ -5029,7 +5027,7 @@ namespace detail
         }
         else if (dimensions == 3)
         {
-            const V *coords_z = coordset_dest["values/z"].value();
+            float_accessor coords_z = coordset_dest["values/z"].value();
 
             for (int i = 0; i < new_num_shapes; i ++)
             {
@@ -5045,6 +5043,7 @@ namespace detail
                 vec3 d = vec3(static_cast<float64>(coords_x[connec[i * 4 + 3]]),
                               static_cast<float64>(coords_y[connec[i * 4 + 3]]),
                               static_cast<float64>(coords_z[connec[i * 4 + 3]]));
+                
                 tri_volumes[i] = tetrahedron_volume(a,b,c,d);
             }
         }
@@ -5055,7 +5054,7 @@ namespace detail
 
         // next we calculate the volume of each polygon
         volumes_info["poly"].set(conduit::DataType::float64(num_orig_shapes));
-        float64 *poly_volumes = volumes_info["poly"].value();
+        float64_array poly_volumes = volumes_info["poly"].value();
 
         for (int i = 0; i < num_orig_shapes; i ++)
         {
@@ -5068,7 +5067,7 @@ namespace detail
 
         // finally we calculate the volume ratio
         volumes_info["ratio"].set(conduit::DataType::float64(new_num_shapes));
-        float64 *ratio = volumes_info["ratio"].value();
+        float64_array ratio = volumes_info["ratio"].value();
 
         for (int i = 0; i < new_num_shapes; i ++)
         {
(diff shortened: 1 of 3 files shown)


src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -4990,29 +4990,26 @@ namespace detail
...
     {
         // first we calculate the volume of each triangle
         volumes_field_values.set(conduit::DataType::float64(new_num_shapes));
-        float64 *tri_volumes = volumes_field_values.value();
+        float64_array tri_volumes = volumes_field_values.value();
 
-        const U *connec = topo_dest["elements/connectivity"].value();
-        const V *coords_x = coordset_dest["values/x"].value();
-        const V *coords_y = coordset_dest["values/y"].value();
+        int_accessor connec = topo_dest["elements/connectivity"].value();
```
cyrush: to be safe, i think these guys should be `int64_accessor` and `float64_accessor`

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -5030,7 +5027,7 @@ namespace detail
         }
         else if (dimensions == 3)
         {
-            const V *coords_z = coordset_dest["values/z"].value();
+            float_accessor coords_z = coordset_dest["values/z"].value();
```
cyrush: `float64_`

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -5069,117 +5067,28 @@ namespace detail
...
-    }
-
-    template<typename U, // U is the type of the new field values (should typically be the same as V)
-             typename V, // V is the type of the old field values (should typically be the same as U)
-             typename W> // W is the type of the new "topo/elements/connectivity" values
+    template<typename T> // T is the type of the new field values
     void
     vertex_associated_field(const Node &topo_dest,
-                            const V *poly_field_data,
+                            float_accessor poly_field_data,
```
cyrush: `float64_`

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -5188,11 +5097,9 @@ namespace detail
         std::map<int, std::set<int>> info;
 
         int iter = dimensions == 2 ? 3 : 4;
-        const W *new_connec = topo_dest["elements/connectivity"].value();
+        int_accessor new_connec = topo_dest["elements/connectivity"].value();
```
cyrush: `int64_`

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -5275,54 +5180,20 @@ namespace detail
                                  const Node &topo_dest)
     {
         // a pointer to the destination for field values
-        U *values_array = field_out["values"].value();
+        T *values_array = field_out["values"].value();
 
         // a pointer to the original field values
-        const V *poly_field_data = field_src["values"].value();
+        float_accessor poly_field_data = field_src["values"].value();
```
cyrush: `float64_`

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -5387,7 +5256,7 @@ namespace detail
             CONDUIT_ERROR(((std::string) "Bad shape in ").append(topo_dest["elements/shape"].as_string()));
         }
 
-        const T *tri_to_poly = d2smap["values"].value();
+        int_accessor tri_to_poly = d2smap["values"].value();
```
cyrush: `int64_`

cyrush (commented):
Looks good - Small suggestion related to the accessor types, we want to uses the 64-bit ones (float64_accessor and int64_accessor)

This cleans things up nicely!

JustinPrivitera:
@cyrush are there any plans to allow
```
float64_accessor myaccessor; // I am not initializing this accessor here but plan to later on
```
without a compiler warning (or was it an error? I don't remember)

cyrush:
> @cyrush are there any plans to allow
> 
> ```
> float64_accessor myaccessor; // I am not initializing this accessor here but plan to later on
> ```
> 
> without a compiler warning (or was it an error? I don't remember)

We fixed so you can you can do this now:

(as of https://github.com/LLNL/conduit/pull/1080)

JustinPrivitera:
I got worried when I saw this:
```
[ 45%] Building CXX object libs/blueprint/CMakeFiles/conduit_blueprint.dir/conduit_blueprint_mesh.cpp.o
/usr/workspace/justin/conduit_builds/develop/conduit/src/libs/blueprint/conduit_blueprint_mesh.cpp: In function ‘void conduit::blueprint::detail::map_fields_to_generated_sides(const conduit::Node&, const conduit::Node&, const conduit::Node&, const conduit::Node&, const conduit::Node&, const conduit::Node&, conduit::Node&, const std::vector<std::__cxx11::basic_string<char> >&, const string&)’:
/usr/workspace/justin/conduit_builds/develop/conduit/src/libs/blueprint/conduit_blueprint_mesh.cpp:5381:64: warning: implicitly-declared ‘conduit::DataAccessor<double>& conduit::DataAccessor<double>::operator=(const conduit::DataAccessor<double>&)’ is deprecated [-Wdeprecated-copy]
 5381 |                     volume_ratio = volumes_info["ratio"].value();
      |                                                                ^
In file included from /usr/workspace/justin/conduit_builds/develop/conduit/src/libs/conduit/conduit_node.hpp:35,
                 from /usr/workspace/justin/conduit_builds/develop/conduit/src/libs/conduit/conduit.hpp:25,
                 from /usr/workspace/justin/conduit_builds/develop/conduit/src/libs/blueprint/conduit_blueprint_mcarray.hpp:17,
                 from /usr/workspace/justin/conduit_builds/develop/conduit/src/libs/blueprint/conduit_blueprint_mesh.cpp:33:
/usr/workspace/justin/conduit_builds/develop/conduit/src/libs/conduit/conduit_data_accessor.hpp:55:9: note: because ‘conduit::DataAccessor<double>’ has user-provided ‘conduit::DataAccessor<T>::DataAccessor(const conduit::DataAccessor<T>&) [with T = double]’
   55 |         DataAccessor(const DataAccessor<T> &accessor);
      |         ^~~~~~~~~~~~
```

JustinPrivitera:
It's your call - do you want me to remove a float64* to replace it with a float64_accessor at the cost of adding this warning to the build? Or should I leave as is?

cyrush:
I see -- it looks like we need to add the assignment op to the `DataAccessor`.

Can you add this? We can follow the example from the `DataArray` class

https://github.com/LLNL/conduit/blob/5aef054fa682d7aa0ff307f7d034f16505f8991c/src/libs/conduit/conduit_data_array.hpp#L66

https://github.com/LLNL/conduit/blob/5aef054fa682d7aa0ff307f7d034f16505f8991c/src/libs/conduit/conduit_data_array.cpp#L80

JustinPrivitera:
done

cyrush:
Awesome! 

Extended thought exercise:

What if we could write to a `float64_accessor`, such that we don't need T *values_array?

JustinPrivitera:
Do you mean a `float64_array`?

I thought about that. For vertex associated and volume dependent fields, we made the choice to always write to float64 since we need to do some extra calculations to figure out what the right values should be. Two years ago, we also made the choice that if a field was neither vertex associated nor volume dependent, we would make sure the new field honored the original field's type.

While I was going through this earlier today, I considered changing it so that we always write out float64. Then we can do away with templating entirely. What do you think?

PR #1143, opened 2023-07-18
data accessor set

cyrush:
Expand DataAccessor API to support set.
Expand DataArray set API


src/libs/conduit/conduit_data_array.hpp
```diff
@@ -355,6 +388,7 @@ class CONDUIT_API DataArray
 // -- stdout print methods ---
 //-----------------------------------------------------------------------------
     /// print a simplified json representation of the this node to std out
+    // TODO: change to to_summary_string?
```
JustinPrivitera: fix before merge?

JustinPrivitera (approved):
Looks good.

src/libs/conduit/conduit_data_array.hpp
cyrush: I pushed this change

PR #1150, opened 2023-07-25
Update t_blueprint_mesh_relay.cpp

acbauer:
Fix spacing Conduit Node names for uniform mesh
    
Names were x and y instead of the proper dx and dy.


acbauer:
@cyrush please take a look. I think this fix is correct.

cyrush (approved):
Yes this looks correct -- thanks for spotting and changing this.

acbauer:
Saw the same issue with a ParaView Catalyst v2 example.

cyrush:
@acbauer  merged, thanks!

PR #1152, opened 2023-07-27
Fix a narrowing conversion

benson31:
It looks like #1129 introduced a narrowing conversion:
```
src/libs/relay/conduit_relay_io_hdf5.cpp:731:36: error: non-constant-expression cannot be narrowed from type 'index_t' (aka 'long') to 'hsize_t' (aka 'unsigned long long') in initializer list [-Wc++11-narrowing]
    hsize_t num_elems_array[1] = { num_elems };
                                   ^~~~~~~~~
```
I avoid this with an explicit `static_cast` -- I assumed that one cannot have a negative number of elements.


cyrush:
@benson31 thanks for this fix

PR #1158, opened 2023-08-04
Helper files for debuggers

agcapps:
This PR implements helper files for three popular debuggers.

For MS Visual Studio, a .natvis file; for Totalview, a data display function to implement C++View; for gdb, a Python file.  So far, we display conduit::Node objects.  Haven't added iterators or other types yet.


src/debug/msvs/ConduitNode.natvis
```diff
@@ -0,0 +1,78 @@
...
+                <Size>m_children.size()</Size>
+                <ValuePointer>m_children._Mypair._Myval2._Myfirst</ValuePointer>
+            </ArrayItems>
+            <CustomListItems Condition="m_schema->m_dtype.m_id == CONDUIT_OBJECT_ID">
+                <Variable Name="i" InitialValue="0"/>
+                <Variable Name="hier" InitialValue="(conduit::Schema::Schema_Object_Hierarchy *)(m_schema->m_hierarchy_data)"/>
+                <Size>m_children.size()</Size>
+                <Loop Condition="i&lt;m_children.size()">
+                    <!-- This works only if the order of the children hasn't been messed up. -->
+                    <!-- I haven't found a way to get a temporary Node here in the loop. -->
```
agcapps, line 31: This works in the simple case, for objects that get created and never have Nodes removed.  I don't know if I can get this to work in the general case, where the order might get perturbed.

JustinPrivitera (commented):
This is cool and seems very useful. I look forward to taking advantage of this in the future.

BradWhitlock:
@agcapps - this is a terrific idea! I'd like to see a short Sphinx doc page that says how to install/use them with the various debuggers.

agcapps:
@JustinPrivitera , thanks!

@BradWhitlock , good idea.  I'll add documentation.

agcapps:
At this point, I'm stuck on the GDB file.  I wrote it as I understood the documentation, and following any examples I could find.  So far, the "hash-like" interface for object Nodes does not work.

cyrush:
should we get the others in and add a ticket for full gdb support in the future?

agcapps:
> should we get the others in and add a ticket for full gdb support in the future?

Yes, I would like to commit.  This PR has languished long enough.  I'll add some screen shots and clean up construction debris, and mark as "not draft."

cyrush:
Sounds good!

PR #1160, opened 2023-08-11
Added field_prefix to the options for conduit::blueprint::mpi::mesh::partition_map_back

BradWhitlock:
The generate_partition_field function takes a "field_prefix" option that is prepended to variables that it creates (such as global_vertex_ids). The partition_map_back function needs global_vertex_ids to map vertex associated fields back to the original mesh. It was looking for "global_vertex_ids" directly. If a field_prefix was used previously then it would not find the field and mapback would fail. This PR adds the same "field_prefix" to the options for partition_map_back so it can find the right field.


cyrush (approved):
Looks good, thank for this fix!

PR #1162, opened 2023-08-27
add support for conduit_json_external protocol

cyrush:
`to_json` gains new protocol (`conduit_json_external`) that embeds the address of each leaf in its schema. 
When parsing,  `conduit_json` handles schemas with optional `address` entry.

With this:
```
Node n, n2;
std::string json = n.to_json("conduit_json_external");

// the following are equivalent:
n2.set_external(n);
n2.parse(json,"conduit_json")

```

Also includes changes to improve performance of a few json parsing cases, avoiding intermediate copies and some intermediate use of json strings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,16 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 
 ### Added
 
+#### General
+- Added `conduit_json_external` protocol. Creates a json schema representation of a node that includes all addresses that the node is pointing to. Parsing this schema will create a node equivalent to `set_external()`.
+
 #### Relay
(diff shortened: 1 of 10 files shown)


cyrush:
@mathstuf @ChristosT  

It has been a bit since we discussed this, but this new feature will allow you to use a json string rep to fully describe an in-memory node and replicate as a new node with external data. This should help with your use case of passing data between to library instances. 

Let me know if you have any questions or hit roadblocks for that case.

mathstuf:
Thanks! We now support external Conduit, so things should be flexible enough. However, this can help with the "I'm stuck with a conduit and internal-conduit catalyst, how can I make the best of this situation" case.

PR #1167, opened 2023-09-20
Create tiled mesh example function.

BradWhitlock:
This PR includes a tiled() example mesh generation function that can take a 2D topology and stamp it out a bunch of times to make either 2D or (in most cases) 3D meshes. The tiled() function generates a single domain but can be used to make a multidomain mesh too with adjsets and boundaries. The mesh is generated using a square tiling of the tile topology but this can be put through a reorder() function that reorganizes the mesh points and elements. Reordering is currently done using a kdtree.

I added a conduit_generate_data utility that can invoke the tiled function to make a multi-domain mesh and save it to a file. It can also call braid(). There are times when I just want to make a mesh with Conduit without having to write my own program.

This update includes a adjset::compare_pointwise function that compares adjsets across domain boundaries to make sure they result in the same point on both sides of the domain boundary. I wrote it because I was messing things up when I built my adjset.

I added some sphinx doc for the tiled() example function and updated the CHANGELOG.md file too.

teton_newmesh_120_pcplot


src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
```diff
@@ -345,6 +345,217 @@ find_domain_id(const Node &node)
...
+    {
+        auto dest(n_dest_values.as_uint16_array());
+        typed_slice_array(n_src_values.as_uint16_array(), ids, dest);
+    }
+    else if(dt.is_uint32())
+    {
+        auto dest(n_dest_values.as_uint32_array());
+        typed_slice_array(n_src_values.as_uint32_array(), ids, dest);
+    }
+    else if(dt.is_uint64())
```
cyrush, line 413: should we add float32 and float64 cases for completeness?

cyrush (approved):
Nice work! 
Thanks for the new example, great docs, and refactoring the utils for broad use.
Found a case where we may want extra type logic for float32 and float64 (this case would be covered by double and float, but for consistency and completeness I think we should add)

src/docs/sphinx/blueprint_mesh.rst
```diff
@@ -1687,6 +1687,90 @@ equal to half the number of prisms.
...
+
+tiled
++++++++++
+
+The ``tiled()`` function repeats a tile (given as a 2D Blueprint topology) into a larger mesh composed of
+a regular square tiling of the supplied tile. If no topology is given, a default pattern consisting of quads
+is used. The output mesh can be either 2D or 3D. For 3D, supply a ``nz`` parameter greater than zero. Note
+that the input tile must consist of a homogeneous set of triangles or quads to extrude the tile into 3D since
+polyhedral output is not yet supported. The ``tiled()`` function produces a single domain comprised of a
+main mesh, a boundary mesh, and adjacency sets of the output mesh is to be part of a multi-domain dataset.
```
agcapps: I think you mean "... and adjacency sets **if** the output mesh is to ...".

BradWhitlock: Good catch. I made a few other typos there that I'm fixing now.

agcapps (approved):
Thanks for the typo fixes.  This work is very useful.

src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
BradWhitlock: I added those cases.

cyrush:
thanks @BradWhitlock !

PR #1172, opened 2023-09-28
Enhance conduit_relay_io_convert so it can make files for VisIt.

BradWhitlock:
This PR enhances conduit_relay_io_convert so it can convert yaml to a blueprint.root file suitable for VisIt. Pass "blueprint" for the read or write protocol.

```
conduit_relay_io_convert input.yaml output --write-protocol blueprint
```


cyrush (approved):
Thanks!

PR #1179, opened 2023-10-26
Node Swap now correctly swaps schema parents

### LINKED ISSUE #1178: Confused about move to non-existing subtree
Hi,

I was using node `move` earlier today in the code which can be represented by the following MVP:

```cpp
#include <conduit.hpp>

int main (int argc, char *argv[]) {
    
    conduit::Node root;
    double* ptr = new double[3]{1, 2, 3};
    conduit::Node subtree;
    subtree["to/data"].set_external(ptr);
    root["path"].move(subtree);
    root.print();
    std::cout << "Node name: " << root["path"].name() << std::endl;
    return 0;
}
```
I am a bit confused by this. On one side, I feel it shouldn't work at all as `"path"` does not exist when I am moving the tree. Printing out the structure though confirms that it was created as intended. When asking explicitly about the name results in an empty string. Can you please advise, what's happening here?

Also, if I use a copy constructor
```cpp
conduit::Node copy = root;
std::cout << "Node name after copy: " << copy["path"].name() << std::endl;
```
I get an intended "path" as output. So only the original now has a missing node name.

JustinPrivitera:
Resolves #1178

### DIFF
--- a/.github/workflows/uberenv_install.yml
+++ b/.github/workflows/uberenv_install.yml
@@ -42,7 +42,7 @@ jobs:
         submodules: 'recursive'
     - name: Install
       run: |
-        python3 scripts/uberenv/uberenv.py --install --prefix="build" --spack-config-dir=scripts/uberenv_configs/spack_configs/configs/alpinedav/ubuntu_18.04_devel/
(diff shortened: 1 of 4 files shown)


JustinPrivitera:
all tests pass

JustinPrivitera:
TODO add entry to the changelog

cyrush (approved):
Looks good, thanks for diving in an fixing this!

PR #1181, opened 2023-10-30
head rtd

cyrush:


cyrush:
also fixes uberenv_install github actions case

PR #1184, opened 2023-10-30
py wrap for mesh bp examples strided_structured and py test

cyrush:


gunney1 (commented):
Looks good.  Thanks @cyrush

PR #1189, opened 2023-11-07
refactor zlib cmake and add smoke test

cyrush:


cyrush:
closed in favor of clean PR #1190

PR #1192, opened 2023-11-09
Blueprint Silo I/O: Fix several minor TODOs

JustinPrivitera:
Replace `CONDUIT_ERROR` with `CONDUIT_ASSERT` where possible.

Refactor domain type info bookkeeping into a helper method.

Refactor assembly of root type domain info.


cyrush (approved):
Thanks for these changes!

PR #1194, opened 2023-11-10
Venn Example: Remove non-blueprint "params"

### LINKED ISSUE #1163: Venn example has non-blueprint "params" stored in its coordset
https://github.com/LLNL/conduit/blame/733ac4e92b1964d9f2e334961bfc9036ff422653/src/libs/blueprint/conduit_blueprint_mesh_examples_venn.cpp#L550

In
```
void venn(const std::string &matset_type,
          index_t nx,
          index_t ny,
          float64 radius, 
          Node &res)
```
we add this:
```
    // Not part of the blueprint, but I want these values handy
    res["coordsets/coords/params/nx"] = nx;
    res["coordsets/coords/params/ny"] = ny;
```
These values are accessed and used in helper functions. They could probably just be passed directly instead of being stored in the blueprint. Should we make that change? It is strange to have non-blueprint data come out of one of the blueprint mesh examples.

JustinPrivitera:
Resolves #1163

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh_examples_venn.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_examples_venn.cpp
@@ -63,11 +63,9 @@ namespace examples
 {
 
 //---------------------------------------------------------------------------//
-void venn_full_matset(Node &res)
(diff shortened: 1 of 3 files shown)


cyrush (approved):
Thanks for cleaning this up!

PR #1195, opened 2023-11-10
fix a wide variety of compiler warnings

ibaned:
1. variables set but not used
2. sprintf is deprecated on Mac, use snprintf
3. variables used uninitialized if error occurs


cyrush (approved):
Thanks for these fixes!

Issue #1111, opened 2023-05-03
Continued Silo/Overlink I/O Support


JustinPrivitera:
Once #850 is complete these will be the next important tasks:

### Short term:
 - [x] materials - #1171
 - [x] ability to output sparse by material and full matsets from silo -> blueprint - #1236
 - [x] row major/column major + striding for quadvars, reorder??? - see visit plugin for inspiration - #1274 
 - [x] specsets - #1242 - #1348
 - [x] adjsets - #1210
 - [x] overlink support - nearly there; requires materials support and a few other considerations - #1249
 - [x] a new solution is needed for the opts mesh name name collision problem. Perhaps for the root only single domain case we could always force 1 domain. Thus mmesh name will always be the topo name. opts mesh name will not be used anymore. If there is going to be a directory name collision we could make it be dom0. - #1337 
 - [x] name schemes - #1424
 - [x] polygonal support - #1253
 - [ ] polyhedral support
 - [x] plumbing the write and read silo calls into blueprint write and read at a high level - blueprint load_mesh needs to call silo load_mesh if it detects the file is a silo file. HDF5 detection. Errors if Silo isn't there. - #1356
 - [x] units/labels - #1276
 - [ ] overlink conduit tests over in the overlink conduit repo
 - [ ] visit tests
 - [x] ability to read files with no multimeshes - #1274 
 - [x] fix naming conventions so there are no name collisions and save names appropriately - #1337
 - [ ] parallel error handling
 - [ ] refactor blueprint and silo i/o so they use the same logic under the hood - some of this was done in #1393 
 - [ ] take advantage of expanded specset representation - for reading, writing, and test utils. Test all the representations diff correctly for round trip. See TODOs around the code. Do it once #1349 is complete.
 - [ ] symlink for non-windows/copy the root file for windows for overlink case
 - [ ] mixed topo case
 - [ ] better file open handling. Perhaps a map of names to open file handles.
 - [ ] understand the too many open file handles issue

### Long Term:
 - [ ] support the "pre-mixed" case, the list case
 - [ ] AMR silo to blueprint support
 - [ ] we would like it if, when reading overlink, we could use the padding dimensions to create a field that paints the nodes one color if they are ghosts and another if they are not. Extra credit if you also check for the existence of this field when writing padding dimensions, walk it and determine if padding dims can be extracted from it.

### Optional:
 - [ ] we can add a recursive read of multivars and multimeshes placed in subdirectories (like what visit does) if someone needs it.
[...]

PR #1197, opened 2023-11-14
Overlink I/O: Read and Write VAR_ATTRIBUTES

JustinPrivitera:
Addresses a bullet in #1111

Also adds volume dependence to blueprint index, if available.


src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -1944,6 +1989,84 @@ read_state(DBfile *dbfile,
...
+{
+    if (! root_node[multimesh_name].has_child("vars"))
+    {
+        // nothing to do if there are no vars
+        return;
+    }
+
+    // TODO it is annoying to see the error message that this was not found in the
+    // silo file every time. Is there a way to check that it is there OR to not call this
+    // unconditionally?
```
cyrush: I think we can use `DBInqVarExists` and `DBInqVarType` to check if this is there and then make sure it is the type we expect.. 

I think that would remove the error messages.

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4003,6 +4154,103 @@ write_multimats(DBfile *dbfile,
...
+
+            elemlengths.push_back(num_attr);
+            nvalues += num_attr;
+
+            // 
+            // centering: ATTR_NODAL 0, ATTR_ZONAL 1, ATTR_FACE, ATTR_EDGE
+            // 
+            if (n_var["association"].as_string() == "vertex")
+            {
+                var_attr_values.push_back(0); // nodal == vertex
```
cyrush, line 4202: Would it help to use the Silo enum types here, instead of the raw values?

cyrush (approved):
Looks great!

Two minor suggestions, one addressing how to avoid the error messages when var atts doesn't exist.

src/libs/relay/conduit_relay_io_silo.cpp
JustinPrivitera: I will add that here, and maybe in a follow up PR add that elsewhere when reading other variables that may not be present.

src/libs/relay/conduit_relay_io_silo.cpp
JustinPrivitera: We don't have access to the raw enum types, as these are Overlink enum values. Even if Overlink and Silo agree on the enum values and meanings (not sure if they do for centering or not), I think it is cleaner to use raw values and clearly label what the Overlink values mean.

cyrush:
My bad , I forgot they were not  proper silo enums. Sorry!

-Cyrus

PR #1199, opened 2023-11-16
Silo I/O: Check existence of vars before reading them

JustinPrivitera:
This avoids many errors and warnings.


cyrush (approved):
Looks good!

PR #1202, opened 2023-11-17
Overlink I/O: Lay the groundwork

JustinPrivitera:
This PR makes numerous changes to bring us in line with the Overlink spec, including changes to reading and writing multimeshes, multimaterials, variable attributes, multivariables, meshes, materials, and variables, both unstructured and structured. The goal of this work is to tighten everything that we have up so that we can move forward with other missing pieces for Overlink.


cyrush (approved):
Looks good!

PR #1208, opened 2023-12-07
Silo/Overlink I/O: Fix matmap logic and support separation of non-scalar vars for overlink case

JustinPrivitera:
- Use simplified material map logic as discussed; defer to multi materials if information is present.
- Separate non-scalar fields into multiple scalar fields for writing.
- Tests for the latter.


cyrush (approved):
New code and test looks good!

PR #1210, opened 2023-12-08
Overlink I/O: Read and Write Support for Adjsets

JustinPrivitera:
Also simplified tests and added new tests for this case.

### DIFF
--- a/src/libs/relay/conduit_relay_io_silo.cpp
+++ b/src/libs/relay/conduit_relay_io_silo.cpp
@@ -1763,7 +1763,7 @@ read_matset_domain(DBfile* matset_domain_file_to_use,
 void
 read_adjset(DBfile *dbfile,
             const std::string &multimesh_name,
-            Node &root_node,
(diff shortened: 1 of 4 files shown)


src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -1759,6 +1759,126 @@ read_matset_domain(DBfile* matset_domain_file_to_use,
...
+    {
+        // there are no neighboring domains, so we cannot create an adjset
+        return;
+    }
+
+    // create the adjset
+    Node &adjset_out = mesh_out["adjsets"]["adjset"];
+    adjset_out["topology"] = multimesh_name;
+    adjset_out["association"] = "vertex";
+    // TODO how do I handle element case?
```
JustinPrivitera: @cyrush what do you think should happen for the element-associated adjset case?

cyrush (approved):
looks good, thanks for these additions!

PR #1212, opened 2023-12-13
Fix kdtree extents to better support queries of planar arrangements of points in 3D.

BradWhitlock:
I was running into a weird problem where an element adjacency set produced from generate_faces() was missing certain groups in the adjacency set. I traced this to a problem in the kdtree where the overall bounding box of the points being queried had the same values for one of the dimensions. Later on, this would prevent successful point lookups because the kdtree determined the query point was outside of the dataset.

This change enhances the kdtree logic a little so the bounding box will always have some thickness, even when all of the points it contains are coplanar and contain the same values in x or y or z. I added tests that used to cause this to fail and now they pass with the fix.


cyrush (approved):
Looks good!

PR #1213, opened 2023-12-14
Fix Fortran fetch int* procedure pointers

rpmcnally:
The fetch_path_as_int*_ptr methods on the Fortran node type were pointing to the non-ptr versions.  Compilation would fail when attempting to use the routines.


cyrush (approved):
Thanks for finding this issue and resolving w/ new tests!

cyrush:
I started CI checks, will merge after those are run.

cyrush:
Merged, thanks again @rpmcnally !

PR #1214, opened 2023-12-16
Enhanced tiled mesh generation so it can work top-down.

BradWhitlock:
This PR enhances the tiled mesh generator so it can work top-down. The user gives the number of dimensions for the overall mesh and the number of domains to create and then it will divide the problem using a block-based approach, and optionally a hilbert-curve splitting method. This tiler enhancement makes it easier to generate meshes suitable for strong-scaling.

There are a number of other fixes/enhancements:
* compare_pointwise was silently passing on non-canonical adjsets
* Adjset node ordering fixes meshes output from the the original tiler
* adjset::to_topo() function makes point meshes or surfaces from adjsets
* The adjset validation program can output a point mesh of all of the adjset groups so they can be visualized in VisIt
* Several testing improvements


cyrush (approved):
thanks for the improvements, and all the extra work on diagnostics to help with debugging.

BradWhitlock:
Thanks @cyrush for reviewing this.

PR #1218, opened 2023-12-21
Fixes for blueprint::utils::ShapeType.

### LINKED ISSUE #1217: Errors in blueprint::utils::ShapeType
I had some code that inadvertently passed a uniform topology to blueprint::utils::ShapeType and then called shape.is_poly() on the resulting shape. That call returned true, causing the topology to be passed to generate_sides(), which then crashed.

* Make ShapeType able to initialize using additional topology types (For uniform, initialize using a quad or hex as needed).
* Fix ShapeType::is_poly() so it only will return true for polygonal and polyhedral shapes, as intended. Do not return true for shapes that did not initialize.
* Check the topology input to unstructured::generate_xxx functions and error out if the topology is not unstructured.

```
#include <conduit_blueprint.hpp>
#include <conduit_blueprint_mesh_examples.hpp>
#include <conduit_blueprint_mesh.hpp>
#include <conduit_blueprint_mesh_utils.hpp>

#include <iostream>
using std::cout;
using std::endl;

int
main(int argc, char *argv[])
{
    conduit::Node out;
    conduit::blueprint::mesh::examples::braid("uniform", 5, 5, 0, out);
    out.print();
    const conduit::Node &n_topo = out.fetch_existing("topologies/mesh");

    conduit::blueprint::mesh::utils::ShapeType shape(n_topo);
    cout << "shape.dims=" << shape.dim << endl;
    cout << "shape.is_poly=" << shape.is_poly() << endl;  // <---- WRONG
    cout << "shape.is_polygonal=" << shape.is_polygonal() << endl;
    cout << "shape.is_polyhedral=" << shape.is_polyhedral() << endl;

    if(shape.is_poly())
    {
        cout << "We think the mesh is polygonal or polyhedral so make sides." << endl;

        try
        {
            // This dies b/c the input is not unstructured.
            conduit::Node s2dmap, d2smap;
            conduit::blueprint::mesh::topology::unstructured::generate_sides(n_topo,
                                                                             out["topologies/sides"],
                                                                             out["coordsets/side_coords"],
                                                                             s2dmap,
                                                                             d2smap);
            out.print();
        }
        catch(...)
        {
            cout << "Caught an exception." << endl;
        }
    }

    return 0;
}

```

BradWhitlock:
This PR includes some fixes for blueprint::utils::ShapeType so it can initialize from additional topology types. It also makes sure that the is_poly() method will return true only if the shape was initialized with polygonal or polyhedral shapes. The unstructured::generate_xxx functions now throw exceptions if their input topologies are not unstructured. I added tests for these changes.

This resolves #1217 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 
 #### Blueprint
 - The `conduit::blueprint::mpi::mesh::partition_map_back()` function was enhanced so it accepts a "field_prefix" value in its options. The prefix is used when looking for the `global_vertex_ids` field, which could have been created with a prefix by the same option in the `conduit::blueprint::mpi::mesh::generate_partition_field()` function.
+- The `conduit::blueprint::mesh::utils::ShapeType` class was enhanced so it can take topologies other than unstructured.
 
 ### Fixed
 
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -4672,6 +4672,11 @@ mesh::topology::unstructured::generate_lines(const Node &topo,
 {
     CONDUIT_ANNOTATE_MARK_FUNCTION;
 
+    if(topo.has_path("type") && topo["type"].as_string() != "unstructured")
+    {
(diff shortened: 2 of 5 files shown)


cyrush (approved):
thanks for fixing this @BradWhitlock

PR #1224, opened 2023-12-29
Add build_adjsets option to partitioner so user can skip adjset creation.

BradWhitlock:
The partitioner unconditionally tries to build adjsets. It even seems to try for topologies that have no adjsets (when there are other adjsets present). I added some logic to skip adjset creation when it appears a topology has no relevant adjsets. I also added a "build_adjsets" option to the partitioner options that lets the user outright skip adjset creation.


cyrush (approved):
Looks good, thanks for the improvements + extra slice testing

PR #1225, opened 2024-01-03
Fix for generate_partition_field so it works on polyhedral data.

BradWhitlock:
The generate_partition_field() function in Blueprint uses Parmetis to create a partition field that will be useful in decomposing a mesh. This function was not working correctly when passed polyhedral topologies since it was using the element faces as vertex ids. This fix changes the code so when polyhedral meshes are given, the function uses the sub_element vertex ids referenced by the element faces.

I added a test that makes a 2 domain polyhedral mesh and I verified that global vertex ids and a suitable parmetis decomposition are produced.


cyrush (approved):
thanks for this fix!

PR #1226, opened 2024-01-05
Unscramble face adjsets in to_topo() used in adjset validation program.

BradWhitlock:
The adjset validation program can output topologies for adjsets so they can be visualized in VisIt. Element adjsets were getting scrambled because the unstructured::points() function was used to extract an elements points. That function was always returning a unique+sorted list of the points, whereas in this case, we wanted the points in their original order. The unstructured::points() function can now accept a bool argument that lets us select whether we want to get the points unique+sorted or in their original order. I used the new argument to get points in their original order when making a new topology for an element adjset. This makes the validation program output surfaces they way they ought to look.


cyrush (approved):
looks good, unique vs non option is a useful feature

PR #1227, opened 2024-01-05
Make adjset::is_canonical more strict

BradWhitlock:
A host code was giving me adjset groups called group_0_1 and group_1_0 on domains 0 and 1, respectively. The canonical way of naming would be group_0_1. The compare_pointwise function used in the adjset validation tool calls is_canonical to determine whether it needs to rename adjset components but the function was not strict enough to detect this. Consequently, the tool would successfully extract a topology for one adjset group and not the other, causing the pointwise comparison to fail.

This change makes adjset::is_canonical check more of the group name to see if the numbers it contains are sorted. That fixes the pointwise comparison.


cyrush (approved):
looks good, thanks again!

PR #1228, opened 2024-01-11
Fix point_merge in partitioner.

BradWhitlock:
I found a case where a series of if statements were out of order in the algorithm that merges points in the partitioner. This could cause some duplicate points when combining domains, later leading to bad adjsets.


cyrush (approved):
Nice work tracking down the issue!

PR #1230, opened 2024-01-11
Matset Converter

JustinPrivitera:
Can convert between any of the three types: "full", "sparse_by_element", and "sparse_by_material".

Later I will add support for fields.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.hpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.hpp
@@ -743,11 +743,21 @@ namespace matset
     bool CONDUIT_BLUEPRINT_API is_material_dominant(const conduit::Node &matset);
 
     //-------------------------------------------------------------------------
-    void CONDUIT_BLUEPRINT_API convert_matset(const conduit::Node &src_matset,
-                                              conduit::Node &dest_matset,
-                                              const std::string &src_matset_type,
-                                              const std::string &dest_matset_type,
-                                              const float64 epsilon = CONDUIT_EPSILON);
+    bool CONDUIT_BLUEPRINT_API is_full(const conduit::Node &matset);
+
+    //-------------------------------------------------------------------------
+    void CONDUIT_BLUEPRINT_API to_full(const conduit::Node &src_matset,
+                                       conduit::Node &dest_matset);
+
+    //-------------------------------------------------------------------------
+    void CONDUIT_BLUEPRINT_API to_sparse_by_element(const conduit::Node &src_matset,
+                                                    conduit::Node &dest_matset,
+                                                    const float64 epsilon = CONDUIT_EPSILON);
+
+    //-------------------------------------------------------------------------
+    void CONDUIT_BLUEPRINT_API to_sparse_by_material(const conduit::Node &src_matset,
+                                                     conduit::Node &dest_matset,
+                                                     const float64 epsilon = CONDUIT_EPSILON);
 
     //-------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/libs/blueprint/conduit_blueprint_mesh.hpp
```diff
@@ -742,6 +742,13 @@ namespace matset
     //-------------------------------------------------------------------------
     bool CONDUIT_BLUEPRINT_API is_material_dominant(const conduit::Node &matset);
 
+    //-------------------------------------------------------------------------
+    void CONDUIT_BLUEPRINT_API convert_matset(const conduit::Node &src_matset,
+                                              conduit::Node &dest_matset,
+                                              const std::string &src_matset_type,
```
cyrush: can we detect src type instead of passing (folks could make a mistake)?

how about method sig variants in the public interface that look like
```
to_full
to_sparse_by_element
to_sparse_by_material
```

JustinPrivitera: Ok I have addressed this @cyrush

JustinPrivitera: what do you think now @cyrush

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -503,6 +503,382 @@ to_silo(const conduit::Node &field,
...
+    for (auto & mapitem : reverse_matmap)
+    {
+        const std::string &matname = mapitem.second;
+        vol_fracs[matname] = std::vector<double>(num_elems);
+    }
+
+    for (int elem_id = 0; elem_id < num_elems; elem_id ++)
+    {
+        int size = sizes[elem_id];
+        int offset = offsets[elem_id];
```
cyrush: I think we should use the O2M helper here to support general cases.

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -503,6 +503,382 @@ to_silo(const conduit::Node &field,
...
+    int_accessor offsets = matset["offsets"].value();
+    int num_elems = sizes.dtype().number_of_elements();
+
+    std::map<std::string, std::vector<double>> vol_fracs;
+    std::map<std::string, std::vector<int>> elem_ids;
+
+    for (int elem_id = 0; elem_id < num_elems; elem_id ++)
+    {
+        int size = sizes[elem_id];
+        int offset = offsets[elem_id];
```
cyrush: o2m helper?

JustinPrivitera:
I used the o2m iterator @cyrush

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -503,6 +503,381 @@ to_silo(const conduit::Node &field,
...
+    int num_elems = o2m_iter.elements(o2mrelation::ONE);
+
+    std::map<std::string, std::vector<double>> vol_fracs;
+    for (auto & mapitem : reverse_matmap)
+    {
+        const std::string &matname = mapitem.second;
+        vol_fracs[matname] = std::vector<double>(num_elems);
+    }
+
+    while (o2m_iter.has_next(o2mrelation::DATA))
```
cyrush, line 592: o2m looks good!

cyrush (approved):
thanks for the updates and refactoring

JustinPrivitera:
Fields done in #1234

PR #1234, opened 2024-01-23
Matset Field Converters

### LINKED ISSUE #1222: Generic matset converter
It would be nice to be able to swap between matset types easily.

full <--> sparse by element
full <--> sparse by material
sparse by element <--> sparse by material

Fields (and specsets?) would also need to be taken into account.

JustinPrivitera:
Resolves #1222

Follow on to #1230

Can convert between any of the three types: "full", "sparse_by_element", and "sparse_by_material". These types have different names.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
@@ -996,14 +996,16 @@ multi_buffer_by_material_to_multi_buffer_by_element(const conduit::Node &src_mat
         dest_field["values"].set(src_field["values"]);
     }
 
-    std::map<std::string, std::pair<double_array, int_array>> sbm_vol_fracs_and_elem_ids;
+    // sparse by material representation
+    // we map material names to volume fractions, element ids, and matset values
+    std::map<std::string, std::tuple<double_array, int_array, double_array>> sbm_rep;
 
     auto vf_itr = src_matset["volume_fractions"].children();
     while (vf_itr.has_next())
     {
         const Node &mat_vol_fracs = vf_itr.next();
         const std::string matname = vf_itr.name();
-        sbm_vol_fracs_and_elem_ids[matname].first = mat_vol_fracs.value();
+        std::get<0>(sbm_rep[matname]) = mat_vol_fracs.value();
     }
 
     auto eid_itr = src_matset["element_ids"].children();
@@ -1011,9 +1013,23 @@ multi_buffer_by_material_to_multi_buffer_by_element(const conduit::Node &src_mat
     {
         const Node &mat_elem_ids = eid_itr.next();
         const std::string matname = eid_itr.name();
-        sbm_vol_fracs_and_elem_ids[matname].second = mat_elem_ids.value();
+        std::get<1>(sbm_rep[matname]) = mat_elem_ids.value();
     }
 
+    bool mat_dep_field = src_field.has_child("matset_values");
+
+    if (mat_dep_field)
+    {
+        auto mvals_itr = src_field["matset_values"].children();
+        while (mvals_itr.has_next())
+        {
+            const Node &matset_vals = mvals_itr.next();
+            const std::string matname = mvals_itr.name();
+            std::get<2>(sbm_rep[matname]) = matset_vals.value();
+        }
+    }
+
+    // load the element ids into a set to find out how many there are
     auto determine_num_elems = [](const conduit::Node &elem_ids)
     {
         std::set<int> elem_ids_set;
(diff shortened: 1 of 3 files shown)


src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -504,71 +504,165 @@ to_silo(const conduit::Node &field,
...
     // set the topology
-    dest["topology"].set(matset["topology"]);
+    dest_matset["topology"].set(src_matset["topology"]);
+
+    if (src_field.dtype().is_object())
+    {
+        // copy over info from the old field
+        dest_field["association"].set(src_field["association"]);
+        dest_field["topology"].set(src_field["topology"]);
+        dest_field["matset"].set(src_field["matset"]);
```
cyrush: do we need to set this according to the new matset name?

Not sure we can always infere from passed dest_matset, so this might need to be an option?

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -504,71 +504,165 @@ to_silo(const conduit::Node &field,
...
+        dest_field["values"].set(src_field["values"]);
+    }
```
cyrush, line 529: Do we need to copy the values here? (Won't we over write later?)

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -504,71 +504,165 @@ to_silo(const conduit::Node &field,
...
-    dest["topology"].set(matset["topology"]);
+    dest_matset["topology"].set(src_matset["topology"]);
 
+    if (src_field.dtype().is_object())
+    {
+        // copy over info from the old field
+        dest_field["association"].set(src_field["association"]);
+        dest_field["topology"].set(src_field["topology"]);
+        dest_field["matset"].set(src_field["matset"]);
+        dest_field["values"].set(src_field["values"]);
```
cyrush, line 661: same qs as other case.

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -577,55 +671,121 @@ uni_buffer_by_element_to_multi_buffer_by_element(const conduit::Node &matset,
...
-    dest["topology"].set(matset["topology"]);
+    dest_matset["topology"].set(src_matset["topology"]);
+
+    if (src_field.dtype().is_object())
+    {
+        // copy over info from the old field
+        dest_field["association"].set(src_field["association"]);
+        dest_field["topology"].set(src_field["topology"]);
+        dest_field["matset"].set(src_field["matset"]);
+        dest_field["values"].set(src_field["values"]);
```
cyrush, line 786: same qs as other case.

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -634,109 +794,242 @@ uni_buffer_by_element_to_multi_buffer_by_material(const conduit::Node &matset,
...
-    while (mat_itr.has_next())
+    if (src_field.dtype().is_object())
     {
-        const Node &mat_vol_fracs = mat_itr.next();
-        std::string matname = mat_itr.name();
+        // copy over info from the old field
+        dest_field["association"].set(src_field["association"]);
+        dest_field["topology"].set(src_field["topology"]);
+        dest_field["matset"].set(src_field["matset"]);
+        dest_field["values"].set(src_field["values"]);
```
cyrush, line 910: same qs as other cases

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -757,57 +1050,121 @@ multi_buffer_by_material_to_multi_buffer_by_element(const conduit::Node &matset,
...
-    dest["topology"].set(matset["topology"]);
+    dest_matset["topology"].set(src_matset["topology"]);
+
+    if (src_field.dtype().is_object())
+    {
+        // copy over info from the old field
+        dest_field["association"].set(src_field["association"]);
+        dest_field["topology"].set(src_field["topology"]);
+        dest_field["matset"].set(src_field["matset"]);
+        dest_field["values"].set(src_field["values"]);
```
cyrush, line 1132: same qs as other cases

cyrush (commented):
looking good, two questions related to how we use src vals to init the des vals.

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
JustinPrivitera: The only thing we write later is `matset_values`. We do not overwrite the `values`.

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
JustinPrivitera: Do you think we should allow users to pass in a new matset name? If we're doing that, why not just let them change the matset name in the result field if they do end up naming their new matset something different?

cyrush: My worry is that the old name will always be wrong, so it would be better to have it passsd 

Here is an example of where using a name would allow us to make a single call to create the new field and matset in the same tree as the src data and have both be valid:

```

blueprint::mesh::field::to_sparse_by_element(mesh["matsets/matsrc"],
   mesh["fields/myfieldsrc"],
   "matdest",
   mesh["matsets/matdest"],
   mesh["fields/myfielddest"]);
```

JustinPrivitera: Ok, I'll add this functionality.

JustinPrivitera:
@cyrush is this all good to go then?

cyrush (approved):
Looks good, thanks for the changes!

PR #1235, opened 2024-01-25
Add `relay.io.silo` to the python interface

JustinPrivitera:


src/libs/conduit/python/py_src/relay/io/silo/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+###############################################################################
+# file: __init__.py
+# Purpose: Main init for the conduit relay io module.
+###############################################################################
+
+# silo support is optional, so drive on if we can't import
+try:
+    from .conduit_relay_io_silo_python import *
+except:
+    pass
```
JustinPrivitera, line 14: @cyrush how do you feel about this bonus try-catch?

None of the other things (MPI, Web, etc.) have logic like this. But it makes the windows CI pass.

cyrush: this is the one we need, you can remove the try catch case for silo (above)

src/libs/conduit/python/py_src/relay/io/__init__.py
```diff
@@ -10,5 +10,11 @@
 
 from . import blueprint
 
+# silo support is optional, so drive on if we can't import
```
cyrush: we can remove this guy and just have one for the compiled module.

PR #1236, opened 2024-01-30
Silo I/O choose matset type on read

JustinPrivitera:
Now when you read silo, you have the option to choose the kind of matset you want it to produce in blueprint.

The downside are all the intermediate matsets and fields that have to be created. Maybe there should be converters that just convert material dependent fields between the types? They still need a matset to do it, but maybe they don't have to produce a matset as a side effect.

Things to discuss:
- I have to save a copy of the sbe matset for the mat-dependent fields to use later when they get converted. Then I have to create temporary matsets for the field converters to use that just get thrown away.
- I have to create intermediate fields and intermediate matsets for all cases. Then I use expensive `set`s when I know what to do. This bothers me, but I'm not sure there's a clean way to do this. Maybe some ugly pointers + logic for setting them.
- Do I error if I get a bad option for the `matset_style`? Or just keep going?

### DIFF
--- a/src/libs/relay/conduit_relay_io_silo.cpp
+++ b/src/libs/relay/conduit_relay_io_silo.cpp
@@ -1541,9 +1541,6 @@ read_variable_domain(const int vartype,
         // collect the sparse by element matset we generated
         const Node &original_matset = matset_field_reconstruction["original_matset"];
 
-        // a throwaway to make the transform happy
-        Node dest_matset;
-
         if (opts_matset_style == "default" || opts_matset_style == "sparse_by_element")
         {
             field_out.move(intermediate_field);
@@ -1553,7 +1550,6 @@ read_variable_domain(const int vartype,
             conduit::blueprint::mesh::field::to_multi_buffer_full(original_matset,
                                                                   intermediate_field,
                                                                   matset_name,
-                                                                  dest_matset,
                                                                   field_out);
         }
         else // "multi_buffer_by_material"
@@ -1561,7 +1557,6 @@ read_variable_domain(const int vartype,
             conduit::blueprint::mesh::field::to_multi_buffer_by_material(original_matset,
                                                                          intermediate_field,
                                                                          matset_name,
-                                                                         dest_matset,
                                                                          field_out);
         }
     }
--- a/src/libs/relay/conduit_relay_io_silo.cpp
+++ b/src/libs/relay/conduit_relay_io_silo.cpp
@@ -1458,7 +1458,7 @@ read_variable_domain(const int vartype,
                      const std::string &multivar_name,
                      const std::string &bottom_level_mesh_name,
                      const std::string &volume_dependent,
-                     const std::string &opts_matset_type,
+                     const std::string &opts_matset_style,
                      const Node &matset_field_reconstruction,
                      Node &mesh_out)
 {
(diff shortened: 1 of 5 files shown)


src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -2484,6 +2540,27 @@ read_root_silo_index(const std::string &root_file_path,
...
+    // Get the selected matset flavor
+    if (opts.has_child("matset_style") && opts["matset_style"].dtype().is_string())
+    {
+        std::string opts_matset_style = opts["matset_style"].as_string();
+        if (opts_matset_style != "default" && 
+            opts_matset_style != "multi_buffer_full" &&
+            opts_matset_style != "sparse_by_element" &&
+            opts_matset_style != "multi_buffer_by_material")
+        {
+            CONDUIT_INFO("read_mesh invalid matset_style option: \""
```
JustinPrivitera: Fail if it is bad

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -1531,6 +1531,45 @@ read_variable_domain(const int vartype,
...
+            conduit::blueprint::mesh::field::to_multi_buffer_by_material(original_matset,
+                                                                         intermediate_field,
+                                                                         matset_name,
+                                                                         dest_matset,
+                                                                         field_out);
+        }
+    }
+    else
+    {
+        field_out.set(intermediate_field);
```
JustinPrivitera: use move operation instead

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -1747,14 +1786,31 @@ read_matset_domain(DBfile* matset_domain_file_to_use,
...
-    matset_out["offsets"].set(offsets.data(), offsets.size());
+    intermediate_matset["material_ids"].set(material_ids.data(), material_ids.size());
+    intermediate_matset["volume_fractions"].set(volume_fractions.data(), volume_fractions.size());
+    intermediate_matset["sizes"].set(sizes.data(), sizes.size());
+    intermediate_matset["offsets"].set(offsets.data(), offsets.size());
 
     matset_field_reconstruction["recipe"].set(field_reconstruction_recipe.data(), 
                                     field_reconstruction_recipe.size());
     matset_field_reconstruction["sizes"].set(sizes.data(), sizes.size());
+    matset_field_reconstruction["original_matset"].set(intermediate_matset);
```
JustinPrivitera: move operation

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -1747,14 +1786,31 @@ read_matset_domain(DBfile* matset_domain_file_to_use,
...
+
+    // create an entry for this matset in the output
+    Node &matset_out = mesh_out["matsets"][multimat_name];
+
+    if (opts_matset_style == "default" || opts_matset_style == "sparse_by_element")
+    {
+        matset_out.set(intermediate_matset);
```
JustinPrivitera: find a way to add logic to not copy the orig matset in the default case

JustinPrivitera:
In the fields namespace, remove the `dest_matset` from all the functions. Make new functions that only need to write fields and not write matsets. No need to have a function that writes both matsets and fields.

cyrush (approved):
Changes to use move look good!

cyrush (approved):

PR #1239, opened 2024-02-01
Improve Blueprint docs for strided structured topology.

BradWhitlock:
This update adds documentation about the strided structured topology and includes an example for how to create one. I also added a few points of clarification to the unibuffer matset docs.


cyrush (approved):
Great additions, thanks for taking time to do this case justice in our docs

PR #1253, opened 2024-02-07
Silo/Overlink I/O Polygon Support + Misc Fixes

JustinPrivitera:
Adds polygons to read and write
Clean up bookkeeping for variables
Refactor tests

### DIFF
--- a/src/libs/relay/conduit_relay_io_silo.cpp
+++ b/src/libs/relay/conduit_relay_io_silo.cpp
@@ -3865,6 +3865,8 @@ void silo_write_ucd_zonelist(DBfile *dbfile,
 
     const std::string zlist_name = (write_overlink ? "zonelist" : topo_name + "_connectivity");
 
+    n_mesh_info[topo_name]["zonelist_name"] = zlist_name;
+
     int silo_error =
         DBPutZonelist2(dbfile,             // silo file
                        detail::sanitize_silo_varname(zlist_name).c_str(), // silo obj name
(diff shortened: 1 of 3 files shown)


src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -3857,32 +3848,32 @@ void silo_write_ucd_zonelist(DBfile *dbfile,
...
 
-    std::string zlist_name = topo_name + "_connectivity";
+    const std::string zlist_name = (write_overlink ? "zonelist" : topo_name + "_connectivity");
 
     int silo_error =
         DBPutZonelist2(dbfile,             // silo file
                        detail::sanitize_silo_varname(zlist_name).c_str(), // silo obj name
                        total_num_elems,    // number of elements
-                       2,                  // spatial dims
+                       2,                  // spatial dims - TODO ?????
```
JustinPrivitera: What do spatial dims mean in the context of a zonelist? Is this just the dimension of the shape type? Does it matter?

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -3789,100 +3777,111 @@ void silo_write_ucd_zonelist(DBfile *dbfile,
...
+        shapetype.push_back(db_zonetype);
+        shapesize.push_back(num_pts);
+        shapecnt.push_back(num_elems);
+
         total_num_elems += num_elems;
+    };
+
+    // TODO should we allow people to write tris, wedges, and pyramids to overlink?
+    // I don't think overlink supports those shape types
+    // but we still produce valid silo
```
JustinPrivitera: What is the right call here?

src/libs/relay/conduit_relay_io_silo.cpp
cyrush: I think it is the dimension of the underlying coordinate space.  
2d coord system would be 2 for rz, or xy, etc
3d coord system would be 3 for xyz

src/libs/relay/conduit_relay_io_silo.cpp
cyrush: if the overlink validator would fail, we should raise an error b/c it;s not inline with overlink std.

In the the error message we can suggest they can use generic silo for this case

src/libs/relay/conduit_relay_io_silo.cpp
JustinPrivitera: Ok. I'll get that worked out.

I'm not sure if matters though - this has been hardcoded to 2 since before I worked on the silo i/o support. Changing it to 3 doesn't seem to have any effect that I can see. But probably a good call just in case downstream tools use the information.

src/libs/relay/conduit_relay_io_silo.cpp
JustinPrivitera: Ok sounds good.

cyrush (approved):
looks good!

PR #1254, opened 2024-02-08
Use o2mIndex for matset converters + refactor

### LINKED ISSUE #1238: Use O2M index for matset converters
https://github.com/LLNL/conduit/blob/develop/src/libs/blueprint/conduit_blueprint_o2mrelation_index.hpp

JustinPrivitera:
Resolves #1238

We can get this in to 0.9.1 if you'd like, or just punt on it until later. I think there is more I could add here that would clean things up.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
@@ -23,6 +23,9 @@
 #include "conduit_blueprint_mesh_utils.hpp"
 #include "conduit_blueprint_o2mrelation.hpp"
 #include "conduit_blueprint_o2mrelation_iterator.hpp"
+#include "conduit.hpp"
+#include "conduit_blueprint_exports.h"
+#include "conduit_blueprint_o2mrelation_index.hpp"
 
 using namespace conduit;
 // access conduit blueprint mesh utilities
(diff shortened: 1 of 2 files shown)


src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -503,6 +504,167 @@ to_silo(const conduit::Node &field,
...
+                      std::map<int, std::string> &reverse_matmap)
+{
+    // fill out map
+    auto matmap_itr = src_matset["material_map"].children();
+    while (matmap_itr.has_next())
+    {
+        const Node &matmap_entry = matmap_itr.next();
+        const std::string matname = matmap_itr.name();
+
+        reverse_matmap[matmap_entry.as_int()] = matname;
```
cyrush: to_int64 (as_int could fail)

cyrush: or `to_int`

cyrush (commented):
Looks good -- one more small type related suggestion

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
JustinPrivitera: ok done

cyrush (approved):

PR #1255, opened 2024-02-08
Change partitioner so it makes unique faces when combining PH meshes.

BradWhitlock:
This PR changes the partitioner so it makes unique faces when combining PH faces. The old code for some reason was appending face definitions.

The utils::ShapeType relies on embedding arrays to make the "faces" that are in each shape. For wedge and pyramid, these make a set of triangles for what should be quad faces. Splitting quad faces into triangles is undesirable when combining PH meshes with other shape types. I ran across this in examining some of the partitioner baselines. I added a get_face() method to ShapeType that preserves the quad/tri faces for pyramids and wedges. The get_face() method is used in the partitioner to traverse faces of 3D shape types when we need to mix them into a PH mesh.


cyrush (approved):
looks great, thanks for the improvements!

PR #1256, opened 2024-02-13
Use Data Accessor `set` to clean up generate sides further

### LINKED ISSUE #1243: Refactor Generate Sides code to take advantage of `data_accessor` set
Now that #1143 is complete, we can finish cleaning up generate_sides templating. See #1138 for more info and discussion.

JustinPrivitera:
Resolves #1243

Thanks to #1143, we can clean up `generate_sides` even more.

How does this look @cyrush?

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -5115,20 +5115,19 @@ namespace detail
         }
     }
 
-    template<typename T> // T is the type of the new field values
     void
     vertex_associated_field(const Node &topo_dest,
                             float64_accessor poly_field_data,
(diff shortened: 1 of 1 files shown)


BradWhitlock:
What's the performance difference? I'm assuming it's slower.

JustinPrivitera:
@BradWhitlock I will do some timings and report back.

JustinPrivitera:
The timings are just about the same. We have a maybe 0.5% speedup with the changes I've made.

I think it's about the same because we were doing a whole lot of static casts before; not much has changed other than the way we do those casts.

cyrush:
thanks @JustinPrivitera  - data accessor is using the same basic strategy as the templating here, its just packaged into a simpler interface that is compiled once.

PR #1259, opened 2024-02-16
remove use of distutils in python cmake introspection logic

cyrush:


cyrush:
needed to support python 3.12

cyrush:
Trying to replace:
```
import sys;from distutils.sysconfig import get_python_lib;sys.stdout.write(get_python_lib())
```

The following do not work on ubunutu:
```
import sys;from sysconfig import get_path;sys.stdout.write(get_path('platlib'))
import sys;from sysconfig import get_path;sys.stdout.write(get_path('purelib'))
```

`get_python_lib()` points to a `dist-packages` dir (specific choice to debian + ubunutu)
however: `get_path('platlib') and get_path('purelib')` return a path to a non-existent site-packages dir


More info:

https://stackoverflow.com/questions/122327/how-do-i-find-the-location-of-my-python-site-packages-directory

PR #1260, opened 2024-02-22
Improve PH element extraction in partitioner

BradWhitlock:
After finding that recombined PH domains get some faces wrong, I thought about why that could happen. It turns out that when the subset of PH zones are pulled out of their original mesh, they were grabbing their faces as-is. For zones that had a face that was defined first by another zone, it was necessary to reverse the face first before adding it into the new mesh subset. This change means that the normals are pointing in a more consistent direction (assuming the input mesh was well-behaved to begin with) when extracted chunks are recombined. I had to fix some partitioner baselines where faces got reversed, and I verified they are better now.


src/libs/blueprint/conduit_blueprint_mesh_partition.cpp
```diff
@@ -3352,55 +3352,86 @@ Partitioner::unstructured_topo_from_unstructured(const conduit::Node &n_topo,
...
         std::map<index_t,index_t> old2new_faces;
         std::vector<index_t> new_sizes, new_offsets, new_se_conn, new_se_sizes, new_se_offsets;
 
         index_t new_offset = 0, new_se_offset = 0;
         for(auto eid : element_ids)
         {
-            auto offset = static_cast<index_t>(offsets[eid]);
-            auto nfaces = static_cast<index_t>(sizes[eid]);
+            const auto offset = static_cast<index_t>(offsets[eid]);
+            const auto nfaces = static_cast<index_t>(sizes[eid]);
```
cyrush: i think you can drop these static casts

cyrush (approved):
Looks good, small suggestion about dropping static casts that I am not sure we need

src/libs/blueprint/conduit_blueprint_mesh_partition.cpp
BradWhitlock: Good point.

PR #1261, opened 2024-03-06
Changed partition_map_back() so it tries to reuse node memory.

BradWhitlock:
This change makes partition_map_back() attempt to reuse the original field memory that it's mapping values back into rather than resetting the node. If the field is not allocated or won't fit then allocation does take place. This change makes it work better with source meshes whose fields were provided using set_external(), as in host codes.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh_partition.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_partition.cpp
@@ -8656,15 +8656,7 @@ determine_schema(const Node &in,
         for(index_t i = 1; i < num_children; i++)
             same_types = (in[i].dtype().id() == in[0].dtype().id());
 
-        // Check whether components appear interleaved.
-        bool interleaved = true;
-        for(index_t i = 1; i < num_children; i++)
-        {
-            const auto dI = in[i].dtype().element_index(0) - in[i - 1].dtype().element_index(0);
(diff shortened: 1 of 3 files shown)


src/libs/blueprint/conduit_blueprint_mesh_partition.cpp
```diff
@@ -8635,33 +8638,104 @@ determine_schema(const Node &in,
...
+        for(index_t i = 1; i < num_children; i++)
+            same_types = (in[i].dtype().id() == in[0].dtype().id());
+
+        // Check whether components appear interleaved.
+        bool interleaved = true;
+        for(index_t i = 1; i < num_children; i++)
+        {
+            const auto dI = in[i].dtype().element_index(0) - in[i - 1].dtype().element_index(0);
+            interleaved &= (dI == in[i].dtype().element_bytes());
+        }
```
cyrush: can mcarray is_interleaved helper work here?

https://github.com/LLNL/conduit/blob/develop/src/libs/blueprint/conduit_blueprint_mcarray.cpp#L252

BradWhitlock: Yes. Thanks. I made the change.

cyrush (approved):
Looks good, thanks!

PR #1262, opened 2024-03-08
Add partition options to set field names for original_element_ids, original_vertex_ids.

BradWhitlock:
When partitioning multiple topologies is can be helpful to be able to specify the names of the fields for original_element_ids, original_vertex_ids so all topologies and their respective "original" fields can all live in the same Conduit node.


cyrush (approved):
thanks!

PR #1263, opened 2024-03-15
Fix single component mcarray partitioning.

BradWhitlock:
An earlier change I made broke sending single component mcarrays through the partitioner when it needs to recombine domains into fewer domains. This PR fixes that mistake and adds a test.


cyrush (approved):
Looks good, thanks for the update

PR #1267, opened 2024-03-27
generate_sides robust to no fields input

### LINKED ISSUE #1265: generate_sides needs w/ fields output needs to be robust to no fields input
This will fail if the source mesh lacks fields:

https://github.com/LLNL/conduit/blob/0f8c890c689b885021327715bc83469282518c20/src/libs/blueprint/conduit_blueprint_mesh.cpp#L5518

JustinPrivitera:
resolves #1265


cyrush (approved):
Awesome, thanks!

PR #1268, opened 2024-04-05
Fix issue with MPI tags on some MPI distributions.

BradWhitlock:
Some blueprint algorithms use MPI tags to help indicate intent of a message and they may use large numbers for the tags. It turns out that some MPI distributions do not like large tag values. This PR changes a communication helper class in relay so it passes tags through a safe_tag() function that clamps tags at MPI_TAG_UB so they work with MPI rather than having MPI abort.


cyrush (approved):
Nice detective work! Thanks for resolving this issue

PR #1269, opened 2024-04-12
Fix memory problems in flatten test

BradWhitlock:
I ran the Conduit tests after having built with the Intel compiler and the flatten test failed again in a new way. I ran through valgrind and it found out of bounds reads for the braid _quads_and_tris_ dataset. When I printed out that dataset, which gets used in flatten to make a 4x3 node mesh, it had some node ids that were larger than 11. The routine that made the connectivity was bad.

* I fixed the problem connectivity and now valgrind shows no errors in the flatten test
* I had to update a few partition baselines due to the connectivity changes
* I fixed some warnings


BradWhitlock:
Wait a minute. I forgot to re-enable a test that started misbehaving.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -1820,16 +1818,14 @@ braid_quads_and_tris_offsets(index_t npts_x,
...
+        for(int32 i = 0; i < nele_x; i++)
         {
-            int32 n1 = yoff + i;
-            int32 n2 = n1 + (nele_x+1);
-            int32 n3 = n1 + 1 + (nele_x+1);
-            int32 n4 = n1 + 1;
+            int32 n1 = j * npts_x + i;
+            int32 n2 = j * npts_x + i + 1;
+            int32 n3 = (j+1) * npts_x + i + 1;
+            int32 n4 = (j+1) * npts_x + i;
```
cyrush, line 1829: thanks for finding these mistakes

BradWhitlock: Sure thing. It's all set now.

cyrush (approved):
thanks again!

PR #1273, opened 2024-04-17
Strided Structured Column Major Test Case

JustinPrivitera:
Prove that we can handle this case successfully with blueprint and downstream in visit.


cyrush (approved):
looks good -- thanks for adding this example

PR #1274, opened 2024-04-18
Silo I/O: Read All Kinds of Silo Data

JustinPrivitera:
- Enable reading of various kinds of silo data, so we are no longer restricted to multimeshes, multivars, multimats, etc.
- handle column major data for many cases
- handle min/max index for many cases


JustinPrivitera:
TODO
- ~~[ ] add test for the various read options~~
- [x] get guidance from mark about pointvar cases - stride min/max index and major order
- [x] investigate overcomplicated logic for deciding what to read

JustinPrivitera:
@cyrush I can split this into two if you'd like, if that would be more digestible. I know how I would split it. Let me know what you think is best.

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1781,12 +1786,60 @@ TEST(conduit_relay_io_silo, read_silo)
         {
             remove_path_if_exists(out_name + "_write_overlink");
             write_opts["file_style"] = "overlink";
-            write_opts["ovl_topo_name"] = meshname;
+            write_opts["ovl_topo_name"] = meshname; // TODO do I even need this
```
cyrush, line 1789: Good Q: Does overlink ignore this?

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1781,12 +1786,60 @@ TEST(conduit_relay_io_silo, read_silo)
...
+        {
+            remove_path_if_exists(out_name + "_write_silo");
+            io::silo::save_mesh(load_mesh, out_name + "_write_silo");
+
+            // overlink requires matsets
+            if (load_mesh[0].has_child("matsets"))
+            {
+                remove_path_if_exists(out_name + "_write_overlink");
+                write_opts["file_style"] = "overlink";
+                write_opts["ovl_topo_name"] = "MMESH"; // TODO do I even need this
```
cyrush, line 1836: Does overlink ignore this (it always has to be MMESH?)

cyrush (approved):

src/tests/relay/t_relay_io_silo.cpp
JustinPrivitera: I think specifying the topo name for overlink is only necessary when there are multiple topos. Since we only read one topo there is only one to output, so I think this should be unnecessary.

src/tests/relay/t_relay_io_silo.cpp
JustinPrivitera: MMESH is actually the name of the blueprint topo in this case, since we read an overlink file into blueprint.

src/tests/relay/t_relay_io_silo.cpp
JustinPrivitera: I'm going to merge this and tweak the tests later; there are other fixes to do in the tests.

cyrush: I was asking b/c these are write opts?

JustinPrivitera:
```
///  In the overlink case...
///   1) We have made the choice to output only ONE topology as a multimesh.
///   2) mesh_name is ignored if provided and changed to "MMESH"
///   3) ovl_topo_name is the name of the topo we are outputting. If it is not
///      provided, we choose the first topology in the blueprint.
```

JustinPrivitera: and yes those are write opts

PR #1278, opened 2024-05-13
Move H5Sclose call to proper scope

nselliott:
Follow-up to #1277 , the `H5SClose(dataspace)` call was in the `else` block of an if-else statement, so the object was never closed in the `if (true)` case.


cyrush (approved):
Thanks!

PR #1296, opened 2024-07-12
add gyre example and high level mesh bp examples driver

cyrush:


src/libs/blueprint/conduit_blueprint_mesh_examples_generate.cpp
```diff
@@ -0,0 +1,642 @@
...
+        basic(mesh_type,nx,ny,nz,res);
+    }
+    else if(example_name == "strided_structured")
+    {
+        conduit::index_t    npts_x = detail::fetch_index_t(opts,example_name,"npts_x");
+        conduit::index_t    npts_y = detail::fetch_index_t(opts,example_name,"npts_y");
+        conduit::index_t    npts_z = detail::fetch_index_t(opts,example_name,"npts_z",0); // optional
+
+        // desc is an output?
+        conduit::Node desc;
```
JustinPrivitera, line 212: I don't think it is an output. Looking at the tests it looks like it is an optional input. Might want to consider how users could provide `desc` for this case when using this generate infrastructure.

src/tests/blueprint/python/t_python_blueprint_mesh.py
```diff
@@ -198,6 +198,60 @@ def test_strided_structured(self):
...
+                  "related_boundary",
+                  "rz_cylinder",
+                  "tiled",
+                  "venn"]
+    
+        for ename in enames:
+            opts = conduit.Node()
+            blueprint.mesh.examples.generate_default_options(opts,ename)
+            print(ename)
+            print(opts)
```
JustinPrivitera, line 228: we don't want to baseline these?

src/tests/blueprint/python/t_python_blueprint_mesh.py
```diff
@@ -198,6 +198,60 @@ def test_strided_structured(self):
...
+                  "related_boundary",
+                  "rz_cylinder",
+                  "tiled",
+                  "venn"]
+        for ename in enames:
+            n    = conduit.Node()
+            info = conduit.Node()
+            blueprint.mesh.examples.generate(ename,n)
+            self.assertTrue(blueprint.mesh.verify(n,info))
+
```
JustinPrivitera, line 254: same q here

src/tests/blueprint/t_blueprint_mesh_examples.cpp
```diff
@@ -1421,6 +1421,26 @@ TEST(conduit_blueprint_mesh_examples, polychain)
...
+        blueprint::mesh::examples::gyre(25,25,2,t,res);
+        Node info;
+        EXPECT_TRUE(blueprint::mesh::verify(res,info));
+        t+= dt;
+        std::string ofbase = conduit_fmt::format("gyre_{:04d}",i);
+        test_save_mesh_helper(res,ofbase);
+    }
+
+}
+
```
JustinPrivitera, line 1443: I guess we don't ever baseline the examples... I feel like that would be a good thing though.

JustinPrivitera (approved):

PR #1304, opened 2024-07-19
resolve a few hdf5 asan issues related to cleanup during exceptions

cyrush:


cyrush:
TIL:

Building HDF5 in Debug mode will abort at process exec end if hdf5 fails to fully close all handles.

Our tests poke at hdf5 error conditions that leave hdf5 in a funky state.

You can use `-DNDEBUG` to avoid abort.

cyrush:
having a hard time suppressing mpi related leaks, here is an example:
```
==33334==ERROR: LeakSanitizer: detected memory leaks
2024-07-19T23:52:18.8699211Z Direct leak of 136 byte(s) in 1 object(s) allocated from:
2024-07-19T23:52:18.8700130Z     #0 0x558ff2f0bc33 in __interceptor_strdup (/home/runner/work/conduit/conduit/build/tests/blt_mpi_smoke+0x8bc33) (BuildId: 0b49975eba2a2bf13f82e575d76e271ff1254609)
2024-07-19T23:52:18.8700984Z     #1 0x7fc80cd6dd88  (<unknown module>)
2024-07-19T23:52:18.8701321Z     #2 0x502f6e6f72656c64  (<unknown module>)
```

Not sure I can just squash (<unknown_module>) ?

https://github.com/google/sanitizers/issues/1731

In our case we are building all relevant TPLS with ASAN, so it might be ok to squash these unknown module + strdup.

In my containers, I can suppress PMPI_Init to solve these, but the github env is a little different. 
Maybe just need to run in a custom container.

PR #1312, opened 2024-08-05
support YAML schema

JustinPrivitera:
There is a bit more to do but this is a good place to do a merge; this has gotten large enough already.

Follow on work: #1313

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,6 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 
 #### Conduit
 - Removed cmake use of distutils for python introspection.
-- Deprecated length in json schema.
 
 #### Blueprint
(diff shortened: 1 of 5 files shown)


cyrush:
```
 error C2429: language feature 'terse static assert' requires compiler flag '/std:c++17
```
Windows is unhappy b/c of fancy c++ :-)

JustinPrivitera:
> ```
>  error C2429: language feature 'terse static assert' requires compiler flag '/std:c++17
> ```
> 
> Windows is unhappy b/c of fancy c++ :-)

:(

JustinPrivitera:
> Windows is unhappy b/c of fancy c++ :-)

Actually it's my fault; gcc let me get away with something that I shouldn't be allowed to do.

cyrush (approved):
Looks good!

PR #1327, opened 2024-10-10
Fixes for tags that affected PointQuery on some systems.

BradWhitlock:
On rzadams, the MPI does not like large tag values. The PointQuery algorithm (and others) were using large tag values. The mechanism put in place to query a communicator's upper tag limit was not quite right before. This PR updates some of that logic and lowers some tag values so they're more likely to work with that MPI.

Also, in relay::mpi::communicate_using_schema it was possible for a user-registered error function to prevent exceptions from being thrown when needed. The class now temporarily suspends user-registered warning/error functions so exceptions will get thrown if MPI errors or other errors occur. This should make errors easier to diagnose.

### DIFF
--- a/src/libs/relay/conduit_relay_mpi.cpp
+++ b/src/libs/relay/conduit_relay_mpi.cpp
@@ -279,12 +279,12 @@ bool invalid_tag(int tag)
         implementation-dependent. If the tag is not in that range, return
         the value for MPI_TAG_UB so it is safe to use with MPI functions.
 
- @param comm The MPI communicator.
  @param tag The input tag.
+ @param comm The MPI communicator.
 
  @return A tag value that is safe to use with MPI.
  */
-int safe_tag(MPI_Comm comm, int tag)
+int safe_tag(int tag, MPI_Comm comm)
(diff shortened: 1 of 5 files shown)


src/libs/relay/conduit_relay_mpi.cpp
```diff
@@ -277,15 +277,35 @@ bool invalid_tag(int tag)
...
  @param tag The input tag.
 
  @return A tag value that is safe to use with MPI.
  */
-int safe_tag(int tag)
+int safe_tag(MPI_Comm comm, int tag)
 {
-    return (tag < MPI_TAG_UB) ? ((tag >= 0) ? tag : MPI_TAG_UB) : MPI_TAG_UB;
+    // Get the tag upper bound for the communicator.
+    // MPI_Comm_get_attr with MPI_TAG_UB is not a very reliable function.
```
aaroncblack: That's always fun :(  Thanks for identifying this issue.

cyrush (approved):
Looks good, thanks for these improvements!

PR #1328, opened 2024-10-18
Document ways of accessing numerical data

agcapps:
Add material to tutorial_cpp_numeric.rst that compares and contrasts three methods of access to numeric array data:
* direct pointer access
* DataArray
* DataAccessor


cyrush:
This looks great!

The output of the docs is not automatically captured, so if we want to update the output to include the new example, we need to run the test and then pipe to `https://github.com/LLNL/conduit/blob/develop/src/docs/sphinx/t_conduit_docs_tutorial_numeric_out.txt`  and commit the change.

cyrush (approved):
Thanks for the update to commit the test text out.
I think that was all we needed, sorry I missed the updates until now!

PR #1329, opened 2024-10-19
Add yyjson support to avoid rapidjson on Debian

mzf-guest:
Conduit uses rapidjson as thirdparty but its license is not compatible with Debian Free Software Guideline because of the additional clause "The Software shall be used for Good, not Evil.". More information here: https://wiki.debian.org/qa.debian.org/jsonevil

As I'm working on packaging conduit for Debian, this Pull Request suggests to use [yyjson](https://github.com/ibireme/yyjson) as a replacement.
I've implemented a thin wrapper to call yyjson throught the rapidjson API, in order to keep the two json backends working.
The wrapper is defined in  `src/thirdparty_builtin/yyjson/conduit_yyjson.h` and the CMake variable `USE_SYSTEM_YYJSON` enables yyjson.

All non-regression tests are passing on my machine with yyjson and rapidjson, but it likely needs additional testing with real cases.

Let me know if you agree this way to push conduit to Debian, and feel free to advise/modify/contribute!

Best,
François

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -40,6 +40,11 @@ option(ENABLE_FORTRAN     "Build Fortran Support"       OFF)
 option(ENABLE_MPI         "Build MPI Support"           OFF)
 option(ENABLE_OPENMP      "Build OpenMP Support"        OFF)
 
+option(USE_SYSTEM_YYJSON "Use yyjson as a replacement for rapidjson." OFF)
+if(${USE_SYSTEM_YYJSON})
+    add_compile_definitions(USE_YYJSON)
+endif()
+
 # Add another option that provides extra 
 # control over conduit tests for cases where 
 # conduit is brought in as a submodule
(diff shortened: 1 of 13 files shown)


cyrush:
@mzf-guest this is great!

Thanks for bringing this to my attention.

We are on a very old version of rapidjson & I am open to replacing with yyjson.

I need a little more time to check our the project, I'll get back to you soon.

cyrush:
@mzf-guest 

I would like to include yyjson as a vendored TPL (just like rapidjson)

Then the option can be:

```
ENABLE_YYJSON = default to off, to keep rapidjson as the default. 
```

Over time we would switch off rapidjson, but still need it as an option if there is a pick compiler somewhere.

cyrush:
@mzf-guest  Update: We want to switch fully over to yyjson (vendored/builtin). 

You work to abstract the interface is a big big help to this process.  I am going to merge your work in to a branch and work on the next steps.

PR #1334, opened 2024-11-16
Avoid relinking python modules on every rebuild

bradking:
Avoid rerunning `conduit_python_py_setup`'s `pip install` command on every invocation of `make` or `ninja` by recording its last execution time with an output time stamp file.  Whenever it does rerun, such as when an input file changes, relink dependent python modules since the `pip install` command wipes out its `--target` directory.  This can be achieved by making `conduit_python_py_setup` an INTERFACE library that propagates the time stamp file as a link dependency of dependent python modules through `INTERFACE_LINK_DEPENDS`.


cyrush:
@bradking thank you very much for this solution, it will be nice to have this cleaned up.

cyrush (approved):

PR #1337, opened 2024-12-19
Silo/Overlink I/O Careful Handling of Names

JustinPrivitera:
- fail when names are not alphanumeric instead of renaming things
- prevent name collisions
- new test to check that name collisions are avoided
- refactor generate_silo_names
- misc fixes


cyrush (approved):
Thanks for these change!

PR #1341, opened 2025-01-07
New Matset Utilities

JustinPrivitera:
Pretty self explanatory:

```
count_zones_from_matset

is_material_in_zone

create_reverse_material_map
```

Will be useful for silo specsets support/testing


cyrush (approved):
thanks for adding these, the could also help us define an interface layer in the future

PR #1347, opened 2025-01-10
specset::to_silo rewrite

JustinPrivitera:
handle the case where not all materials have species, use some silo optimizations, tests for many strange cases.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
@@ -1748,36 +1748,17 @@ to_silo(const conduit::Node &specset,
             return sum;
         }();
 
-        // WARNING: this can produce indices that are out of bounds in specific cases:
-        // example:
-        //   nmatspec: [0, 2, 2, 0]
-        //   specnames: 
-        //     - "a_spec1"
-        //     - "a_spec2"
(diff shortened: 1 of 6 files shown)


src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -1711,42 +1751,54 @@ to_silo(const conduit::Node &specset,
...
-        // I think the point of this optimization is to also leave stuff out of the 
-        // species_mf array. If I want to do this then I should explore that as well.
+        // This can produce an out of bounds index in very specific cases.
+        // If a material has no species, the index produced by this function is 
+        // useless, but downstream data consumers shouldn't be reading the index
+        // anyway. If a material has no species and it is the last one in the 
+        // material map and the final zone is mixed and contains that material,
+        // then we can get an index that is out of bounds. This is ok because 
+        // downstream tools like VisIt read based on the number of species, so
+        // even though the index is garbage it goes unused.
```
cyrush, line 1765: Add to comment that Silo requires a slot for this data because the buffer indexing needs be consistent with the material indices? (Is that the right way to say it?)

JustinPrivitera: note included below

Issue #811, opened 2021-08-06
wire silo read mesh to blueprint read mesh


cyrush:
Implementation will be exposed via:

`conduit::relay::io::silo::read_mesh()`

But we also want 
`conduit::relay::io::blueprint::read_mesh()`

to be able to read overlink meshes 

(this will be the unified interface folks call)

cyrush:
resolved in #1356

PR #1356, opened 2025-01-21
add helper to silo detect files

cyrush:


src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -72,6 +72,34 @@ TEST(conduit_relay_io_silo, conduit_silo_cold_storage_generic_iface)
...
+    // make sure blueprint hdf5 file doesn't return true
+    n_test.reset();
+    blueprint::mesh::examples::braid("uniform", 5, 5, 0, n_test);
+    io::blueprint::save_mesh(n_test,"tout_bp_hdf5_mesh_vs_silo_open","hdf5");
+    EXPECT_FALSE(io::silo::is_silo_file("tout_bp_hdf5_mesh_vs_silo_open.root"));
+
+    std::string silo_input_file = utils::join_file_path("overlink", "box2d.silo");
+    silo_input_file = relay_test_silo_data_path(silo_input_file);
+    // make sure known silo file *does* return true
+    EXPECT_TRUE(io::silo::is_silo_file(silo_input_file));
```
JustinPrivitera: For completeness it might be good to have it try to open a Silo file written with the PDB driver.

JustinPrivitera (approved):

src/tests/relay/t_relay_io_silo.cpp
cyrush: Do we have any examples in already in our repo?

JustinPrivitera: As far as I know, we do not. However, we can create them using the Silo Writer.

```
Node opts;
opts["silo_type"] = "pdb";
Node save_mesh;
blueprint::mesh::examples::basic("rectilinear", 3, 4, 0, save_mesh);
const std::string basename = "gimme_pdb";
io::silo::save_mesh(save_mesh, basename, opts);
```

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -72,6 +72,51 @@ TEST(conduit_relay_io_silo, conduit_silo_cold_storage_generic_iface)
...
+
+    // make sure blueprint hdf5 file doesn't return true
+    n_test.reset();
+    blueprint::mesh::examples::braid("uniform", 5, 5, 0, n_test);
+    io::blueprint::save_mesh(n_test,"tout_bp_hdf5_mesh_vs_silo_open","hdf5");
+    EXPECT_FALSE(io::silo::is_silo_file("tout_bp_hdf5_mesh_vs_silo_open.root"));
+
+    // make sure blueprint yaml file doesn't return true
+    io::blueprint::save_mesh(n_test,"tout_bp_yaml_mesh_vs_silo_open","yaml");
+    EXPECT_FALSE(io::silo::is_silo_file("tout_bp_yaml_mesh_vs_silo_open.root"));
```
cyrush: added a few more test cases, including explicit `pdb` case

JustinPrivitera (approved):

cyrush:
Something we need to resolve:

A failed open with Silo -  seems to undermine opening a file with hdf5.

cyrush:
To avoid echoed errors, we can use:
```
DBShowErrors(NULL,NULL);
```
We should push/pop the silo handlers.

cyrush:
if hdf5, we can ask hdf5 directly if:

`_silolibinfo`

exists to see if we are hdf5 + silo.

cyrush:
@JustinPrivitera  I added more smarts to the detection logic and learned how to spell Suppressor. Can you take a look at this?

There is some duplication with the silo specific detection implementation and the logic in identify file format.
This is b/c I want to minimize file opens during detection.  Calling the silo check specific to hdf5 would result in one extra open.

I am also going to add a check for the magic number so we can avoid any silo opens when we are sure it is not a pdb file.

src/libs/relay/conduit_relay_io_identify_protocol.cpp
```diff
@@ -159,12 +193,14 @@ identify_file_type(const std::string &path,
                 if(test_str.find("{") != std::string::npos)
                 {
                    file_type = "json";
-                }
```
JustinPrivitera: we lost the close brace

src/libs/relay/conduit_relay_io_identify_protocol.cpp
```diff
@@ -159,12 +193,14 @@ identify_file_type(const std::string &path,
                 if(test_str.find("{") != std::string::npos)
                 {
                    file_type = "json";
```
JustinPrivitera, line 209: we still have this issue no? #1291

src/libs/relay/conduit_relay_io_identify_protocol.cpp
```diff
@@ -119,38 +124,67 @@ identify_protocol(const std::string &path,
...
 {
     file_type = "unknown";
-
-    // goal: check for: hdf5, json, or yaml
-
-    // first check for hdf5
-#ifdef CONDUIT_RELAY_IO_HDF5_ENABLED
-    if(conduit::relay::io::is_hdf5_file(path))
+    std::string hdf5_magic_number = "\211HDF\r\n\032\n";
+    std::string pdb_magic_number  = "<<PDB:";
```
JustinPrivitera: I would make these const 🙄

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -227,13 +278,164 @@ void silo_read(DBfile *dbfile,
...
+//---------------------------------------------------------------------------//
+bool
+is_silo_file(const std::string &file_path, const std::string &silo_driver)
+{
+    SiloErrorSuppressor ses;
+
+    bool res = false;
+    if(silo_driver == "hdf5")
+    {
+        std::string hdf5_magic_number = "\211HDF\r\n\032\n";
```
JustinPrivitera: I would make this const

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -227,13 +278,164 @@ void silo_read(DBfile *dbfile,
...
+silo_open_file_for_read(const std::string &file_path)
+{
+    bool res = false;
+    // Note: we use DB_UNKNOWN here b/c we expect the common case to be 
+    // where we are testing a non-silo file. For this case, we don't want
+    // a cascade of tries at opening.
+    //
+    // For silo_open_file_for_read(), we make a different choice
+    // b/c for that API call, we expect that in most cases the path
+    // passed will be an hdf5 flavored silo file
```
JustinPrivitera: I don't understand this comment - it says we are using DB_UNKNOWN here b/c we expect to be calling the mpi version of `silo_open_file_for_read` on non-silo files. Why? Then the next comment says that for `silo_open_file_for_read`, we make a different choice. But this function _is_ `silo_open_file_for_read`, so I don't know what you mean.

src/tests/blueprint/t_blueprint_mesh_relay.cpp
```diff
@@ -845,3 +845,45 @@ TEST(conduit_blueprint_mesh_relay, sparse_topos)
...
+        std::string output_dir  = output_base + ".cycle_000000";
+        std::string output_root = output_base + ".cycle_000000.root";
+        relay::io::blueprint::write_mesh(data, output_base, protocol);
+
+        // make sure we can load back, this tests the auto detection
+        // of the file type
+        // read the mesh back in diff to make sure we have the same data
+
+        Node n_read, info;
+        relay::io::blueprint::read_mesh(output_root,n_read);
```
JustinPrivitera: I guess you still need to diff?

JustinPrivitera (changes requested):

src/libs/relay/conduit_relay_io_identify_protocol.cpp
cyrush: this resolves that issue. It strips out any quoted strings in yaml or json, and then only looks for `{` in the result

src/libs/relay/conduit_relay_io_silo.cpp
cyrush: yes, this is the old implementation -- we are no longer using this to test for the silo file, I need to update this.

cyrush: hmm, is this still there, I think I removed it?

Here is the current implementation:


```
DBfile *
silo_open_file_for_read(const std::string &file_path)
{
    // this open cascade is an optimization -- we expect most open cases
    // will need hdf5 driver, DB_UNKNOWN has more logic that is slightly
    // more expensive.

    DBfile *res = DBOpen(file_path.c_str(), DB_HDF5, DB_READ);
    if(res != NULL)
    {
        return res;
    }

    res = DBOpen(file_path.c_str(), DB_PDB, DB_READ);
    if(res != NULL)
    {
        return res;
    }

    res = DBOpen(file_path.c_str(), DB_UNKNOWN, DB_READ);
    if(res != NULL)
    {
        return res;
    }

    return res;
}

//------------

File: src/libs/conduit/conduit_utils.cpp [hunk 9]
```diff
@@ -932,6 +932,41 @@ check_num_char(const char v)
     return res;
 }
 
+//---------------------------------------------------------------------------//
+std::string
+strip_quoted_strings(const std::string &input, const std::string &quote_char)
+{
+    std::string res;
+    bool in_string = false;
+    bool emit = false;
```
JustinPrivitera, line 941: do you mean "omit"? What does "emit" mean?

src/libs/conduit/conduit_utils.cpp
```diff
@@ -932,6 +932,41 @@ check_num_char(const char v)
...
+        {
+            if(in_string)
+            {
+                in_string = false;
+                emit = false;
+            }
+            else
+            {
+                in_string = true;
+            }
```
JustinPrivitera: what is going on here?

"If it is in the string, then it is not in the string. If it is not in the string, then it is in the string."

src/libs/conduit/conduit_utils.cpp
JustinPrivitera: the character gets emitted?

src/libs/conduit/conduit_utils.cpp
JustinPrivitera: I guess this is saying that when we hit a quote_char, if we were in the quoted substring already then now we are exiting, and if we weren't then now we are entering. I would put comments to make it easier to understand.

src/libs/relay/conduit_relay_io_identify_protocol.cpp
JustinPrivitera: looks good

src/libs/relay/conduit_relay_io_silo.cpp
cyrush: there was an old implementation, I removed it.
github wasn't linking me to the right part of the source :-(

src/libs/relay/conduit_relay_io_identify_protocol.cpp
JustinPrivitera: although it is still strange to have
```
if (yaml case)
   yaml
if (json case)
   json
```

couldn't it be
```
if (yaml case)
   yaml
else if (json case)
   json
```
instead?

src/tests/blueprint/t_blueprint_mesh_relay.cpp
cyrush: The error before would fail to even open the file, throwing an exception. 
I think we are ok without a diff.

JustinPrivitera: then I would remove the comment saying you are going to diff

src/libs/relay/conduit_relay_io_silo.cpp
JustinPrivitera: either github is being strange or it is still there

JustinPrivitera: I think github it being strange. So there should be no mpi version of the function.

src/libs/conduit/conduit_utils.cpp
cyrush: you only want to emit a character when you are not in a string, including the opening and closing quote.  This logic implements this. 

if you aren't in a string and you have encountered a quote, you are starting a string

If you are in a string and have encountered an ending quote, you are no longer in a string but you still don't want to emit the quote char.

src/tests/blueprint/t_blueprint_mesh_relay.cpp
cyrush: gotcha, cut and paste issue thanks for pointing that out

src/libs/conduit/conduit_utils.cpp
JustinPrivitera: I know, it's just hard to read. I would take these words here and put them in the code.

src/tests/blueprint/t_blueprint_mesh_relay.cpp
```diff
@@ -845,3 +845,45 @@ TEST(conduit_blueprint_mesh_relay, sparse_topos)
...
+    if(hdf5_enabled)
+    {
+        protocols.push_back("hdf5");
+    }
+
+    // TODO: finish wire up and we can add this case
+    // if(silo_enabled)
+    // {
+    //     protocols.push_back("silo");
+    // }
```
JustinPrivitera: TODO?

cyrush: not fully ready yet.

This adds logic to detect silo,but it doens't wire up the blueprint write or read yet.

cyrush:
` if(input[i] == quote_char[0] && ( i > 0 && ( input[i-1] = '\\' )))`
Second half `=` needs to be fixed.

src/tests/blueprint/t_blueprint_mesh_relay.cpp
cyrush: this is now wired up

src/libs/relay/conduit_relay_io_identify_protocol.cpp
cyrush: that will not work b/c json often includes `:` which is what we look for for yaml, so if we are strict matching to  yaml - we will never identify json.

JustinPrivitera: if it possible to switch the order? if json then json else if yaml then yaml

JustinPrivitera: if not then I will add a comment so this doesn't confuse anyone in the future

cyrush:
resolves #811

cyrush:
resolves https://github.com/LLNL/conduit/issues/1291

Issue #1365, opened 2025-02-21
Replace conduit handler function pointers with std::function


Arlie-Capps:
Propose to change `set_memcpy_handler()`, `set_memset_handler()`, and `AllocManager` and all associated interfaces to use `std::function`.  This would require no change to existing code, since existing code assigns or passes a function pointer and function pointers are assignable to std::function.  This change would add valuable functionality, since user code would then be able to set up lambdas or other types of callable as the copy, set, alloc, and free operations.

cyrush:
resolved in #1366

PR #1366, opened 2025-02-22
Use std::function for memory handler callbacks

Arlie-Capps:
Use std::function objects instead of function pointers for memory routines.

In C++, std::function is the equivalent of a function pointer.  It can store a function pointer, a lambda, or a functor.  This allows several use cases that a function pointer cannot handle.

This MR addresses #1365 .

### DIFF
--- a/src/libs/conduit/conduit_utils.cpp
+++ b/src/libs/conduit/conduit_utils.cpp
@@ -108,30 +108,22 @@ default_memcpy_handler(void * destination, const void * source, size_t num)
 
 //-----------------------------------------------------------------------------
 // Private namespace member that holds our memcpy callback.
-void (*conduit_handle_memcpy)(void * destination,
-                              const void * source,
-                              size_t num) = default_memcpy_handler;
+Handle_Memcpy conduit_handle_memcpy = default_memcpy_handler;
 
 //-----------------------------------------------------------------------------
 // Private namespace member that holds our memset callback.
-void (*conduit_handle_memset)(void * ptr,
-                              int value,
-                              size_t num ) = default_memset_handler;
+Handle_Memset conduit_handle_memset = default_memset_handler;
 
 //-----------------------------------------------------------------------------
 void
-set_memcpy_handler(void(*conduit_hnd_copy)(void*,
-                                           const void *,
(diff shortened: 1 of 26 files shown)


cyrush:
@Arlie-Capps looks like we have an issue with the static libs case that we will have to understand before merging.

src/libs/conduit/conduit_utils.cpp
```diff
@@ -277,6 +281,18 @@ void conduit_memset(void *ptr,
                     int value,
                     size_t num)
 {
+    struct Cb_init
+    {
+        Cb_init()
+        {
+            conduit_handle_memset = Handle_Memset(default_memset_handler);
+        }
+    };
```
cyrush: I think we are close, can we try moving this into the existing AllocManager, which already handles deferred setup for allocators?

src/libs/conduit/conduit_utils.hpp
```diff
@@ -273,10 +273,10 @@ namespace utils
     //-----------------------------------------------------------------------------
     /// Function types for memory handlers
     //-----------------------------------------------------------------------------
-    typedef std::function<void(void*, const void*, size_t)> Handle_Memcpy;
-    typedef std::function<void(void*, int, size_t)> Handle_Memset;
-    typedef std::function<void*(size_t, size_t)> Handle_Alloc;
-    typedef std::function<void(void*)> Handle_Free;
+    typedef std::function5<void(void*, const void*, size_t)> handle_memcpy_type;
```
cyrush:
```suggestion
    typedef std::function<void(void*, const void*, size_t)> handle_memcpy_type;
```
looks like a 5 sneaked in

Arlie-Capps: woops---thanks Cyrus

src/libs/conduit/conduit_utils.cpp
Arlie-Capps: done.

Arlie-Capps:
All right, @cyrush , if you have any more comments, I'd welcome them.

cyrush (approved):
Looks good, thanks for working through this change.

PR #1370, opened 2025-02-25
Let if() test the variable unevaluated

Arlie-Capps:
Cmake if() takes the name of the variable to test.

This PR changed `if(${tpl_all_dlls})` to `if(tpl_all_dlls)` .  See https://cmake.org/cmake/help/v3.29/command/if.html#variable.  (I cite documentation for CMake 3.29 because that is the version shipped with VS2022.)


cyrush (approved):
Thank you for this fix. I believe I hit this and forgot to solidify the fix.

PR #1371, opened 2025-02-25
Use rand() and srand()

Arlie-Capps:
The standard library includes rand() and srand().  It does not include random() and srandom(). 

This MR lets t_relay_io_hdf5.cpp build on Windows.  I read lots of commentary about how random/srandom are better than rand/srand, but rand and srand are sufficient for the needs of t_relay_io_hdf5.cpp and allow the test to build on Windows.


cyrush (approved):
Thank you Arlie!

PR #1393, opened 2025-03-18
Helper to generate the root filename

### LINKED ISSUE #1392: helper in relay that creates the root file name, given the mesh, path, and options.

JustinPrivitera:
Resolves #1392


cyrush:
Reminder: Changelog entry

JustinPrivitera:
> Reminder: Changelog entry

Thanks. Will do tomorrow.

src/libs/relay/conduit_relay_io_blueprint.hpp
```diff
@@ -143,6 +143,61 @@ void CONDUIT_RELAY_API write_mesh(const conduit::Node &mesh,
...
+                                                 int &par_rank,
+                                                 int &par_size);
+
+//-----------------------------------------------------------------------------
+// Generate root file name given parameters
+//-----------------------------------------------------------------------------
+
+//-----------------------------------------------------------------------------
+std::string CONDUIT_RELAY_API get_root_filename(const conduit::Node &mesh,
+                                                const std::string &path);
```
cyrush: can we change these to `generate_root_filename`?

src/libs/relay/conduit_relay_mpi_io_blueprint.hpp
```diff
@@ -157,6 +157,71 @@ void CONDUIT_RELAY_API write_mesh(const conduit::Node &mesh,
...
+                                                    std::string &opts_suffix,
+                                                    int &cycle,
+                                                    bool gen_name,
+                                                    const std::string &path,
+                                                    MPI_Comm comm);
+
+//-----------------------------------------------------------------------------
+// make sure some MPI task has data
+//-----------------------------------------------------------------------------
+bool CONDUIT_RELAY_API check_mesh_valid_for_save(int &cycle,
```
cyrush: check_for_non_empty_mesh()?

src/libs/relay/conduit_relay_io_blueprint.hpp
```diff
@@ -143,6 +143,61 @@ void CONDUIT_RELAY_API write_mesh(const conduit::Node &mesh,
...
 
+// -----------------------------------------------------------
+// get the number of local domains and the cycle info
+// -----------------------------------------------------------
+void CONDUIT_RELAY_API get_num_local_doms_and_cycle(int &local_num_domains,
+                                                    const Node &multi_dom,
+                                                    const bool &is_valid,
+                                                    std::string &opts_suffix,
+                                                    int &cycle,
+                                                    bool gen_name,
```
cyrush: We should break up this this function, it's doing more than its name suggests.
Also -- there are blueprint helpers to get the number of local domains -- can we use thoses?

In general, we want in params first and out params last in the list.

src/libs/relay/conduit_relay_io_blueprint.cpp
```diff
@@ -1825,6 +1751,403 @@ void write_mesh(const Node &mesh,
...
+    //-----------------------------------------------------------------------------
+    int cycle;
+    Node multi_dom;
+    bool is_valid;
+    int par_rank;
+    int par_size;
+#ifdef CONDUIT_RELAY_IO_MPI_ENABLED
+    Node n_local, n_reduced; // nodes used for MPI comm (share them for many operations)
+    if (!check_mesh_valid_for_save(cycle,
+                                   n_local,
```
cyrush: why do we have to pass n_local and n_reduced here?  Can't the reduce be hidden behind the interface?

src/libs/relay/conduit_relay_io_blueprint.cpp
```diff
@@ -1825,6 +1751,403 @@ void write_mesh(const Node &mesh,
...
+    opts_suffix = n_opts_suffix.as_string();
+
+#endif
+}
+
+//-----------------------------------------------------------------------------
+// make sure some MPI task has data
+//-----------------------------------------------------------------------------
+bool check_mesh_valid_for_save(int &cycle,
+#ifdef CONDUIT_RELAY_IO_MPI_ENABLED
```
cyrush: this is also cleaning the mesh, that's not obvious from the name.

src/libs/relay/conduit_relay_io_blueprint.cpp
```diff
@@ -1825,6 +1751,403 @@ void write_mesh(const Node &mesh,
...
+
+    // reduce to check to see if any valid data exists
+
+    n_local = (int)cycle;
+    relay::mpi::sum_all_reduce(n_local,
+                               n_reduced,
+                               mpi_comm);
+
+    global_boolean = n_reduced.as_int();
+
```
cyrush: what happens to is_valid, that is only used if it's valid locally vs the return code means global?
Do we need to pass is valid as an out param?

src/libs/relay/conduit_relay_io_blueprint.cpp
```diff
@@ -1825,6 +1751,403 @@ void write_mesh(const Node &mesh,
...
+            CONDUIT_ERROR("write_mesh invalid suffix option: \"" 
+                          << opts_suffix << "\"\n"
+                          " expected: \"default\", \"cycle\", or \"none\"");
+        }
+    }
+
+    //-----------------------------------------------------------------------------
+    // make sure some MPI task has data
+    //-----------------------------------------------------------------------------
+    int cycle;
```
cyrush: I think we only need mpi logic here if we are trying to resolve the cycle?

that's b/c we need to find some domain on some mpi task to find the cycle. 
If we aren't using the cycle, I think we can skip mpi logic?

cyrush (commented):
A few suggestions to refactor some of the helpers.

src/libs/relay/conduit_relay_io_blueprint.cpp
JustinPrivitera: Both are reused repeatedly. I didn't go through the motions of checking if their values were reused or not. If they are reused simply to avoid declaring new variables and not because they have values that are used later, then I can hide them.

JustinPrivitera: I can make new ones each time.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1985,6 +1985,94 @@ mesh::number_of_domains(const conduit::Node &mesh)
...
+        {
+            const Node &dom = doms_itr.next();
+            if (dom.has_child("state"))
+            {
+                state.set(dom["state"]);
+                break;
+            }
+        }
+    }
+}
```
JustinPrivitera, line 2013: is there a better way to put state on all tasks?

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1985,6 +1985,94 @@ mesh::number_of_domains(const conduit::Node &mesh)
...
+    index_t cycle = counters[path];
+    // if we're just using this function to ask what the cycle could be, 
+    // then we don't want to mess with the static counter. If we are
+    // in the process of writing files, then we can make up a new cycle.
+    if (! just_asking)
+    {
+        counters[path] ++;
+    }
+    return cycle;
+}
```
JustinPrivitera: It is confusing to have two methods for getting the cycle, but one of them has special logic in it to handle the case where we generate a cycle number for data with no cycle. This generation is dependent on the path we are using to save data.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1985,6 +1985,94 @@ mesh::number_of_domains(const conduit::Node &mesh)
...
+
+//-------------------------------------------------------------------------
+bool
+mesh::is_cycle_present(const conduit::Node &mesh)
+{
+    Node state;
+    ::conduit::blueprint::mesh::state(mesh, state);
+    return state.has_child("cycle");
+}
```
JustinPrivitera: I had to separate `is_cycle_present()` from cycle calculations because unifying them causes things to fall apart. Namely, if we generate a cycle for data with no cycle but we didn't end up needing to (like in the case where the requested suffix is none or default), then we have incremented our counter when we didn't have to, breaking several tests, just because we wanted to know if the cycle was present or not. Separating it into two functions avoids this issue.

src/libs/blueprint/conduit_blueprint_mesh.cpp
cyrush: I think we should handle the non cycle case outside of the cycle function.
It can happen in the gen filenames function.

JustinPrivitera: We can do it outside, but it can't just happen in the generate filenames function. The blueprint and silo writers don't use the generate filenames function.

JustinPrivitera: The issue with doing it outside is that now we need to bring back MPI communication to make sure all tasks agree on the cycle.

JustinPrivitera: Can we assume that `opts_suffix` will be the same across all tasks? Before we go about changing it depending what the cycle is. At the beginning of the function, after we've parsed the options, can we assume `opts_suffix` will be the same across all tasks?

src/libs/blueprint/conduit_blueprint_mpi_mesh.cpp
```diff
@@ -237,6 +237,69 @@ number_of_domains(const conduit::Node &n,
...
+state(const conduit::Node &mesh,
+      conduit::Node &state,
+      MPI_Comm comm)
+{
+    Node local_state;
+    ::conduit::blueprint::mesh::state(mesh, local_state);
+
+
+    Node global_state_nodes;
+    relay::mpi::all_gather_using_schema(local_state,
```
cyrush: we could find the first instance of state and broadcast that to all, vs an all gather, which will send copies of all state to all mpi tasks

JustinPrivitera: done

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: `is_cycle_present()` is gone now

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: yes there is and we did it.

JustinPrivitera (commented):
I think ready for you to take a final look @cyrush

cyrush (approved):
looks good, thanks!

PR #1419, opened 2025-04-02
Example using Conduit allocators

Arlie-Capps:
This is an example showing how to use an adapter class to use Umpire allocators with Conduit.


cyrush (approved):
Looks great, thanks for this example!

We can wire this up in CI when we have umpire builds in the future.

cyrush:
@Arlie-Capps  I am merging this in, so we it will be part of the release.

PR #1424, opened 2025-04-15
Silo/Overlink I/O Nameschemes

JustinPrivitera:
TODO
- [x] have all new tests trigger the empty list
- ~~[ ] read silo test with a real live silo data file with nameschemes~~ - will do in the future
- [x] check if it is possible to scramble the partition map with MPI and then test that case
- [x] changelog entries for nameschemes, block types, and overlink n domains to m files case bug fix


src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -411,6 +411,261 @@ namespace silo
...
 
+//-----------------------------------------------------------------------------
+// -- begin conduit::relay::<mpi>::io::silo::detail::name_generator_tools --
+//-----------------------------------------------------------------------------
+namespace name_generator_tools // tools for making `SiloTreePathGenerator`s
+{
+
+//-----------------------------------------------------------------------------
+
+class SiloTreePathGenerator
```
cyrush, line 422: why do we need to use std::unique ptrs and uses rvalues?
If get paths, etc are members of the  path generator -- we can just setup instance members and avoid moves?

JustinPrivitera: `std::unique_ptr` was used because I needed to use pointers to instances of `SiloTreePathGenerator`s. I needed pointers because I ran into the strangest bug when setting up my maps of name generators:
```c++
std::map<std::string, std::unique_ptr<SiloNameGenerator>> mat_path_gen
```
Originally I just had maps from object names (mat names, var names, specset names) to `SiloNameGenerator`s which are defined as follows:
```c++
using SiloNameGenerator = detail::name_generator_tools::SiloTreePathGenerator;
```
But I kept getting strange segfaults when I called `SiloNameGenerator`'s methods - this was fixed by switching to maps from object names to pointers. I believe the issue was that I did not have a copy constructor defined for `SiloTreePathGenerator`s and maps do strange things with the objects they control involving their copy constructors. Once I got things working I didn't spend more time to understand why my original design didn't work. Unique pointers are better than regular pointers since I don't have to worry about correctly freeing everything at the end; they handle that for me.

To your other point, `get_paths` and friends are _not_ members of `SiloTreePathGenerator`. In my opinion, Silo provides minimal support for nameschemes, and ought to provide you with something like a `SiloTreePathGenerator`, since every code that wants to read Silo nameschemes has to implement the same thing. Maybe it makes sense to have a `SiloTreePathGeneratorGenerator` that makes `SiloTreePathGenerator`s, and then all those functions could be member methods of that class.

JustinPrivitera:
@cyrush can we go ahead and merge then?

cyrush:
I still want to discuss why we have to use the unique pointers. Something is off if you were seeing crashes with and I want to understand.

JustinPrivitera:
Ok, let's look side-by-side. Debugging it the first time took several hours and I never could find out what was going wrong.

JustinPrivitera:
I don't think we can use zero copy to fill the silo root index. How do we zero copy character pointers?

JustinPrivitera:
I have investigated the issue with `unique_ptr`s thoroughly. Without `unique_ptr`s being used to house `SiloTreePathGenerator`s, we hit a segfault because the default copy constructor for `SiloTreePathGenerator`s is invoked in my method `create_silo_tree_path_generator()`. The default copy constructor attempts to shallow copy the members, which are as follows:
```c++
    int                      nblocks;
    std::vector<std::string> names_list;
    DBnamescheme            *file_namescheme;
    DBnamescheme            *block_namescheme;
    int                      empty_count;
    const index_t           *empty_list;
```
My belief is that the `DBnamescheme*` objects, in being shallow copied, are copied inappropriately, yielding a segfault when `DBGetName` is called using one of them.

How to address this? I attempted to be more judicious with `unique_ptr` use and use them within the `SiloTreePathGenerator` objects to house the namescheme objects, but we run into the same issue when it comes time to invoke the default copy constructor.

I could write my own copy constructor, but I can hardly see how it could be any better than the default, as I don't know of a way to appropriately copy a `DBnamescheme` object.

Why is the copy constructor being invoked in the first place? In my method `create_silo_tree_path_generator()`, I expected that the way I wrote it would cause the compiler to use RVO to avoid any copying. The compiler is not doing this, probably because we are on C++14 and are not using a new enough compiler. However, we shouldn't write code that only works if we have a new enough compiler.

After looking at this a lot, I think it makes the most sense to use pointers to `SiloTreePathGenerator`s instead of using the objects directly (as I was doing originally). The modern way to use pointers is to use one of the four special pointers; hence unique pointers. This will avoid all copying of the objects themselves, and is a tidy way to ensure things are cleaned up at end of life. We only copy pointers, not the objects themselves. No matter what version of the compiler you have it works.

JustinPrivitera:
With my two comments above I believe I have addressed all of your feedback @cyrush. I think this is ready to be merged. Please let me know if there is anything else we should discuss.

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -411,6 +411,269 @@ namespace silo
...
+        root_file,
+        
+        // objpath - we are at the root of the root file
+        ".",
+        
+        // number of blocks
+        num_domains,
+        
+        // mesh paths
+        std::move(get_paths(n_item, what_kind_of_paths, num_domains)),
```
cyrush, line 604: It would be better to have the class init this vs using a move in the constructor

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -411,6 +411,269 @@ namespace silo
...
+                      std::map<std::string, SiloTreePathGenerator*> &path_gen_map)
+{
+    if (mesh_index.has_child(item))
+    {
+        auto item_itr = mesh_index[item].children();
+        while (item_itr.has_next())
+        {
+            const Node &n_item = item_itr.next();
+            const std::string item_name = item_itr.name();
+            path_gen_map.emplace(item_name,
```
cyrush, line 645: push_back vs emplace_back is simple:

```
Do you already have a T
yes? use push_back

else use emplace_back 
```

JustinPrivitera: I think insert would create a temporary pair. Emplace is more efficient, however marginally. It is more idiomatic for this case. This design was more relevant when the path gen map was dealing with `SiloTreePathGenerator`s directly and not pointers to them. Leaving the design this way is still probably best, but there aren't as many good reasons.

src/libs/relay/conduit_relay_io_silo.cpp
JustinPrivitera: I will fix this - the move is not necessary. Not sure what I was thinking a few months ago when I did this.

PR #1430, opened 2025-04-24
Improve compare_pointwise() and conduit_adjset_validate program.

BradWhitlock:
I've been looking at a mesh with adjsets that appear valid but turn out to have slightly different point ordering on either side of the domain interface. The ``adjset::validate()`` function says they are good _(for what it checks)_ and then a more stringent test is done using ``compare_pointwise()``. When a dataset failed this test, node differences were output. This output was not useful in constructing a point mesh of where the adjsets differed, making it hard to identify differences.

This PR:
* Makes ``compare_pointwise()`` output info that contains the pointwise differences needed to make a point mesh
* Enhances conduit_adjset_validate to make point meshes from ``compare_pointwise()`` differences.


cyrush (approved):
Looks good, I am working on resolving the ci hiccup (which isn't related to your changes at all) on develop.

PR #1433, opened 2025-04-28
ci heal

cyrush:


cyrush:
one of our github actions cases was still using a ubuntu 20.04 runner, which was removed b/c it is out of support.

PR #1438, opened 2025-06-02
parallel hdf5 and cgns support in build script

mlohry:
@cyrush  this is a pre-cursor PR for parallel CGNS support. I set the build to enable parallel support for HDF5 and CGNS when `enable_mpi` is on. If you want a separate option to opt-in to pcgns and phdf5 support that's fine by me as well.

@brtnfld how are these for default build settings?


cyrush (approved):
@mlohry looks good, I think it's fine to have HDF5 MPI on whenever MPI is enabled.

cyrush:
@mlohry these jobs failed b/c the windows image in ci was removed around the time of testing. If we merge develop into your branch, we can fully test

PR #1439, opened 2025-06-03
Change adjset spatial sort

BradWhitlock:
This PR takes a different approach to sorting points for adjsets that are made in ``generate_corners()``. The new algorithm computes extents and min edge length for the mesh and uses this information to make an implicit mesh that can bin the points. This enables the code to compute a unique id for each point, which is used in sorting entities.

* Completely replaces ffloat64 approach
* Adds MeshInfo struct and functions to compute it on input topology
* Adds Quantizer struct to make indices from point
* Adds MeshInfoCollection (and parallel version)
* Plumbs generate_xxx functions so they can use MeshInfoCollection and Quantizer
* Add some using statements for type readability
* Use accessors to populate the adjset data
* Added test that makes adjsets for a 4 domain mesh that gets rotated around to test adjset spatial sorting.
* Clean up some warnings


BradWhitlock:
This is working with Conduit (and the host code). I plan to do some more testing to see whether other test suites are affected.

cyrush:
@BradWhitlock  windows issue is due to a problem with the github windows runner environment. 

See: https://github.com/actions/runner-images/issues/12435

I am working on a fix.

rhornung67:
@cyrush have you tried a slightly earlier version of windows than "latest" that is still supported, like 2022?

cyrush:
Latest is 2022. Was forced to move on from 2019. Also tried 2025, bc that was suggested. I think for now we have to install newer clang needed by the visual studio version in the base image.

-Cyrus

cyrush:
windows issue should be resolved in: https://github.com/LLNL/conduit/pull/1448

Merge develop in and that should heal the ci.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -3640,14 +3746,30 @@ mesh::generate_centroids(conduit::Node& mesh,
...
-
     generate_decomposed_entities(
         mesh, src_adjset_name, dst_adjset_name, dst_topo_name, dst_cset_name, s2dmap, d2smap,
         conduit::blueprint::mesh::topology::unstructured::generate_centroids, identify_centroid, centroid_dims,
-        query);
+        query, mic);
 }
 
+//-----------------------------------------------------------------------------
+//[[deprecated]]
```
Arlie-Capps: Why is this commented out?  (and similar below)

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -7038,13 +7182,11 @@ mesh::adjset::is_maxshare(const Node &adjset)
...
         const Node &group = group_itr.next();
-        const Node &group_values = group["values"];
+        const auto values = group["values"].as_index_t_accessor();
 
-        for(index_t ni = 0; ni < group_values.dtype().number_of_elements(); ni++)
+        for(index_t vi = 0; vi < values.number_of_elements(); vi++)
         {
-            Node temp(DataType(group_values.dtype().id(), 1),
-                (void*)group_values.element_ptr(ni), true);
-            const index_t next_id = temp.to_index_t();
```
Arlie-Capps, line 7047: KILL THE PATTERN WITH FIRE!!

Thanks.

src/tests/blueprint/t_blueprint_mpi_mesh_utils.cpp
```diff
@@ -156,21 +184,21 @@ TEST(conduit_blueprint_mpi_mesh_utils, adjset_validate_element_2d)
...
     // Now, adjust the adjset for domain1 so it includes an element not present in domain 0
     if(root.has_child("domain1"))
         root["domain1/adjsets/main_adjset/groups/domain0_1/values"].set(std::vector<int>{0,2,4});
     info.reset();
     save_mesh(root, "adjset_validate_element_2d_bad");
-    res = validate(root, "main_adjset", info);
+    res = validate(root, "main_adjset", opts, info);
     EXPECT_FALSE(res);
-    //info.print();
+    //printNode(info);
```
Arlie-Capps: If we don't need this, please remove.  And similar lines further on in the file.

src/tests/blueprint/t_blueprint_mpi_mesh_utils.cpp
```diff
@@ -278,49 +307,358 @@ TEST(conduit_blueprint_mpi_mesh_utils, adjset_compare_pointwise_2d)
...
+        B.m_angle = angle;
+        B.m_resolution = res;
+        B.m_permute = permute > 0; // Permute node/zone order
+        B.build(n_mesh);
+#if 0
+        in_rank_order(comm, [&](int /*r*/)
+        {
+            printNode(n_mesh);
+        });
+#endif
```
Arlie-Capps: Please remove unless necessary. Also remove other `#if 0` sections unless they are necessary.

Arlie-Capps (approved):
Thanks for this work! Just found a few minor things.

src/libs/blueprint/conduit_blueprint_mesh.cpp
BradWhitlock: I copy/pasted a previous version of the function where it was commented-out. I had added `[[deprecated]]` in a previous version of the method but the attribute might not have been supported at the time across compilers so I commented it out.

I thought about it more and [[deprecated]] for those functions is not the right idea since they are still useful in serial (without wanting to deal with the extra arguments). I'm going to remove the [[deprecated]] comments on these.

src/tests/blueprint/t_blueprint_mpi_mesh_utils.cpp
BradWhitlock: Fixed it.

src/tests/blueprint/t_blueprint_mpi_mesh_utils.cpp
BradWhitlock: Fixed it.

src/libs/blueprint/conduit_blueprint_mesh_utils.hpp
```diff
@@ -984,14 +1165,16 @@ namespace adjset
...
      @param[out] info A node that contains any errors.
 
      @return True if the adjsets in all domains contained no errors; False if
              there were errors.
      */
-    bool CONDUIT_BLUEPRINT_API validate(const Node &doms,
+    bool CONDUIT_BLUEPRINT_API validate(const conduit::Node &doms,
                                         const std::string &adjsetName,
-                                        Node &info);
+                                        const conduit::Node &options,
```
JustinPrivitera: Should the options node be an optional argument?

src/tests/blueprint/t_blueprint_mesh_utils.cpp
```diff
@@ -164,18 +164,18 @@ TEST(conduit_blueprint_mesh_utils, shapetype)
 //-----------------------------------------------------------------------------
 TEST(conduit_blueprint_mesh_utils, adjset_validate_element_0d)
 {
-    conduit::Node root, info;
+    conduit::Node root, opts, info;
     create_2_domain_0d_mesh(root, 0, 1);
     save_mesh(root, "adjset_validate_element_0d");
-    bool res = conduit::blueprint::mesh::utils::adjset::validate(root, "main_adjset", info);
+    bool res = conduit::blueprint::mesh::utils::adjset::validate(root, "main_adjset", opts, info);
```
JustinPrivitera: I asked my earlier question about if we needed to have opts be a mandatory part of the API because it seems like for all of these tests we now have to add opts in when they are unused.

src/tests/blueprint/t_blueprint_mpi_mesh_tiled.cpp
```diff
@@ -118,17 +127,17 @@ make_tiled(conduit::Node &mesh, const int dims[3], const int domains[3],
...
-                std::string domainName;
-                if(!domainNumbering.empty())
-                {
-                    domainName = conduit_fmt::format("domain_{:07}",domainNumbering[domainIndex]);
-                }
-                else
-                {
-                    domainName = conduit_fmt::format("domain_{:07}",domainIndex);
-                }
+                std::string domainName = conduit_fmt::format("domain_{:07}",domainId);
```
JustinPrivitera, line 140: I don't understand why we can delete the special logic for indirect domain ids.

src/tests/blueprint/t_blueprint_mpi_mesh_tiled.cpp
```diff
@@ -388,8 +411,8 @@ top: [3,2]
     EXPECT_EQ(selectedDomains[rank].size(), conduit::blueprint::mesh::domains(mesh).size());
 
     // Make sure that the adjset is compares pointwise.
-    conduit::Node info;
-    bool same = conduit::blueprint::mpi::mesh::utils::adjset::compare_pointwise(mesh, "main_adjset", info, MPI_COMM_WORLD);
+    conduit::Node popts, info;
+    bool same = conduit::blueprint::mpi::mesh::utils::adjset::compare_pointwise(mesh, "main_adjset", popts, info, MPI_COMM_WORLD);
```
JustinPrivitera: is `popts` intentional or a typo?

src/tests/blueprint/t_blueprint_mpi_mesh_transform.cpp
```diff
@@ -927,9 +927,9 @@ TEST(conduit_blueprint_mesh_examples, generate_corners_wonky)
     }
     test_adjset_points(mesh, "corner_pairwise_adjset");
 
-    conduit::Node info;
+    conduit::Node info, opts;
     bool same_pointwise = conduit::blueprint::mpi::mesh::utils::adjset::compare_pointwise(
-                              mesh, "corner_pairwise_adjset", info, MPI_COMM_WORLD);
+                              mesh, "corner_pairwise_adjset", opts, info, MPI_COMM_WORLD);
```
JustinPrivitera: do opts need to be mandatory for `compare_pointwise`?

src/tests/blueprint/t_blueprint_mpi_mesh_tiled.cpp
BradWhitlock: `domainId` is set from `domainNumbering` about 30 lines earlier now.

src/libs/blueprint/conduit_blueprint_mesh_utils.hpp
BradWhitlock: I changed the `options` argument so it comes last and is an optional argument, defaulted to `conduit::Node()`. I updated places in Conduit where the function is called. I did this for adjset `validate` and `compare_pointwise` functions.

src/tests/blueprint/t_blueprint_mesh_utils.cpp
BradWhitlock: I removed opts from all places where a default value will suffice.

src/tests/blueprint/t_blueprint_mpi_mesh_tiled.cpp
BradWhitlock: Removed it so it uses the default options.

src/tests/blueprint/t_blueprint_mpi_mesh_transform.cpp
BradWhitlock: Again, removed it in favor of default options.

PR #1447, opened 2025-06-24
Silo/Overlink IO Nameschemes Minor Tweaks

JustinPrivitera:
Follow up to #1424


JustinPrivitera:
@cyrush I didn't change the use of insert versus emplace since I stand by what I did before.

At first I removed std::move here but then I went back and iterated and decided this was better.

I don't think you will approve my PR. 😞

cyrush (approved):

cyrush:
:-)

Issue #1445, opened 2025-06-20
set_X_vector fails for 0 length vectors


mlohry:
Setting a conduit node to a 0-size vector causes a segfault:

```
const std::vector<int> empty_vec = {};
conduit::Node node;
node.set_int32_vector(empty_vec);
```

segfaults as the function tries to get the data pointer by calling `&empty_vec[0]` instead of `empty_vec.data()`. I believe the expected behavior is that zero-length arrays are valid.

aside: do you have a .clang-format for conduit?

cyrush:
@mlohry - thanks for the report, we will fix this.

cyrush:
This was resolved in #1450 (it was done along side build system changes)

PR #1450, opened 2025-07-07
macos 15.5 updates for spack builds

cyrush:


cyrush:
This also resolves #1445

PR #1452, opened 2025-07-24
AMR bent grid example

Arlie-Capps:
Enhances the "bent_grid" example to allow nested domains.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 
 ### Added
 
+#### Blueprint
+- Added `amr_bent_multi_grid` example mesh.  This enhances the `bent_multi_grid` example to allow a user to specify, in one of the domains, a nested domain with its ID and refinement ratio.  This will allow example AMR meshes.
+
 #### Relay
 - Added namescheme read and write support to the Relay Silo/Overlink Reader/Writer.
 - Added support for block types or unified types to the Relay Silo/Overlink Reader/Writer.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 ### Added
 
 #### Blueprint
-- Added `amr_bent_multi_grid` example mesh.  This enhances the `bent_multi_grid` example to allow a user to specify, in one of the domains, a nested domain with its ID and refinement ratio.  This will allow example AMR meshes.
+- Added `bent_multi_grid_amr` example mesh.  This enhances the `bent_multi_grid` example to allow a user to specify, in one of the domains, a nested domain with its ID and refinement ratio.  This will allow example AMR meshes.
 
 #### Relay
 - Added namescheme read and write support to the Relay Silo/Overlink Reader/Writer.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 
 ### Added
 
+#### Blueprint
+- Added `amr_bent_multi_grid` example mesh.  This enhances the `bent_multi_grid` example to allow a user to specify, in one of the domains, a nested domain with its ID and refinement ratio.  This will allow example AMR meshes.
```
cyrush: lets change this to `bent_multi_grid_amr`

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2679,6 +2679,155 @@ braid_bent_hexs(const Node & spec, Node &res)
...
+    //   nest_ratio: 3
+    // res gets filled with a new domain refined from parent, including a nestset
+    // parent node gets a nestset tying it to res
+
+    // Basically, call braid_bent_quads, then do the nestset.
+    // spec must contain details about nesting, like this:
+    //   nest_child_id: 13
+    //   nest_ratio: 3
+    // res gets filled with a new domain refined from parent, including a nestset
+    // parent node gets a nestset tying it to res
```
cyrush, line 2713: spec details are duplicated here

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -2679,6 +2679,155 @@ braid_bent_hexs(const Node & spec, Node &res)
...
+                Node& window = nset["windows"][windowname];
+                window["domain_id"] = domain_id;
+                window["domain_type"] = domain_type;
+                window["origin/i"] = 0;
+                window["origin/j"] = 0;
+                window["dims/i"] = dims_i;
+                window["dims/j"] = dims_j;
+                window["ratio/i"] = nest_ratio;
+                window["ratio/j"] = nest_ratio;
+        };
```
cyrush: I find it confusing that some are passed and some are captured, should we just create a helper function instead of a lambda?

cyrush (changes requested):
Hi Arlie, looks good, I have a few suggestions.

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
Arlie-Capps: The operations are the same.  The values that are different get passed in.  The (one) value that is the same gets captured.  Sure, I'll  move this to a helper function instead of a lambda, no problem.

cyrush: Yes, I just think its easier to parse as a human w/o the lambda in this case.

CHANGELOG.md
Arlie-Capps: done

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
Arlie-Capps: removed

src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
Arlie-Capps: done.  @cyrush , when you have a moment, please take another look.

cyrush (approved):
Looks good, thanks!

PR #1460, opened 2025-09-03
Add adjacency set between adjacent refined domains.

Arlie-Capps:
Complete the AMR example: add an adjacency set between each neighboring domain at the same level of refinement.


src/libs/blueprint/conduit_blueprint_mesh_examples.cpp
```diff
@@ -3964,8 +3967,12 @@ void CONDUIT_BLUEPRINT_API bent_multi_grid(const conduit::Node &spec,
                 {
                     nest_hexs(dom_node, res[doms_it.name()], res[nest_name]);
                 }
+                nested_doms[nest_name].set_external(res[nest_name]);
             }
         }
+
+        // Tie adjacent refined domains together with adjsets.
+        braid_init_example_adjset(nested_doms);
```
Arlie-Capps, line 3974: At this point, I intended to have `nested_doms` contain (set external) each refined domain: the node `res` contains the full output, the original domains, the refined domains, and of course the nestsets.  Then, I expect that calling `braid_init_example_adjset(nested_doms);` would add the appropriate adjsets between adjacent refined domains.  But when `res` finally gets saved out to disk, there are no adjsets between nested domains.  Stepping through  `braid_init_example_adjset(nested_doms)` in the debugger shows the adjsets getting created, as I expect, but after stepping out, the Nodes that I can view through `res` do not have the structures I just added.

@cyrush , what am I missing here?  Can't I process an object Node whose values are set_external Nodes?

Thanks.

cyrush: @Arlie-Capps  
`nested_doms[nest_name].set_external(res[nest_name])` will be setup with its own schema that fully describes data array leaves in `res[nest_name]`.

But any changes to `nested_doms` will have no impact on res.

Arlie-Capps: I would like changes to `nested_doms[nest_name]` to show up in `res`.  Is there a way to do this, other than following line 3975 with assignments back into `res`?

cyrush: Why not pass res? Why is nested doms needed?

Arlie-Capps: Changed this so that we store nested domains in their own Node, then compute adjsets, then store the nested doms back into the result.

cyrush (approved):
Looks good, can you add a note in the changelog?

PR #1467, opened 2025-09-17
Add missing relay io silo function descriptions

### LINKED ISSUE #1465: Document silo read/write options.
I was going to try `relay::io::silo::load_mesh()` to read some Silo files and I wanted to see what options it might take. I'd like to control which MPI ranks read which domains. The options are not documented in the [function's doxygen comments](https://github.com/LLNL/conduit/blob/8543892a582737eb5941b1e90df068a75543df92/src/libs/relay/conduit_relay_io_silo.hpp#L113).

* Document the options so we know what to provide.

JustinPrivitera:
Resolves #1465


cyrush (approved):
thanks for adding these details

PR #1474, opened 2025-11-05
Updates to 2D polygonal transformation

nselliott:
This brings in updated features to the `to_polygonal` transformation of 2D structured AMR meshes:

- Handles block rotations
- Fixes case where a corner of a fine domain is not a vertex on the coarse mesh
- Includes fields in the transformation, including communication of vertex field values from fine domains to coarse domains at hanging vertices, matching the communication of coordinate positions at those same vertices.


cyrush (approved):
@nselliott  this looks good. 

More comments in conduit_blueprint_mpi_mesh about the cases would help our future selves. 

Do you think we should we wire up the more complex bent grid example as part of this PR, or in a future PR (future PR is fine by me)

nselliott:
> @nselliott this looks good.
> 
> More comments in conduit_blueprint_mpi_mesh about the cases would help our future selves.
> 
> Do you think we should we wire up the more complex bent grid example as part of this PR, or in a future PR (future PR is fine by me)

@cyrush I think we can do the bent grid in the next PR.

PR #1478, opened 2025-11-20
Added CONDUIT_VERSION_VALUE and CONDUIT_MAKE_VERSION_VALUE macros.

### LINKED ISSUE #1469: Conduit version macro niceties.
Can something like this _(or better)_ be added to conduit's _conduit_config.h_ to make it easier to do conditionally-compiled logic based on the Conduit version?

```
#define CONDUIT_MAKE_VERSION(MAJOR, MINOR, PATCH) (((MAJOR)*10000) + ((MINOR)*100) + (PATCH))
#define CONDUIT_VERSION_VALUE CONDUIT_MAKE_VERSION(CONDUIT_VERSION_MAJOR, CONDUIT_VERSION_MINOR, CONDUIT_VERSION_PATCH)
```

Then we can do things like:
```
#if  CONDUIT_VERSION_VALUE >= CONDUIT_MAKE_VERSION(0, 9, 5)
  // Use something only available in 0.9.5 or later
#endif 
```

BradWhitlock:
This PR resolves #1469.

* Added CONDUIT_VERSION_VALUE macro that encodes current Conduit version as an integer.
* Added CONDUIT_MAKE_VERSION_VALUE that makes a version number from major,minor,patch version numbers.
* Added a test for the macros.

```
// User code
#if CONDUIT_VERSION_VALUE >= CONDUIT_MAKE_VERSION_VALUE(0, 9, 5)
 // Use that new feature that appeared in Conduit 0.9.5
#endif
```


cyrush (approved):
Thanks for adding these!

PR #1481, opened 2025-11-21
Speed up mesh::topology::unstructured::to_polygonal()

### LINKED ISSUE #1476: Braid example for polyhedra is slow.
The braid example is slow when making hex polyhedra. Profilers show this function is taking the time: `conduit::blueprint::mesh::topology::unstructured::to_polygonal(conduit::Node const&, conduit::Node&)`.

```
conduit::blueprint::mesh::examples::braid("hexs_poly", dims[0], dims[1], dims[2], n_mesh);
```
Image

the plot was generated using a LivChat-generated Python script that invoked a program that called `braid` and timed it.
```
import subprocess
import re
import numpy as np
import matplotlib.pyplot as plt

# Configurable number of samples
num_samples = 10

# Range of number of zones
min_zones = 1
max_zones = 27000 #1_000_000

# Generate num_samples values between min_zones and max_zones (log spaced for better coverage)
zone_counts = np.logspace(np.log10(min_zones), np.log10(max_zones), num=num_samples, dtype=int)

generation_times = []
actual_zone_counts = []

for nzones in zone_counts:
    # Calculate the cube root and round to nearest integer
    d = int(round(nzones ** (1/3)))
    dims = [d, d, d]
    # Calculate the actual number of zones used
    actual_nzones = dims[0] * dims[1] * dims[2]
    actual_zone_counts.append(actual_nzones)
    cmd = ['./conduit_polyhedral', '--dims'] + [str(x) for x in dims]
    try:
        result = subprocess.run(cmd, capture_output=True, text=True, check=True)
        output = result.stdout
        # Extract "generation: xxx" using regex
        match = re.search(r'generation:\s*([0-9.]+)', output)
        if match:
            gen_time = float(match.group(1))
        else:
            gen_time = None
            print(f"Warning: Could not find generation time in output for dims {dims}")
    except subprocess.CalledProcessError as e:
        print(f"Error running command {cmd}: {e}")
        gen_time = None
    generation_times.append(gen_time)
    print(generation_times)

# Filter out any None values (failed runs)
filtered_zone_counts = []
filtered_generation_times = []
for z, t in zip(actual_zone_counts, generation_times):
    if t is not None:
        filtered_zone_counts.append(z)
        filtered_generation_times.append(t)

# Plotting
plt.figure(figsize=(10, 6))
plt.plot(filtered_zone_counts, filtered_generation_times, marker='o')
plt.xscale('log')
plt.xlabel('Number of Zones')
plt.ylabel('Generation Time')
plt.title('Generation Time vs Number of Zones')
plt.grid(True)
plt.show()
```

BradWhitlock:
This PR resolves #1476.

* `mesh::topology::unstructured::to_polygonal()` was rewritten to be faster. ~600x faster
* The new method will preserve quad faces in wedge and pyramid elements instead of splitting into triangles
* Support was added for "mixed" elements too
* `conduit::blueprint::mesh::utils::ShapeType` was enhanced with better "mixed" element support
* `ShapeType` was enhanced so it should be faster to construct. a) removed a string member  b) changed how specific id values are initialized so they are not initialized each time through the ShapeType constructor.
* New tests were added for mixed element mesh conversion

image


cyrush (approved):
Awesome! Thank you!

PR #1486, opened 2025-12-18
Make sure conduit_blueprint_mesh_utils_iterate_elements.hpp gets installed

BradWhitlock:
I wanted to use `conduit_blueprint_mesh_utils_iterate_elements.hpp` in some code outside of Conduit and found that was not in the installed `include` directory.

* Make sure the file gets installed
* I moved converted an inline function into a lambda in one of the template functions.


cyrush (approved):
Thanks!

PR #1489, opened 2025-12-30
Wire up Node::load and Node::save conduit_bin yaml case

### LINKED ISSUE #1313: YAML schema TODOs
- [x] Make
https://github.com/LLNL/conduit/blob/92e22b350fe6320e6f298b9f7429b6bf1085b542/src/libs/conduit/conduit_node.cpp#L13338
look like
https://github.com/LLNL/conduit/blob/92e22b350fe6320e6f298b9f7429b6bf1085b542/src/libs/conduit/conduit_node.cpp#L13178
(Add `conduit_yaml`, `conduit_yaml_external`, `conduit_base64_yaml`) - #1335

- [x] Get the final test in `t_conduit_generator.cpp` to work. - #1335
- [x] do the `conduit_base64_yaml` case - #1335
- [ ] for `conduit_bin` based i/o, those paths look for a related `.conduit_json` file to get schema info. 
Now that we have new fancy yaml schemas, we should also look for related `.conduit_yaml` file and also wire up their creation. Node::load and Node::save `conduit_bin` case.
- [x] schema should be able to save to yaml as well
- [x] schema yaml save tests

JustinPrivitera:
Resolves #1313

- Additionally fixes a longstanding bug preventing explicit length 0 in yaml schema
- Additionally fixes a longstanding bug where empty objects or lists were not written correctly to yaml schema.

- [x] TODO need to understand how to choose to save conduit bin yaml. Will this be a new protocol? Something we prefer? What do we use to make this choice?
- [x] TODO need to add tests in `t_conduit_node_save_load.cpp`


src/libs/conduit/conduit_generator.cpp
```diff
@@ -2574,7 +2573,7 @@ Generator::Parser::YAML::parse_leaf_dtype(yaml_document_t *yaml_doc,
                 length = get_yaml_sequence_length(value_node);
             }
             // support explicit length 0 in a schema
-            else if (fetch_yaml_node_from_object_by_name(yaml_doc, yaml_node, "number_of_elements"))
+            else if (! fetch_yaml_node_from_object_by_name(yaml_doc, yaml_node, "number_of_elements"))
```
cyrush, line 2576: Thanks for finding this!

src/libs/conduit/conduit_node.cpp
```diff
@@ -339,22 +339,45 @@ Node::load(const std::string &ibase,
 
     if(proto == "conduit_bin")
     {
+        bool loaded_schema = false;
+        std::string error_message;
+
         Schema s;
-        std::string ifschema = ibase + "_json";
+        std::string ifschema_json = ibase + "_json";
+        std::string ifschema_yaml = ibase + "_yaml";
```
cyrush: Should we try to look for the file before parsing?

We may not have had the "file exists" helper when the json schema read was first implemented.

src/libs/conduit/conduit_node.cpp
```diff
@@ -416,7 +448,42 @@ Node::mmap(const std::string &stream_path)
     std::string ifschema = stream_path + "_json";
 
     Schema s;
-    s.load(ifschema);
+    bool loaded_schema = false;
+    std::string error_message;
+
+    std::string ifschema_json = stream_path + "_json";
+    std::string ifschema_yaml = stream_path + "_yaml";
```
cyrush: same q about checking if the file exists before trying to parse, to avoid exceptions in the case where we have yaml

src/libs/conduit/conduit_schema.cpp
```diff
@@ -627,32 +627,46 @@ Schema::to_yaml_stream(std::ostream &os,
 {
     if(m_dtype.id() == DataType::OBJECT_ID)
     {
-        os << eoe;
         size_t nchildren = children().size();
-        for(size_t i=0; i <  nchildren;i++)
+        if (nchildren == 0)
+        {
```
cyrush, line 632: thanks for fixing this

src/libs/conduit/conduit_schema.cpp
```diff
@@ -627,32 +627,46 @@ Schema::to_yaml_stream(std::ostream &os,
...
         }
     }
     else if(m_dtype.id() == DataType::LIST_ID)
     {
-        os << eoe;
         size_t nchildren = children().size();
-        for(size_t i=0; i < nchildren;i++)
+        if (nchildren == 0)
+        {
+            os << "[]" << eoe;
```
cyrush, line 655: and thanks for fixing this!

cyrush (commented):
Thanks for adding this! Please add a note to the changelog.

Question about detection logic:

Should we check if the json or yaml files exists before parsing to avoid an exception to detect the yaml case?

src/libs/conduit/conduit_node.cpp
JustinPrivitera: fixed

src/libs/conduit/conduit_node.cpp
JustinPrivitera: fixed

JustinPrivitera:
@cyrush I made the requested changes. I also added conduit_bin_yaml and conduit_bin_json as protocols for node::load. The tests are currently configured to load with conduit_bin protocol, which looks for the existence of json and yaml files and will prefer json. That means if you run the tests multiple times in a row, you may get false passes, since we do not clear results as we go. I am wondering if I should
1. leave as is, and remember to clear results in the future when I run these tests
2. clear results as I go in the tests
3. have all the yaml tests use conduit_bin_yaml to load to explicitly exclude json

cyrush:
@JustinPrivitera  best practice is to clear results if they exist before the test

cyrush:
Another Q:
When writing, if an alternate file is there from an old write that could cause an issue in the future, should we remove it?

```
my_data.bin
my_data.conduit_json
```
and we now want to write with yaml schema
```
my_data.bin
my_data.conduit_json
my_data.conduit_yaml
```

Should we remove the json file?

JustinPrivitera:
json files are removed before writing yaml, yaml is removed before writing json. I no longer need to modify the tests since they clear conflicting results automatically as a result of this change.

cyrush (approved):
Thanks!

PR #1491, opened 2026-01-05
Silo I/O: Refactor Generate Silo Multi-block data method

JustinPrivitera:
- Refactored the dreadfully coded `generate_silo_mb_data` method.
 - prevented copying of silo mesh and variable types; now types are created once and use directly when writing to silo
 - fixed a bug when checking if all types for a given multi-block object are the same; previously the check would have given the wrong answer if the first domain was empty


cyrush (approved):
Looks good!

PR #1503, opened 2026-01-21
Overlink species set requirements

JustinPrivitera:
Overlink has several rules for writing species sets. We now enforce them.

Also adds species set utilities.
Fixes a bug preventing the writing of multiple species sets to Overlink.


cyrush (approved):
Looks good thanks!

PR #1504, opened 2026-01-24
Changes to topology::compute_mesh_info() that avoid SIGFPE

BradWhitlock:
We observed that `compute_mesh_info()` could issue a SIGFPE on 1-d meshes and polyhedral meshes when using Intel 2025 with C++20. This happened when comparing doubles to `numeric_limits<double>::max()` and `numeric_limits<double>::lowest()`.

This PR:
 * Changes the comparison so it is based on a bool
 * Computes diagonal values for polyhedral zones (face diagonals)


cyrush (approved):
Thanks for the fix. Surprised the compiler was misbehaving, but best to avoid as you have done here.

aaroncblack:
Thanks @BradWhitlock, confirmed it fixes our code and tests also.

PR #1505, opened 2026-01-24
Fix a bug in Node::swap() that affected find_reference_node().

### LINKED ISSUE #1485: Failure with find_reference_node.
I ran across a strange case where `find_reference_node` fails when I move mesh domain node to a different node. I would expect that since the new node took ownership of the mesh data, it would serve as the root and ought to basically be the same as the copy case.
Here is a reproducer:

```
#include <conduit.hpp>
#include <conduit_blueprint_mesh_utils.hpp>

#include <cassert>
#include <iostream>

int main(int argc, char *argv[])
{
  const char *yaml = R"(
mesh:
  coordsets: 
    coords: 
      type: explicit
      values: 
        x: [0.0, 1.0, 2.0, 0.0, 1.0, 2.0, 0.0, 1.0, 2.0]
        y: [0.0, 0.0, 0.0, 1., 1., 1., 2., 2., 2.]
  topologies: 
    topo: 
      type: unstructured
      coordset: coords
      elements:
        shape: quad
        connectivity: [0,1,4,3, 1,2,5,4, 3,4,7,6, 4,5,8,7]
        sizes: [4,4,4,4]
        offsets: [0,4,8,12]
  matsets: 
    mat: 
      topology: topo
      material_map: 
        mat0: 0
        mat1: 1
        mat2: 2
      material_ids: [0, 1, 2, 1, 2]
      volume_fractions: [1., 1., 1., 0.5, 0.5]
      sizes: [1, 1, 1, 2]
      offsets: [0, 1, 2, 3]
      indices: [0, 1, 2, 4, 5]
  fields: 
    elems: 
      topology: topo
      association: element
      values: [0, 1, 2, 3]
  state: 
    cycle: 0
    time: 0.0
    domain_id: 0
)";

  // "Read" the mesh.
  conduit::Node whole;
  whole.parse(yaml);

  // Get the path to the mesh. The path will have the "mesh/" prefix. Ew.
  {
    conduit::Node &mesh = whole["mesh"];
    std::cout << "Path to topologies/topo: " << mesh["topologies/topo"].path() << std::endl;
  }

  { // Use find_reference_node with a copy.
    conduit::Node mesh;
    mesh.set(whole["mesh"]);

    // Get the topo through the matset.
    const conduit::Node *n_topo = conduit::blueprint::mesh::utils::find_reference_node(
     mesh.fetch_existing("matsets/mat"), "topology");
    std::cout << "n_topo = " << (void*)n_topo << std::endl;
    assert(n_topo != nullptr);
    std::cout << "OK!\n";
  }

  { // Use find_reference_node with a moved mesh.
    conduit::Node mesh;
    mesh.move(whole["mesh"]);

    // Get the topo through the matset.
    const conduit::Node *n_topo = conduit::blueprint::mesh::utils::find_reference_node(
     mesh.fetch_existing("matsets/mat"), "topology");
    std::cout << "n_topo = " << (void*)n_topo << std::endl;
    // IT WILL BE NULLPTR AND IT WILL ABORT HERE
    assert(n_topo != nullptr);
    std::cout << "OK!\n";
  }

  return 0;
}
```

BradWhitlock:
This PR resolves #1485

* Changed `Node::swap()` so it makes children that were swapped point at their new parents.
* Added tests for swap/move and `find_reference_node()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 - Added `set` methods to `DataAccessor` that take `DataArray`s and `DataAccessor`s.
 - Added optional device execution support via RAJA and Umpire.
 - Added `conduit_bin_yaml` protocol case to `Node::load()` and `Node::save()`. Also added `conduit_bin_json`, which does the same thing as `conduit_bin` (creates a json schema file to go with the `conduit_bin` file).
+- Fixed an issue in `Node::swap()` or `Node::move()` where child nodes moved to a new parent did not point to their new parent. This caused operations that call `Node::parent()` to traverse upwards to malfunction.
 
 #### Blueprint
 - Finished `bent_multi_grid_amr` mesh by adding adjacency sets between spatially adjacent domains at the same level of refinement.
--- a/src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
@@ -440,7 +440,7 @@ find_reference_node(const Node &node, const std::string &ref_key)
 
     if(node.has_child(ref_key))
     {
-        const std::string &ref_value = node.fetch(ref_key).as_string();
+        const std::string ref_value = node.fetch(ref_key).as_string();
 
         const Node *traverse_node = node.parent();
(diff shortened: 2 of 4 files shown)


cyrush (approved):
Looks great -- thanks for the fix and the new tests.

PR #1506, opened 2026-01-27
Re-address floating point exception in compute_mesh_info()

BradWhitlock:
We ran into more floating point exception trouble with a different compiler inside `compute_mesh_info()`. Some math functions can cause `FE_INEXACT`, which docs say is mostly harmless.

* Added a call to clear the `FE_INEXACT` floating point exception prior to where the code was dying with a SIGFPE.
* Changed some double to float64, just to better match MeshInfo (which uses float64)


cyrush (approved):
Looks good, thanks!

PR #1520, opened 2026-02-05
python warnings pass

### LINKED ISSUE #1518: cast warnings in python interface
For METH_KEYWORDS, use the following cast

```
(PyCFunction)(void*)
```

instead of current 

```
(PyCFunction)
```

cyrush:
resolves #1518


src/libs/blueprint/python/conduit_blueprint_mpi_python.cpp
```diff
@@ -127,24 +114,8 @@ PyBlueprint_MPI_verify(PyObject *, //self
...
-    {
-        rank = relay::mpi::rank(comm);
-    }
-    catch(conduit::Error &e)
-    {
-        PyErr_SetString(PyExc_Exception,
-                        e.message().c_str());
-        return NULL;
-    }
-
```
cyrush, line 143: I want to investigate why these clauses were there.

cyrush: These are there to check for a valid communicator before calling more complex functions.
I will add them back in but not use the return value so we don't get unused var warnings.

cyrush: Added them back with better form

cyrush:
Looks like there is a change to the python mpi behavior in azure tests -- need to find the root cause in azemu.

cyrush:
```

The following tests FAILED:
	108 - t_python_relay_mpi (Failed)
	159 - t_python_blueprint_mpi_mesh (Failed)
	183 - t_conduit_docs_tutorial_python_relay_mpi_examples (Timeout)
Errors while running CTest

```

cyrush:
resolved a typo undermining return of rank result

PR #1521, opened 2026-02-06
Find python dep

mathstuf:


mathstuf:
I'll note that I ran into issues with the Limited API and MPI enabled at the same time; I suspect that's a hole in testing coverage (if it is cared about).

cyrush:
Thank you for working on this, it will be nice to move on from the old logic. 

A few questions: 

The compiled modules still land inside of the folder structure setup by the main python module (`conduit_python_build`). What allows us to skip the that dependency now?

Can we set `PYTHON_FOUND` as the indicator that things are ok instead of using `ENABLE_PYTHON`.

I strive to use `ZZZ_FOUND` for all tpls uniformly.

mathstuf:
Then let's use `Python3_FOUND`?

cyrush:
That sounds good, thanks!

cyrush:
It looks like detection  / finding numpy is failing on windows + msvc:
```
CMake Error at C:/Program Files/CMake/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find Python3 (missing: Python3_NumPy_INCLUDE_DIRS NumPy) (found
  version "3.14.2")
  C:/Program Files/CMake/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
  C:/Program Files/CMake/share/cmake-3.31/Modules/FindPython/Support.cmake:4002 (find_package_handle_standard_args)
  C:/Program Files/CMake/share/cmake-3.31/Modules/FindPython3.cmake:602 (include)
  cmake/Setup3rdParty.cmake:95 (find_package)
  CMakeLists.txt:92 (include)
```

The other failures are due to the spack builds using to old of a cmake.

I have a spack update PR that should resolve those.

cyrush:
> Thank you for working on this, it will be nice to move on from the old logic.
> 
> A few questions:
> 
> The compiled modules still land inside of the folder structure setup by the main python module (`conduit_python_build`). What allows us to skip the that dependency now?
> 
> Can we set `PYTHON_FOUND` as the indicator that things are ok instead of using `ENABLE_PYTHON`.
> 
> I strive to use `ZZZ_FOUND` for all tpls uniformly.

I understand the `conduit_python_build` change now, I mistook this for the pure module python setup, but it's the build step used to pass the limited api defs. 

This week I will recreate the CI env and see if I can get more details about why NumPy is not found.

cyrush:
@mathstuf  The CI issue are due to PYTHON_EXECUTABLE, not Python3_EXECUTABLE being set -- also we need to change the targets used to include/link numpy. 

I am going to merge this into an integration branch and then make some changes to get this sorted.

I'll create a new PR into develop from that branch, and we can discuss further evolution there.

PR #1522, opened 2026-02-19
update to use pyproject toml files

cyrush:


pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+build-type = "Release"
+source-dir = "src"
+
+[tool.scikit-build.cmake.define]
+BUILD_SHARED_LIBS = "OFF"
+ENABLE_PYTHON = "ON"
+ENABLE_TESTS  = "OFF"
+ENABLE_DOCS   = "OFF"
+PYTHON_MODULE_INSTALL_PREFIX = { env = "SKBUILD_PLAT_DIR" }
+CMAKE_INSTALL_PREFIX = { env = "SKBUILD_PLAT_DIR" }
```
cyrush: Note: This isn't working yet, and it's going to be a challenge to adjust the paths with what we have to make a declarative entry

cyrush:
for the pip driving cmake build case, I switched to use `build-backend = "setuptools.build_meta"` along with `setup.py` since there is some path logic that is not possible with declarative

cyrush:
gact pip is 22
```
2026-02-20T18:33:32.9299566Z python3-pip is already the newest version (22.0.2+dfsg-1ubuntu0.7).
```

This isn't new enough -- need pip 24 or later to properly handle our pyproject.toml + setup.py scheme


https://stackoverflow.com/questions/78034052/unknown-project-name-and-version-number-for-my-own-pip-package

cyrush:
also resolves #1523

PR #1524, opened 2026-02-23
sidre mesh load fix

cyrush:


src/libs/relay/conduit_relay_io_blueprint.cpp
```diff
@@ -2433,9 +2433,63 @@ void read_mesh(const std::string &root_file_path,
...
+                std::string outer_name = outer_itr.name();
+
+                // special logic for state, since it may not include paths
+                if(outer_name == "state" )
+                {
+                    if(outer.has_child("path"))
+                    {
+                        const std::string fetch_path =
+                            utils::join_path(conduit_fmt::format("{}",i),
+                                             outer["path"].as_string());
```
JustinPrivitera, line 2455: What is this? Turning `i` into a string? So for `i` = 5 we get `5/the/path/`?

JustinPrivitera: I guess that's how this data is normally structured? Why do we need to use `conduit_fmt` to convert `i` to a string? That isn't as straightforward as just using `std::to_string()`

src/libs/relay/conduit_relay_io_blueprint.cpp
```diff
@@ -2433,9 +2433,63 @@ void read_mesh(const std::string &root_file_path,
...
+                    {
+                        if(outer.has_child("cycle"))
+                        {
+                             mesh_out[outer_name]["cycle"] = outer["cycle"];
+                        }
+
+                        if(outer.has_child("time"))
+                        {
+                            mesh_out[outer_name]["time"] = outer["time"];
+                        }
```
JustinPrivitera, line 2471: So we only care about cycle and time in the case where a path is not provided? Is that right?

src/libs/relay/conduit_relay_io_blueprint.cpp
```diff
@@ -2433,9 +2433,63 @@ void read_mesh(const std::string &root_file_path,
...
+                while(itr.has_next())
+                {
+                    const Node &entry = itr.next();
+                    if(entry.has_child("path"))
+                    {
+                        const std::string entry_name = itr.name();
+                        const std::string entry_path = entry["path"].as_string();
+                        const std::string fetch_path =
+                            utils::join_path(conduit_fmt::format("{}",i),
+                                             entry_path);
```
JustinPrivitera, line 2485: Same thing here about using conduit_fmt to make a string out of an integer being overkill

JustinPrivitera (commented):
Do all the tests pass? Or maybe we have no tests for this case? I know you said this worked with sample data that you used.

src/libs/relay/conduit_relay_io_blueprint.cpp
cyrush: not sure it matters much for these cases - but lore states that fmt is faster than std::to_string()

src/libs/relay/conduit_relay_io_blueprint.cpp
cyrush: `state/cycle` and `state/time` are optional, but directly in the index if they exist.

There is no need to read them via a path-based pattern

src/libs/relay/conduit_relay_io_blueprint.cpp
JustinPrivitera: I trust lore

Issue #1514, opened 2026-01-31
Generalized way to walk material sets


JustinPrivitera:
we need a more general way to walk matsets; no more bespoke solutions

PR #1527, opened 2026-03-04
Matset Accessor

JustinPrivitera:
Adds `MatsetAccessor` class that can be used for extracting material data. It mostly abstracts away the matset layout differences. It also extracts field and species data in the same manner. Examples of use are included in the data retrieval tests.

This partially addresses #1514

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh_matset_accessor.hpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_matset_accessor.hpp
@@ -71,11 +71,11 @@ public:
 //-----------------------------------------------------------------------------
 using GetMatIdPtr        = index_t (MatsetAccessor::*)(index_t, index_t) const;
 using GetElemIdPtr       = index_t (MatsetAccessor::*)(index_t, index_t) const;
-using GetVolFracPtr      = double  (MatsetAccessor::*)(index_t, index_t) const;
-using GetMsetValPtr      = double  (MatsetAccessor::*)(index_t, index_t) const;
-using GetMassFracPtr     = double  (MatsetAccessor::*)(index_t, index_t, index_t) const;
-using GetNMatsForZonePtr = index_t (MatsetAccessor::*)(index_t) const;
-using GetNZonesForMatPtr = index_t (MatsetAccessor::*)(index_t) const;
+using GetVolFracPtr      = float64 (MatsetAccessor::*)(index_t, index_t) const;
+using GetMsetValPtr      = float64 (MatsetAccessor::*)(index_t, index_t) const;
+using GetMassFracPtr     = float64 (MatsetAccessor::*)(index_t, index_t, index_t) const;
+using GetNMatsForElemPtr = index_t (MatsetAccessor::*)(index_t) const;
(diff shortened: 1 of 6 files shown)


src/libs/blueprint/conduit_blueprint_mesh_matset_accessor.hpp
```diff
@@ -0,0 +1,394 @@
...
+//-----------------------------------------------------------------------------
+    inline
+    index_t     get_mat_id(const index_t zone_idx,
+                           const index_t mat_idx) const
+    {
+        return (this->*m_get_mat_id)(zone_idx, mat_idx);
+    }
+
+    inline
+    index_t     get_elem_id(const index_t zone_idx,
```
cyrush: we are mixing elem and zone, can we clarify names?

src/libs/blueprint/conduit_blueprint_mesh_matset_accessor.hpp
```diff
@@ -0,0 +1,394 @@
...
+// -- conduit::blueprint::mesh::matset::MatsetAccessor public members --
+//
+//-----------------------------------------------------------------------------
+
+//-----------------------------------------------------------------------------
+/// MatsetAccessor Function pointer types
+//-----------------------------------------------------------------------------
+using GetMatIdPtr        = index_t (MatsetAccessor::*)(index_t, index_t) const;
+using GetElemIdPtr       = index_t (MatsetAccessor::*)(index_t, index_t) const;
+using GetVolFracPtr      = double  (MatsetAccessor::*)(index_t, index_t) const;
```
cyrush: should these be float64?

float64 is used later in the interface

src/libs/blueprint/conduit_blueprint_mesh_matset_accessor.hpp
```diff
@@ -0,0 +1,394 @@
...
+    {
+        return m_num_mats;
+    }
+
+//-----------------------------------------------------------------------------
+/// Get information about the sizes
+//-----------------------------------------------------------------------------
+    // use this for elem-dom sparse matsets
+    inline
+    index_t     num_mats_for_zone(const index_t zone_idx) const
```
cyrush: zone vs element 

(blueprint prefers `element`)

cyrush (commented):
Looks great, a few questions about naming.

src/libs/blueprint/conduit_blueprint_mesh_matset_accessor.hpp
JustinPrivitera: do you have a preference?

src/libs/blueprint/conduit_blueprint_mesh_matset_accessor.hpp
JustinPrivitera: they should be; my mistake!

src/libs/blueprint/conduit_blueprint_mesh_matset_accessor.hpp
JustinPrivitera: sounds good, element it is.

PR #1539, opened 2026-03-07
Add Matset Walkers

### LINKED ISSUE #1514: Generalized way to walk material sets

JustinPrivitera:
These will not be part of the public API but will be used for rewriting to_silo and the other material set transformations.

Resolves #1514


cyrush (approved):
Looks good!

PR #1541, opened 2026-03-10
Matset converters now support the existence of the 4th matset layout type

JustinPrivitera:
- API change: `to_multi_buffer_full` -> `to_multi_buffer_by_element`
- Added `to_uni_buffer_by_material`, which just throws an error
- All converter functions now error if they detect a uni-buffer by material matset.

Note: we don't support the 4th matset type, we just handle the case where someone gives it to us by throwing an error instead of doing the wrong thing.


cyrush (approved):
Looks good!

PR #1543, opened 2026-03-10
misc changes

cyrush:


HUMANS.md
JustinPrivitera, line 1: is this a primer for agentic tools to come up to speed?

JustinPrivitera (approved):

PR #1545, opened 2026-03-17
Add vertex-based adjsets to to_polygonal

nselliott:
The implementation was missing the needed adjsets on the output of the transformation from structured AMR domains to the polygonal mesh.


nselliott:
I have put additional changes and features into this PR.  These include:

- Documentation and verification of the structured window adjset variant that I developed to enable the entire to_polytopal feature for AMR meshes.  These adjsets describe adjacency in terms of the structured index space, rather than storing shared vertices.
- Full generation of unstructured shared-index adjsets on the output mesh of to_polytopal.
- A new generate_boundary function that adds an edge or face-based boundary topology to the unstructured mesh.  This is not part of to_polytopal, but can be called afterward to add this topology when needed.
- Handling of ranks that have no domains to provide to Partitioner::combine(), which can happen in AMR mesh flattening situations.

nselliott:
I'm not sure it's time to remove the Draft label, but it may be close, and folks can take a look at it.

PR #1547, opened 2026-03-18
Rewrite Matset/Field/Specset Conversion Routines using MatsetAccessors and Walkers

### LINKED ISSUE #1497: Relax material ordering restriction for material dependent field, specset, and matset style converters
Material dependent field style converters need not present materials in the same order as their associated matset. The same is true of specsets. Yet there is an artificial restriction imposed by the material dependent field and specset style converters.

### LINKED ISSUE #1507: Material set conversion routines need to be sensitive to optionally included material map
Currently, the material set conversion routines expect that there will only be a material map in the case of uni buffer element dominant (sparse by element) data. Therefore, a material map is created as needed. But the Blueprint specification allows for there to be material maps in all material set representations, so we must handle the case that a material map is provided and honor it, as it may be different from the default constructed map we create.

### LINKED ISSUE #1511: Support material dependent fields for matset field converters

JustinPrivitera:
Resolves #1507
Resolves #1497 
Resolves #1511


cyrush (approved):
Looks great!

PR #1556, opened 2026-03-26
Silo I/O: Fix an issue with topology connectivity int width

JustinPrivitera:
tested with data that originally caused the issue, but I cannot add that test data to Conduit.

Also be consistent about passing int arrays to silo calls.


cyrush (approved):
Thanks for the fix!

PR #1560, opened 2026-03-31
Reduce Parmetis output from generate_partition_field()

BradWhitlock:
The `generate_partition_field()` function can generate a *LOT* of console output in parallel. The function takes a `conduit::Node` of options so I added an option called "verbose" that will cause the Parmetis output to be printed. When the option is missing or set to 0, no Parmetis output is printed to the console, making it more compatible with very large runs.


JustinPrivitera (approved):

cyrush (approved):
Looks good, thanks for this change!

PR #1561, opened 2026-04-01
Tiled mesh fixes

BradWhitlock:
This PR fixes a problem generating a 1x1x1 tiled mesh.

* When 1x1x1 was requested with 2 domains, it caused infinite recursion when trying to split the tile block
* Fixed an issue with this mesh where the Z boundary faces were not being added
* Added a test for the 1x1x1 tile generation case
* Fixed an issue with merging 'MeshInfo` objects that might contain zero-length edge and diagonal lengths. This was causing a test to throw an exception from the quantizer. I think some domains had zero-length edges b/c of a spherical/polar coordinate transformation and then the merge caused those zero lengths to be used. Fixed now.


cyrush (approved):
Thanks for the fixes.

PR #1572, opened 2026-04-13
Device Support 1: CMake scaffolding

JustinPrivitera:
authored mostly by codex


JustinPrivitera:
We went with a different approach, see #1358

PR #1583, opened 2026-04-20
add guard to avoid issues errors with multiple find_package conduit c…

### LINKED ISSUE #1582: cmake targets creation is not guarded
Just for the sake of reproducing the problem easily, this is the `CMakeLists.txt`

```cmake
cmake_minimum_required(VERSION 3.21)
project(test C CXX)

# this double call is the relevant part
find_package(Conduit REQUIRED)
find_package(Conduit REQUIRED)
```

This situation is quite common, since the `find_package` call might be done not just in the "root" project, but also in its dependencies. However, overall the behavior is the same as calling twice in the root project as done in the reproducer above.

Everything works fine if called once, and it was not present in release `0.9.5`. The problem shows up as

```console
CMake Error at /user-environment/env/default/lib/cmake/conduit/conduit_setup_targets.cmake:65 (add_library):
  add_library cannot create imported target "conduit::conduit" because
  another target with the same name already exists.
  /user-environment/env/default/lib/cmake/conduit/ConduitConfig.cmake:86 (include)
  CMakeLists.txt:5 (find_package)


CMake Error at /user-environment/env/default/lib/cmake/conduit/conduit_setup_targets.cmake:83 (add_library):
  add_library cannot create imported target "conduit::conduit_python" because
  another target with the same name already exists.
  /user-environment/env/default/lib/cmake/conduit/ConduitConfig.cmake:86 (include)
  CMakeLists.txt:5 (find_package)


CMake Error at /user-environment/env/default/lib/cmake/conduit/conduit_setup_targets.cmake:101 (add_library):
  add_library cannot create imported target "conduit::conduit_mpi" because
  another target with the same name already exists.
  /user-environment/env/default/lib/cmake/conduit/ConduitConfig.cmake:86 (include)
  CMakeLists.txt:5 (find_package)

```

where it complains that targets that are going to be defined by `conduit_setup_targets` were already defined (by the previous call of `find_package`).

I preferred to open an issue instead of a PR with a potential fix, because it does not look like the "classic and automatic" cmake management of targets (`install(TARGETS ... EXPORT)`), so I might be missing some relevant information. If you give me any pointers, I can eventually try to fix it and open a PR.

cyrush:
…alls

resolves #1582

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 
 #### Conduit
 - Fixed an issue with fmt runtime use of a string that causes a compile error with gcc 13 and c++20.
+- Add guard to prevent multiple calls of `find_package(Conduit)` from failing in downstream CMake logic. Guard logic existed in releases prior to 0.9.6, now restored.
 
 #### Blueprint
 - Fixed missing install of blueprint non-mpi mesh examples and mesh partition c++ headers.
--- a/src/config/ConduitConfig.cmake.in
+++ b/src/config/ConduitConfig.cmake.in
@@ -16,6 +16,8 @@ cmake_minimum_required(VERSION 3.8 FATAL_ERROR)
 
 @PACKAGE_INIT@
 
+if(NOT Conduit_FOUND)
+
 # Compute the installation prefix relative to this file.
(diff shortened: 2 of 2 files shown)


albestro:
tested, it works! thanks for the quick fix!

cyrush:
@albestro glad this helps, thanks for testing!

PR #1590, opened 2026-04-27
avoid downstream dep search for features conduit was not built with

cyrush:


cyrush:
@JustinPrivitera can you give this another set of eyes

JustinPrivitera:
> @JustinPrivitera can you give this another set of eyes

looking now

src/config/conduit_setup_targets.cmake
```diff
@@ -24,8 +24,10 @@ set(CONDUIT_INCLUDE_DIRS "${_IMPORT_ROOT}/include/conduit")
 
 # check for conduit fortran support
 if(EXISTS ${_IMPORT_ROOT}/include/conduit/conduit.mod)
+    set(CONDUIT_USE_FORTRAN TRUE)
     set(CONDUIT_FORTRAN_ENABLED TRUE)
```
JustinPrivitera, line 28: what is the difference between CONDUIT_USE_FORTRAN and CONDUIT_FORTRAN_ENABLED?

JustinPrivitera (approved):

src/config/conduit_setup_targets.cmake
cyrush: nothing, but I am trying to normalize on providing the `USE` variants since that style has become used more

JustinPrivitera: why not remove CONDUIT_FORTRAN_ENABLED? Or is that outside the scope of this work?

cyrush: Legacy -- folks may be testing for it in their build system.

PR #1592, opened 2026-04-28
Device Support PR Feedback: Fix atomics and reducers

JustinPrivitera:
- fix atomics
 - fix reducers

### DIFF
--- a/src/libs/conduit/conduit_execution.hpp
+++ b/src/libs/conduit/conduit_execution.hpp
@@ -79,21 +79,6 @@ namespace execution
 namespace detail
 {
 
-template <typename ExecPolicyTag, typename T>
-using ReduceSumImpl = RAJA::ReduceSum<typename ExecPolicyTag::reduce_policy, T>;
-
-template <typename ExecPolicyTag, typename T>
-using ReduceMinImpl = RAJA::ReduceMin<typename ExecPolicyTag::reduce_policy, T>;
-
-template <typename ExecPolicyTag, typename T>
-using ReduceMinLocImpl = RAJA::ReduceMinLoc<typename ExecPolicyTag::reduce_policy, T>;
-
-template <typename ExecPolicyTag, typename T>
-using ReduceMaxImpl = RAJA::ReduceMax<typename ExecPolicyTag::reduce_policy, T>;
-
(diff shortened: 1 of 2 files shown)


src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -544,47 +544,52 @@ TEST(conduit_execution, test_reductions)
...
-        conduit::execution::ReduceMinLoc<index_t> minloc_reducer(policy);
-        conduit::execution::forall(policy, 0, size, [=] CONDUIT_EXEC(index_t i)
-        {
-            minloc_reducer.minloc(vals_ptr[i], i);
-        });
-        CONDUIT_DEVICE_ERROR_CHECK(policy);
-        EXPECT_EQ(minloc_reducer.get(), -10);
-        EXPECT_EQ(minloc_reducer.getLoc(), 1);
+            conduit::execution::ReduceMin<Exec, index_t> min_reducer;
+            conduit::execution::forall<Exec>(0, size, [=] CONDUIT_EXEC(index_t i)
```
JustinPrivitera, line 567: I don't think all of this should be in one big dispatch

src/libs/conduit/conduit_execution.hpp
```diff
@@ -164,35 +149,181 @@ atomic_max_exec(T *acc, T value)
...
+inline void
+forall_exec(ExecPolicyTag,
+            const int& begin,
+            const int& end,
+            Kernel&& kernel) noexcept
+{
+    for (int i = begin; i < end; i ++)
+    {
+        kernel(i);
+    }
```
JustinPrivitera, line 200: why did all this move?

src/libs/conduit/conduit_execution.hpp
```diff
@@ -379,802 +536,62 @@ class ReduceMaxLocImpl
...
-            const int& begin,
-            const int& end,
-            Kernel&& kernel) noexcept
-{
-    for (int i = begin; i < end; i ++)
-    {
-        kernel(i);
-    }
-}
+#endif
```
JustinPrivitera, line 553: need comments explaining where we are using raja and where we are not

PR #1594, opened 2026-05-01
use conduit specific define to ensure fmt header only and exception use

cyrush:


cyrush:
@Arlie-Capps  let me know if these fixes resolve the fmt issues you encountered

Arlie-Capps (approved):
Looks great!

PR #1595, opened 2026-05-06
Silo/Overlink I/O: Fix crash with null material name

### LINKED ISSUE #1593: Silo/Overlink I/O: crash when reading file with no material names
data is on RZ under my `user_support/help_jacob` directory

```python
import conduit

silo_root = "***DumpWithColors.silo"

load_mesh = conduit.Node()
conduit.relay.io.silo.load_mesh(load_mesh, silo_root)

print(load_mesh)
```

```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
```

Unclear if we should error, warn and skip, or make up a name, but any option would be better than crashing like this.

JustinPrivitera:
Resolves #1593


src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -3825,7 +3825,14 @@ read_multimats(DBtoc *toc,
...
                 {
-                    material_map[matnames[i]] = matnos[i];
+                    if (nullptr != matnames[i])
+                    {
+                        material_map[matnames[i]] = matnos[i];
+                    }
+                    else
+                    {
+                        material_map[std::to_string(matnos[i])] = matnos[i];
+                    }
```
cyrush, line 3835: Looks good, thanks!

cyrush (approved):

PR #1603, opened 2026-05-12
github actions windows ci heal

cyrush:


cyrush:
addressing env drift in github actions runner, lead to a linking issue with umpire and a find package issue with zlib

cyrush:
Something in the runner env has resulted in our paths been transformed 

from:

```
D:/path/to/stuff
```

to:

```
/d/path/to/stuff
```

I don't think this happens with my manual windows testing, but will confirm soon.

cyrush:
We are relying on cygpath to fix these paths for us, we using `-m`

```
-m, --mixed: Convert to Windows format with forward slashes.
```

Perhaps this is now missing or different due to env update

cyrush:
The issue was env changed was not longer being detected as windows:
```
2026-05-12T19:57:33.7896686Z build_windows=OFF
```

Added a case for `OSTYPE=cygwin` to map to windows.

PR #1607, opened 2026-05-15
add mesh bp remove util

cyrush:
add `conduit::mesh::blueprint::remove`, which allows you to remove components of a mesh (like topologies, fields, etc) along with there dependencies. 

For example, removing a topology will remove any fields defined on that topology, removing a matset will remove references and matset specific values for any related fields.


cyrush:
TODO: Test cases are WIP, need to add many more:
- [x] add multi domain test cases
- [x] add matset, adjset, and nestset test cases
- [x] make sure matset test cases include a field that with matset_values
- [x] think about other bad options scenarios

cyrush:
presence of * as category name indicates that we should remove everything from that category
```
fields: ["den","radial"]
matsets: *
state: *
```

will fully remove `state` and all `matsets`, along with the fields `den` and `radial`


for state, other than `*`, any entry is assumed to be a path
```
fields: ["den","radial"]
matsets: *
state: time
```

cyrush:
added specset test and python interface and python test

JustinPrivitera:
There are many places we could leverage this (perhaps in Silo tests) that would provide additional testing coverage.

src/tests/blueprint/t_blueprint_mesh_remove.cpp
```diff
@@ -0,0 +1,457 @@
...
+        echo_node("[before]",n_mesh,false);
+        echo_node("[options]",n_opts);
+        conduit::blueprint::mesh::remove(n_opts,n_mesh);
+        EXPECT_TRUE(conduit::blueprint::mesh::verify(n_mesh, n_info));
+        echo_node("[after]",n_mesh,false);
+        echo_node("info",n_info);
+
+        EXPECT_TRUE(n_mesh.has_path("matsets/mesh"));
+        EXPECT_FALSE(n_mesh.has_path("specsets/mesh"));
+        return;
```
cyrush, line 305:
```suggestion
```

JustinPrivitera (commented):
Reviewing now. Are you going to add a changelog entry?

cyrush:
Thanks, yes I will add Change log entry

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -9172,6 +9173,285 @@ void mesh::convert(const conduit::Node &n_mesh, const conduit::Node &n_options,
...
+    }
+
+    // adjsets and nestsets don't have downstream dependent relationships
+
+    // fields need special logic to remove matset values, that is handled after higher
+    // level tree pruning
+
+    // helper to record dependent components to a set that will be used
+    // in the final removal process
+    auto remove_components = [&](conduit::Node *dom,
```
JustinPrivitera:
```suggestion
    // helper to remove dependent components to a set
    auto remove_components = [&](conduit::Node *dom,
```

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -9172,6 +9173,285 @@ void mesh::convert(const conduit::Node &n_mesh, const conduit::Node &n_options,
...
+                    {
+                        comp_pat_adj = ".*";
+                    }
+                    // std::cout << "checking \"" << comp_name << "\" vs \""  << comp_pat_adj << "\" : ";
+                    std::regex comp_regex(comp_pat_adj);
+                    if(std::regex_match(comp_name, comp_regex))
+                    {
+                        //std::cout << "true" << std::endl;
+                        n_comp.remove(comp_name);
+                    }
```
JustinPrivitera: I don't understand why we need to do regex match again. We already used it earlier to populate the set, so can't we just remove everything in the set unconditionally?

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -9172,6 +9173,285 @@ void mesh::convert(const conduit::Node &n_mesh, const conduit::Node &n_options,
...
+        remove_components(dom, "nestsets", nestsets);
+    }
+
+    // past to clean up matsets
+    // matsets: for any field that refs a matset,
+    //          remove that ref and any matset values
+
+    for(size_t i = 0; i < domains.size(); i++)
+    {
+        conduit::Node *dom = domains[i];
```
JustinPrivitera, line 9384: Why not remove fields that depend on a matset entirely? The field may or may not be in a valid state after the operation you've implemented here. If the field is material-dependent, then it will still be valid, but if it is mixed (https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#fields), then it is no longer valid and should be removed entirely.

src/tests/blueprint/t_blueprint_mesh_remove.cpp
```diff
@@ -0,0 +1,462 @@
...
+        echo_node("[before]",n_mesh);
+        echo_node("[options]",n_opts);
+        conduit::blueprint::mesh::remove(n_opts,n_mesh);
+        EXPECT_TRUE(conduit::blueprint::mesh::verify(n_mesh, n_info));
+        echo_node("[after]",n_mesh);
+        echo_node("[info]",n_info);
+
+        // removing the matset should remove the specset as well
+        EXPECT_TRUE(n_mesh.has_path("matsets/mesh"));
+        EXPECT_FALSE(n_mesh.has_path("specsets/mesh"));
```
JustinPrivitera, line 328:
```suggestion
        // removing the matset should remove the specset as well
        EXPECT_FALSE(n_mesh.has_path("matsets/mesh"));
        EXPECT_FALSE(n_mesh.has_path("specsets/mesh"));
```

JustinPrivitera: looks like this was copied from above and not updated to reflect the comment. It doesn't test what the comment says it does.

src/tests/blueprint/t_blueprint_mesh_remove.cpp
JustinPrivitera, line 1: currently we can either give a name or give `"*"`? Do we want to support "`*_name`" or any other regex stuff?

JustinPrivitera (changes requested):
Love the lambdas. I have some comments about a few things.

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: comment looks like it was copied from above and not updated

src/libs/blueprint/conduit_blueprint_mesh.cpp
cyrush: fully material-dependent is rare, usually `values` exists and is useful even without the matset values.  That said, I this should check for that case. 

Matset values can be very large, removing them when not needed is a savings if folks only want to save or process the overall field values.

src/libs/blueprint/conduit_blueprint_mesh.cpp
cyrush: The set contains the regex, not the expanded names. 

The prior walk explicitly records dependents, but the outer matches are not explicitly recorded.

cyrush:
For filtering, what we really want is glob matching, not regex matching. 

We can convert glob style matching into a regex, I will add a utility to do that. 

This will also be useful in other cases (extending mesh blueprint read to read subsets of a mesh with same style of args used here)

src/tests/blueprint/t_blueprint_mesh_remove.cpp
cyrush: See my comment about glob vs regex, I will work on support for that so that `*` isn't a special case

src/tests/blueprint/t_blueprint_mesh_remove.cpp
cyrush: It is a copy and paste problem -- but the above comment is what is wrong. 
In this case we are removing the specset, that matset should still exist.

src/tests/blueprint/t_blueprint_mesh_remove.cpp
```diff
@@ -0,0 +1,462 @@
...
+        n_opts["specsets"] = "mesh";
+
+        echo_node("[before]",n_mesh);
+        echo_node("[options]",n_opts);
+        conduit::blueprint::mesh::remove(n_opts,n_mesh);
+        EXPECT_TRUE(conduit::blueprint::mesh::verify(n_mesh, n_info));
+        echo_node("[after]",n_mesh);
+        echo_node("[info]",n_info);
+
+        // removing the matset should remove the specset as well
```
cyrush, line 326:
```suggestion
        // after removing specset, matset should remain
```

src/libs/blueprint/conduit_blueprint_mesh.cpp
cyrush: - [ ]  check for material-dependent case (no values), and remove field if when no useful data is left over.

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: I see. The set gets populated even further above in `init_set_from_opts`.

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: I think it is sufficient to check for the existence of values after scrubbing the matset association, and remove if there are no values. However, fields are handled differently from everything else in `remove`. Maybe there is an option to aggressively remove fields versus conservatively removing them? It just seems like an unexpected behavior of the function. If I am removing a matset would I not want associated fields to also be removed? I think both choices are valid.

src/tests/blueprint/t_blueprint_mesh_remove.cpp
JustinPrivitera: No, look closer - this test is copied from the case right before it. Right now both tests are identical. It makes sense to test three cases for specsets:
1. removing a specset, which is tested twice
2. removing a matset which should remove the specset also, which your comment suggested you wanted to test, but is not tested as things currently stand
3. removing a topo which should remove the matset which should remove the specset, which is tested at the end

I think you want to keep the original comment and fix the options that are tested and the resulting EXPECT_***.

src/tests/blueprint/t_blueprint_mesh_remove.cpp
```diff
@@ -0,0 +1,462 @@
...
+        echo_node("[before]",n_mesh);
+        echo_node("[options]",n_opts);
+        conduit::blueprint::mesh::remove(n_opts,n_mesh);
+        EXPECT_TRUE(conduit::blueprint::mesh::verify(n_mesh, n_info));
+        echo_node("[after]",n_mesh);
+        echo_node("[info]",n_info);
+
+        // after removing specset, matset should remain
+        EXPECT_TRUE(n_mesh.has_path("matsets/mesh"));
+        EXPECT_FALSE(n_mesh.has_path("specsets/mesh"));
```
JustinPrivitera, line 328: this test should be updated to test removing a matset: https://github.com/llnl/conduit/pull/1607#discussion_r3282756485

src/tests/blueprint/t_blueprint_mesh_remove.cpp
cyrush: thank, you I think have this sorted out now. Can you take one more look?

src/tests/blueprint/t_blueprint_mesh_remove.cpp
cyrush: I think this is sorted, please give it one more look (sorry for the double comment)

cyrush:
I created a glob match util and remove refactored to use it.

src/libs/blueprint/conduit_blueprint_mesh.cpp
cyrush: The relationship between fields and matsets is a bit different.

I think removing the entire field is too aggressive in this case b/c it removes data that is still useful.
Most viz uses cases will use values and never touch matset values unless the user is doing something sophisticated. 

They can always remove any field fully by name or pattern.

src/tests/blueprint/t_blueprint_mesh_remove.cpp
JustinPrivitera: yes looks good

src/tests/blueprint/t_blueprint_mesh_remove.cpp
JustinPrivitera: yes looks good

JustinPrivitera (approved):

PR #1608, opened 2026-05-15
Partitioner: handle field `matset_values`

### LINKED ISSUE #1579: partitioner seems to blow away field matset_values
data/reproducer on the closed side

blueprint partitioner seems to blow away material-dependence for fields

need to make some example data and figure this out

JustinPrivitera:
Resolves: #1579

handles partitioning matset_values for the split and combine cases.

Talk to Lila once this is merged


src/libs/blueprint/conduit_blueprint_mesh_matset_accessor.cpp
```diff
@@ -585,12 +605,26 @@ MatsetAccessor::init(const Node &matset,
...
+            else
+            {
+                m_multi_vol_fracs.push_back(m_internal_data["zeroes"].value());
+            }
 
             if (nullptr != field)
             {
                 // save matset values array
-                m_multi_mset_vals.push_back((*field)["matset_values"][matname].value());
+                if ((*field)["matset_values"].has_child(matname))
```
cyrush: how about:
```suggestion
                if (field->fetch_existing("matset_values").has_child(matname))
```

cyrush (commented):
This looks good, very minor pointer method call suggestion. 
(could be applied to a few other places)

cyrush:
Thanks for fixing this!

PR #1615, opened 2026-05-26
add mesh bp rename util

cyrush:


src/libs/blueprint/conduit_blueprint_mesh.hpp
```diff
@@ -428,6 +429,46 @@ void CONDUIT_BLUEPRINT_API remove(const conduit::Node &n_options,
...
+ * @brief Rename specific components of mesh and update their dependent
+ *        entires.
+ *
+ *        For example, rename topology to `my_topo` and update
+ *        all associated fields
+*
+ * @param n_mesh The node containing the input mesh (coordsets, topos,
+ *               fields, matsets, specsets, adjsets, and nestsets)
+ *
+ * @param n_options A node containing options. Categories entries
```
JustinPrivitera:
```suggestion
 * @param n_options A node containing options. Entries
```

src/libs/blueprint/conduit_blueprint_mesh.hpp
```diff
@@ -428,6 +429,46 @@ void CONDUIT_BLUEPRINT_API remove(const conduit::Node &n_options,
...
+ *   matsets:
+ *       matset_old : matset_new
+ *   specsets:
+ *       specset_old : specset_new
+ *   adjsets:
+ *       adjset_old : adjset_new
+ *   nestsets:
+ *       nestset_old : nestset_new
+ *
+ * @endverbatim
```
JustinPrivitera, line 472: can we ever have multiple things to rename in a given category? Maybe we rename multiple fields?

src/tests/blueprint/t_blueprint_mesh_rename.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+//-----------------------------------------------------------------------------
+TEST(conduit_blueprint_mesh_rename, bad_options)
+{
+    echo_title("remove bad options");
+    conduit::Node n_mesh;
+    conduit::Node n_opts;
+    n_opts["topologies"] = 1;
+    EXPECT_THROW(conduit::blueprint::mesh::rename(n_opts,n_mesh),conduit::Error);
+    // entries are current name to desired naem
```
JustinPrivitera:
```suggestion
    // entries are current name to desired name
```

src/libs/blueprint/conduit_blueprint_mesh.hpp
JustinPrivitera: looks like we can, so should this docstring be updated to make that clearer?

JustinPrivitera (approved):

src/libs/blueprint/conduit_blueprint_mesh.hpp
cyrush: Yes, I should add an example that shows this

- [ ] add example

PR #1619, opened 2026-06-12
Fix NVCC extended lambda compilation error

siramok:
The project currently encounters a compilation error when building with CUDA enabled, seemingly due to an incompatibility with GTest and CUDA lambdas.

This fix came from AI and while I've verified that it does workaround the compilation problem, it's a bit invasive and I haven't yet verified that it leaves the tests in an equivalent state.


siramok:
Closing because we've since implemented better fixes.

PR #1622, opened 2026-06-12
Device Support: Global Execution Settings API Improvement & Tests

JustinPrivitera:
- Beefed up the API for setting global execution settings.
- Changed some names to make things clearer.
- Added tests for the global execution settings

### DIFF
--- a/src/tests/conduit/t_conduit_execution.cpp
+++ b/src/tests/conduit/t_conduit_execution.cpp
@@ -88,6 +88,255 @@ TEST(conduit_execution, policy_aliases)
 }
 
 //-----------------------------------------------------------------------------
+TEST(conduit_execution, execution_settings)
+{
+    // container for opts whenever we fetch them
+    Node get_opts;
+
+    // if new allocators are added earlier, then these values could change
+    const index_t DEVICE_ALLOC_ID = 1;
+    const index_t HOST_ALLOC_ID = 2;
+
+    // put nodes on host and device
+    Node host_data, device_data;
+
+    // ------------------------------
+    //
+    // Test host and device allocators
+    //
+    // ------------------------------
+    {
+        const index_t host_alloc_id = execution::get_host_allocator_id();
+        const index_t device_alloc_id = execution::get_device_allocator_id();
+        host_data.set_allocator(host_alloc_id);
+        device_data.set_allocator(device_alloc_id);    
+        const std::vector<float64> src_vals = make_execution_src_vals(EXECUTION_TEST_ARRAY_SIZE);
+        host_data.set(src_vals);
+        device_data.set(src_vals);
+
+        // test that allocators are what we expect
+        EXPECT_EQ(device_alloc_id, DEVICE_ALLOC_ID);
+        EXPECT_EQ(host_alloc_id, HOST_ALLOC_ID);
+
+        // prove that memory is alloc'd in the right spot
+        EXPECT_FALSE(execution::DeviceMemory::is_device_ptr(host_data.data_ptr()));
+        EXPECT_TRUE(execution::DeviceMemory::is_device_ptr(device_data.data_ptr()));
+    }
+
+
+    
+    // ------------------------------
+    //
+    // test default settings
+    //
+    // ------------------------------
+    {
+        Node default_opts;
+        execution::execution_options(default_opts);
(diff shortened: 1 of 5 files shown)


src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -87,6 +87,255 @@ TEST(conduit_execution, policy_aliases)
...
+    // ------------------------------
+    //
+    // Test Execution Policy settings
+    //
+    // ------------------------------
+
+    // test bad execution policy
+    {
+        Node exec_opts;
+        exec_opts["execution_policy"] = "banana";
```
siramok, line 160: Nice.

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -87,6 +87,255 @@ TEST(conduit_execution, policy_aliases)
...
+        execution::reset_execution_options();
+    }
+
+    // test input_location exec policy
+    {
+        Node exec_opts;
+        exec_opts["execution_policy"] = "input_location";
+        execution::execution_set_options(exec_opts);
+        execution::execution_options(get_opts);
+        EXPECT_THROW(execution::get_execution_policy(), conduit::Error);
```
siramok, line 206: Why does using "input_location" as the execution policy throw `conduit::Error`?

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -87,6 +87,255 @@ TEST(conduit_execution, policy_aliases)
...
+    // test bad execution policy
+    {
+        Node exec_opts;
+        exec_opts["execution_policy"] = "banana";
+        EXPECT_THROW(execution::execution_set_options(exec_opts),
+                     conduit::Error);
+        execution::reset_execution_options();
+    }
+
+    // test host exec policy
```
siramok, line 166: It seems that these next 3 blocks are largely identical, maybe it'd be worth looping over a vector of the device policy strings to save on boilerplate?

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -87,6 +87,255 @@ TEST(conduit_execution, policy_aliases)
...
+    {
+        Node exec_opts;
+        exec_opts["output_allocator"] = "banana";
+        EXPECT_THROW(execution::execution_set_options(exec_opts),
+                     conduit::Error);
+        execution::reset_execution_options();
+    }
+
+    // test host output allocator
+    {
```
siramok, line 232: Same feedback here as above wrt looping over a vector of device policy strings.

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -87,6 +87,255 @@ TEST(conduit_execution, policy_aliases)
...
+    // test bad sync strategy
+    {
+        Node exec_opts;
+        exec_opts["sync_strategy"] = "banana";
+        EXPECT_THROW(execution::execution_set_options(exec_opts),
+                     conduit::Error);
+    }
+
+    // test sync sync strategy
+    {
```
siramok, line 313: Same feedback as above, although 2 written out cases isn't the worst.

siramok (commented):
The content looks good, just a question + feedback that doesn't necessarily have to be addressed.

src/tests/conduit/t_conduit_execution.cpp
JustinPrivitera: They are subtlety different; they create "host" versus "device" and then EXPECT_TRUE for if it is a host or device policy. Then the third one, "input_location", has to be different because it requires input data in order to create a policy.

src/tests/conduit/t_conduit_execution.cpp
JustinPrivitera: We want to enforce that you must provide input data in order to use the "input_location" setting. "input_location" is for the case where users don't care where things run they just want things to run where their data already is. But data must be provided in order to know where that data is.

src/tests/conduit/t_conduit_execution.cpp
JustinPrivitera: it's not just the policy strings that are changing each time, but fair point.

PR #1623, opened 2026-06-12
Add initial sort implementation

siramok:
Uses RAJA to implement sorting for conduit's execution model.

### DIFF
--- a/scripts/build_conduit/2026_06_15_raja_sort_partition_ambiguity_fix.patch
+++ b/scripts/build_conduit/2026_06_15_raja_sort_partition_ambiguity_fix.patch
@@ -1,0 +1,11 @@
+--- a/include/RAJA/util/sort.hpp
++++ b/include/RAJA/util/sort.hpp
+@@ -389,7 +389,7 @@
+     }
+ 
+     // partition
+-    mid = partition(begin, last, [&](Iter it) {
++    mid = detail::partition(begin, last, [&](Iter it) {
+       return comp(*it, *pivot);
+     });
+ 
--- a/scripts/build_conduit/build_conduit.sh
+++ b/scripts/build_conduit/build_conduit.sh
@@ -585,6 +585,10 @@ fi
 if [ ! -d ${raja_src_dir} ]; then
   echo "**** Extracting ${raja_tarball}"
   tar ${tar_extra_args} -xzf ${raja_tarball} -C ${source_dir}
+  # fix ambiguous unqualified partition call
+  cd ${raja_src_dir}
+  patch -p1 < ${script_dir}/2026_06_15_raja_sort_partition_ambiguity_fix.patch
+  cd ${root_dir}
 fi
 
 raja_extra_cmake_args=""
--- a/scripts/build_conduit/build_conduit.sh
+++ b/scripts/build_conduit/build_conduit.sh
@@ -863,6 +863,7 @@ echo 'set(CMAKE_INSTALL_PREFIX ' ${conduit_install_dir} ' CACHE PATH "")' >> ${c
 echo 'set(ENABLE_TESTS ' ${enable_tests} ' CACHE BOOL "")' >> ${cmake_host_config}
 echo 'set(ENABLE_MPI ' ${enable_mpi} ' CACHE BOOL "")' >> ${cmake_host_config}
 echo 'set(ENABLE_FIND_MPI ' ${enable_find_mpi} ' CACHE BOOL "")' >> ${cmake_host_config}
+echo 'set(ENABLE_OPENMP ' ${enable_openmp} ' CACHE BOOL "")' >> ${cmake_host_config}
 echo 'set(ENABLE_FORTRAN ' ${enable_fortran} ' CACHE BOOL "")' >> ${cmake_host_config}
 echo 'set(ENABLE_PYTHON ' ${enable_python} ' CACHE BOOL "")' >> ${cmake_host_config}
 if ${build_pyvenv}; then
--- a/scripts/build_conduit/build_conduit.sh
+++ b/scripts/build_conduit/build_conduit.sh
@@ -585,7 +585,8 @@ fi
 if [ ! -d ${raja_src_dir} ]; then
   echo "**** Extracting ${raja_tarball}"
   tar ${tar_extra_args} -xzf ${raja_tarball} -C ${source_dir}
-  # fix ambiguous unqualified partition call
+  # Fix ambiguous unqualified partition call
+  # TODO: Remove this patch after upgrading to RAJA v2026.xx.x
   cd ${raja_src_dir}
   patch -p1 < ${script_dir}/2026_06_15_raja_sort_partition_ambiguity_fix.patch
   cd ${root_dir}
--- a/src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
@@ -938,6 +938,7 @@ private:
                 // encode element and face into element_face.
                 uint64 element_face = facestart + face;
 
+                // TODO: Can this use conduit::execution::sort_ascending?
                 std::sort(face_pts_start, face_pts_end);
                 uint64 faceid = conduit::utils::hash(face_pts_start,
                                     static_cast<unsigned int>(points_per_face));
@@ -956,6 +957,8 @@ private:
         // general faceid, then by their elemface "ef", which should keep the
         // elements in order.
         CONDUIT_ANNOTATE_MARK_BEGIN("Sort labels");
+        // std::vectors of std::pairs are sorted by their first element, so we
+        // don't need a custom comparator.
         conduit::execution::sort_ascending(policy, faceid_to_ef.begin(), faceid_to_ef.end());
         CONDUIT_ANNOTATE_MARK_END("Sort labels");
 #ifdef DEBUG_PRINT
@@ -975,6 +978,8 @@ private:
 #endif
 
         // Sort on ef to get back to a ef->unique mapping.
+        // std::vectors of std::pairs are sorted by their first element, so we
+        // don't need a custom comparator.
         conduit::execution::sort_ascending(policy, ef_to_unique.begin(), ef_to_unique.end());
         CONDUIT_ANNOTATE_MARK_END("Sort ef->unique");
 #ifdef DEBUG_PRINT
@@ -1161,6 +1166,8 @@ private:
 
         // Sort edgeid_to_ee so any like edges will be sorted.
         CONDUIT_ANNOTATE_MARK_BEGIN("Sort labels");
+        // std::vectors of std::pairs are sorted by their first element, so we
+        // don't need a custom comparator.
         conduit::execution::sort_ascending(policy, edgeid_to_ee.begin(), edgeid_to_ee.end());
         CONDUIT_ANNOTATE_MARK_END("Sort labels");
 #ifdef DEBUG_PRINT
@@ -1177,6 +1184,8 @@ private:
 #endif
 
         // Sort on ef to get back to a ef->unique mapping.
+        // std::vectors of std::pairs are sorted by their first element, so we
+        // don't need a custom comparator.
         conduit::execution::sort_ascending(policy, ee_to_unique.begin(), ee_to_unique.end());
         CONDUIT_ANNOTATE_MARK_END("Sort ef->unique");
 #ifdef DEBUG_PRINT
@@ -2176,8 +2185,9 @@ TopologyMetadata::Implementation::build_edge_key_to_id(
 
     // Sort the edges by the ids.
     conduit::execution::ExecutionPolicy sort_policy = conduit::execution::ExecutionPolicy::host();
-    conduit::execution::sort_ascending(
-        sort_policy, edge_key_to_id.begin(), edge_key_to_id.end());
+    // std::vectors of std::pairs are sorted by their first element, so we
+    // don't need a custom comparator.
+    conduit::execution::sort_ascending(sort_policy, edge_key_to_id.begin(), edge_key_to_id.end());
 }
 
 //---------------------------------------------------------------------------
@@ -2466,6 +2476,8 @@ TopologyMetadata::Implementation::build_child_to_parent_association(int e, int a
     std::cout << "p2c=" << p2c << std::endl;
 #endif
     // Sort p2c by child.
+    // TODO: see if p2c can be constucted in reverse to make it compatible
+    // with conduit::execution::sort_ascending
     std::sort(p2c.begin(), p2c.end(),
         [&](const std::pair<index_t, index_t> &lhs,
             const std::pair<index_t, index_t> &rhs)
@@ -2491,6 +2503,7 @@ TopologyMetadata::Implementation::build_child_to_parent_association(int e, int a
         {
             index_t *start = &mapCP.data[mapCP.offsets[i]];
             index_t *end = start + mapCP.sizes[i];
+            // TODO: Can this use conduit::execution::sort_ascending?
             std::sort(start, end);
         }
     }
@@ -3721,6 +3734,7 @@ TopologyMetadata::TopologyMetadata(const conduit::Node &topology,
                     }
                 }
             }
+            // TODO: Can this use conduit::execution::sort_ascending?
             std::sort(vert_ids.begin(), vert_ids.end());
 
             // Look up the entity in the map, make global_id.
--- a/src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
@@ -2466,7 +2466,6 @@ TopologyMetadata::Implementation::build_child_to_parent_association(int e, int a
     std::cout << "p2c=" << p2c << std::endl;
 #endif
     // Sort p2c by child.
-    conduit::execution::ExecutionPolicy policy = conduit::execution::ExecutionPolicy::host();
     std::sort(p2c.begin(), p2c.end(),
         [&](const std::pair<index_t, index_t> &lhs,
             const std::pair<index_t, index_t> &rhs)
--- a/src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
@@ -956,7 +956,7 @@ private:
         // general faceid, then by their elemface "ef", which should keep the
         // elements in order.
         CONDUIT_ANNOTATE_MARK_BEGIN("Sort labels");
-        conduit::execution::sort(policy, faceid_to_ef.begin(), faceid_to_ef.end());
+        conduit::execution::sort_ascending(policy, faceid_to_ef.begin(), faceid_to_ef.end());
         CONDUIT_ANNOTATE_MARK_END("Sort labels");
 #ifdef DEBUG_PRINT
         std::cout << "faceid_to_ef.sorted = " << faceid_to_ef << std::endl;
@@ -975,13 +975,7 @@ private:
 #endif
 
         // Sort on ef to get back to a ef->unique mapping.
-        conduit::execution::sort(
-            policy, ef_to_unique.begin(), ef_to_unique.end(),
-            [&](const std::pair<uint64, uint64> &lhs, const std::pair<uint64, uint64> &rhs)
-            {
-                // Only sort using the ef value.
-                return lhs.first < rhs.first;
-            });
+        conduit::execution::sort_ascending(policy, ef_to_unique.begin(), ef_to_unique.end());
         CONDUIT_ANNOTATE_MARK_END("Sort ef->unique");
 #ifdef DEBUG_PRINT
         std::cout << "ef_to_unique.sorted = " << ef_to_unique << std::endl;
@@ -1167,7 +1161,7 @@ private:
 
         // Sort edgeid_to_ee so any like edges will be sorted.
         CONDUIT_ANNOTATE_MARK_BEGIN("Sort labels");
-        conduit::execution::sort(policy, edgeid_to_ee.begin(), edgeid_to_ee.end());
+        conduit::execution::sort_ascending(policy, edgeid_to_ee.begin(), edgeid_to_ee.end());
         CONDUIT_ANNOTATE_MARK_END("Sort labels");
 #ifdef DEBUG_PRINT
         std::cout << "edgeid_to_ee.sorted = " << edgeid_to_ee << std::endl;
@@ -1183,13 +1177,7 @@ private:
 #endif
 
         // Sort on ef to get back to a ef->unique mapping.
-        conduit::execution::sort(
-            policy, ee_to_unique.begin(), ee_to_unique.end(),
-            [&](const std::pair<uint64, uint64> &lhs, const std::pair<uint64, uint64> &rhs)
-            {
-                // Only sort using the ee value.
-                return lhs.first < rhs.first;
-            });
+        conduit::execution::sort_ascending(policy, ee_to_unique.begin(), ee_to_unique.end());
         CONDUIT_ANNOTATE_MARK_END("Sort ef->unique");
 #ifdef DEBUG_PRINT
         std::cout << "ee_to_unique.sorted = " << ee_to_unique << std::endl;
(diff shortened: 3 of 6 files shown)


src/libs/conduit/conduit_execution.hpp
```diff
@@ -101,14 +101,59 @@ forall_exec(ExecPolicyTag,
...
+template <typename Iterator, typename Predicate>
+inline void
+sort_exec(OpenMPExec,
+          Iterator begin,
+          Iterator end,
+          Predicate &&predicate) noexcept
+{
+    std::sort(begin, end, std::forward<Predicate>(predicate));
+}
+#endif
```
JustinPrivitera: what is preventing us from using RAJA openMP sort? I think RAJA can do this for us

scripts/build_conduit/build_conduit.sh
```diff
@@ -863,6 +863,7 @@ echo 'set(CMAKE_INSTALL_PREFIX ' ${conduit_install_dir} ' CACHE PATH "")' >> ${c
 echo 'set(ENABLE_TESTS ' ${enable_tests} ' CACHE BOOL "")' >> ${cmake_host_config}
 echo 'set(ENABLE_MPI ' ${enable_mpi} ' CACHE BOOL "")' >> ${cmake_host_config}
 echo 'set(ENABLE_FIND_MPI ' ${enable_find_mpi} ' CACHE BOOL "")' >> ${cmake_host_config}
+echo 'set(ENABLE_OPENMP ' ${enable_openmp} ' CACHE BOOL "")' >> ${cmake_host_config}
```
JustinPrivitera, line 871: why do we need this?

src/libs/conduit/conduit_execution.hpp
```diff
@@ -89,6 +90,61 @@ index_t get_device_allocator_id();
 index_t get_host_allocator_id();
 
 
+// Defined here so that both the RAJA and non-RAJA codepaths can use this.
```
JustinPrivitera: shouldn't the RAJA+openMP case use the RAJA version, not a bespoke implementation?

Hint here: 
Image

https://raja.readthedocs.io/en/develop/sphinx/user_guide/tutorial/sort.html#sort-pairs

JustinPrivitera: or perhaps the comment is out of date?

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -23,7 +23,8 @@
 using namespace conduit;
 using conduit::execution::ExecutionPolicy;
 
-index_t EXECUTION_TEST_ARRAY_SIZE = 4;
+// 2^20 => 1,048,576 elements
+index_t EXECUTION_TEST_ARRAY_SIZE = 1 << 20;
```
JustinPrivitera: we want this number to be small so that the test runs quickly; we can adjust this value using command line argument. Running `./t_conduit_execution <big number>` will make this value larger.

src/libs/conduit/conduit_execution.hpp
siramok: Since this change is now outdated, I'll reply to the newer comment that asks essentially the same thing.

scripts/build_conduit/build_conduit.sh
siramok: This line seems to be necessary to get the tests to recognize openmp as a valid policy type, otherwise the serial and device policies are the only things that execute. This change could also go in a separate "fixes" PR if you prefer.

src/libs/conduit/conduit_execution.hpp
siramok: You're correct, theoretically both the serial and openmp policies should be able to use RAJA's sort. In practice there seems to be a RAJA util function named `partition` that conflicts with the `std::partition` that `std::vector` brings into scope. This results in a compilation error due to ambiguity over which `partition` definition is the right one to use. Therefore the PR currently just uses the same bespoke quicksort for both the RAJA and no-RAJA cases.

I'll need to do a bit of extra work to 1) confirm that the issue is on RAJA's end and 2) report it to them, but once the naming conflict is resolved, we should be able to migrate our CPU cases to using RAJA.

src/tests/conduit/t_conduit_execution.cpp
siramok: Makes sense, reverting.

siramok: Fixed.

scripts/build_conduit/build_conduit.sh
JustinPrivitera: this is fine here, thanks for catching this

src/libs/conduit/conduit_execution.hpp
```diff
@@ -89,6 +90,66 @@ index_t get_device_allocator_id();
...
 
+// Defined in this scope so that both the RAJA and non-RAJA paths can use it.
+#if defined(CONDUIT_USE_OPENMP)
+//-----------------------------------------------------------------------------
+// Parallel quicksort using OpenMP: tasks are spawned for
+// at most ceil(log2(num_threads)) levels of recursion.
+//
+// TODO: Needs benchmarking. This is surely only better than std::sort beyond
+// a certain size/thread threshold. A solution is to add a heuristic that does
+// a plain std::sort for small inputs. That would need benchmarking as well.
```
JustinPrivitera: do we want to fall back for small arrays to std::sort?

siramok: I guess it depends on the average array size we expect to be sorting. If we know it will generally be large arrays, then perhaps there's no need to bother.

src/libs/conduit/conduit_execution.hpp
```diff
@@ -116,14 +177,93 @@ forall_exec(ExecPolicyTag,
...
+    }
+}
+#endif // defined(CONDUIT_USE_OPENMP)
+
+//-----------------------------------------------------------------------------
+// Maps std::less/std::greater to their RAJA equivalents so that RAJA can
+// dispatch to more performant sort implementations. All other predicates
+// pass through unchanged, falling back to the slower general-purpose sort.
+template <typename ValueType, typename Predicate>
+auto normalize_predicate(Predicate &&p)
```
siramok: I'll preemptively give you an explanation about this (because I think it's neat):

According to the RAJA docs, the CUDA/HIP back-ends can leverage fast radix sort implementations if a few conditions are met. One of those conditions is to use RAJA's built-in less-than and greater-than operators. This implies that using anything else (e.g. custom predicates) will force a slower implementation to be used. In practice (if I understand the RAJA source correctly), passing anything other than a built-in RAJA operator will actually just fail outright for CUDA/HIP.

Through that lens, our default case is already optimized because when you don't supply a predicate, RAJA will automatically use its built-in less-than operator. We can't necessarily control that in the custom predicate case. However, if someone were to pass in `std::less` or `std::greater` as their custom predicate (seems reasonable, given that they're part of C++ itself), the purpose of this function is to intercept that and map it to one of the built-in RAJA operators instead.

If the predicate isn't trivially mappable, it just gets passed along as-is for RAJA to do with it as it pleases (like crash). Maybe there's a way to emit a warning that the user may be getting a sub-optimal sort in this case?

src/libs/conduit/conduit_execution.hpp
siramok: I've verified that the fix is super simple on the RAJA side, and it fixes both the serial and openmp sorts.

siramok: I added a .patch in the meantime and fixed the code to not use the bespoke implementation accordingly.

src/libs/conduit/conduit_execution.hpp
siramok: FWIW RAJA seems to switch to insertion sort if the input is below a size threshold.

scripts/build_conduit/2026_06_15_raja_sort_partition_ambiguity_fix.patch
```diff
@@ -0,0 +1,11 @@
+--- a/include/RAJA/util/sort.hpp
++++ b/include/RAJA/util/sort.hpp
+@@ -389,7 +389,7 @@
+     }
+ 
+     // partition
+-    mid = partition(begin, last, [&](Iter it) {
++    mid = detail::partition(begin, last, [&](Iter it) {
```
siramok, line 8: Fixes the RAJA compilation error (we can decide if this still matters after looking at the perf results).

siramok:
Testing with array sizes of 99,999,999 seems to show that RAJA is slower for the CPU cases (on my local workstation, but I'd be curious to see if RAJA does other optimizations at scale):

```
(RAJA - serial)
Path   Min time/rank Max time/rank Avg time/rank
sort        2.128020      2.128020      2.128020

(no RAJA - serial)
Path   Min time/rank Max time/rank Avg time/rank
sort        0.488480      0.488480      0.488480

------------------------------------------------------------------------------
(RAJA - openmp)
Path   Min time/rank Max time/rank Avg time/rank
sort        0.499935      0.499935      0.499935

(no RAJA - openmp)
Path   Min time/rank Max time/rank Avg time/rank
sort        0.185750      0.185750      0.185750
```

src/libs/conduit/conduit_execution.hpp
siramok: A possible alternative to this could be to provide our own `conduit::operators::less` and `conduit::operators::greater` predicates that map to the RAJA equivalents behind the scenes. However, that doesn't solve the problem that users may inadvertently ignore them and use `std::less` and `std::greater` anyways.

src/libs/conduit/conduit_execution.hpp
JustinPrivitera: It wouldn't hurt to have our own size threshold then. We can tweak down the line if experience tells us is too big or too small.

scripts/build_conduit/2026_06_15_raja_sort_partition_ambiguity_fix.patch
JustinPrivitera: this seems like a useful fix to have. WDYM the perf results?

src/libs/conduit/conduit_execution.hpp
JustinPrivitera: This is very interesting. Let's consult @cyrush and see what he thinks. My main concern is with code readability, since this style is unfamiliar to many. I think good comments for each if constexpr block would go a long way, and a comment at the call site for this function would help clear things up as well.

src/libs/conduit/conduit_execution.hpp
```diff
@@ -202,81 +231,145 @@ using ReduceMaxLoc = RAJA::ReduceMaxLoc<typename ExecPolicyTag::reduce_policy, T
...
-            Kernel&& kernel) noexcept
+// Parallel quicksort using OpenMP: tasks are spawned for
+// at most ceil(log2(num_threads)) levels of recursion.
+//
+// TODO: Needs benchmarking. This is surely only better than std::sort beyond
+// a certain size/thread threshold. A solution is to add a heuristic that does
+// a plain std::sort for small inputs. That would need benchmarking as well.
+template <typename Iterator, typename Compare>
+void
+omp_quicksort(Iterator begin,
```
JustinPrivitera: Like I said elsewhere I think it would be best to have a threshold below which we prefer std::sort. We don't necessarily have to figure out the right value for that threshold right away.

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -626,6 +626,10 @@ TEST(conduit_execution, strawman_data_accessor)
 
     // our main loop iterates over these items
     // comment out specific entries to test a specific combination
+    // TODO: This crashes if RAJA is built without CUDA or HIP because the
+    // Umpire "DEVICE" allocator is unavailable when "device" src/des locations
+    // are used. The "device" entries should be skipped when device memory is
+    // not available.
     const std::vector<std::string> src_locations = {"host", "device"};
```
JustinPrivitera, line 686: Looks like a bug I've introduced. We need to figure this out, but probably not in this PR. You can leave the TODO there and we will get to it later.

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -1496,6 +1504,117 @@ TEST(conduit_execution, strawman_data_array)
...
+        // Fill the array with ascending numbers
+        conduit::execution::forall(policy, 0, size, [=] CONDUIT_EXEC(index_t i)
+        {
+            vals_ptr[i] = i;
+        });
+        CONDUIT_DEVICE_ERROR_CHECK(policy);
+
+        // Using std::greater<index_t> will automatically map to RAJA::operators::greater<index_t>
+        conduit::execution::sort(policy, vals_ptr, vals_ptr + size, std::greater<index_t>{});
+        CONDUIT_DEVICE_ERROR_CHECK(policy);
```
JustinPrivitera: That is pretty nifty

JustinPrivitera (commented):
I left some comments; in general this looks very good. Looking forward to chatting more and maybe getting some input from Cyrus.

It looks like some things shifted around in `src/libs/conduit/conduit_execution.hpp`, but I am guessing that things didn't change for forall and that we are only using our bespoke openMP sort when RAJA is off?

scripts/build_conduit/2026_06_15_raja_sort_partition_ambiguity_fix.patch
JustinPrivitera: are you saying why bother fixing/using RAJA if it is slower?

JustinPrivitera:
> Testing with array sizes of 99,999,999 seems to show that RAJA is slower for the CPU cases (on my local workstation, but I'd be curious to see if RAJA does other optimizations at scale):
> 
> ```
> (RAJA - serial)
> Path   Min time/rank Max time/rank Avg time/rank
> sort        2.128020      2.128020      2.128020
> 
> (no RAJA - serial)
> Path   Min time/rank Max time/rank Avg time/rank
> sort        0.488480      0.488480      0.488480
> 
> ------------------------------------------------------------------------------
> (RAJA - openmp)
> Path   Min time/rank Max time/rank Avg time/rank
> sort        0.499935      0.499935      0.499935
> 
> (no RAJA - openmp)
> Path   Min time/rank Max time/rank Avg time/rank
> sort        0.185750      0.185750      0.185750
> ```

I am a little confused by this - why the disparity in runtimes between serial and serial w/ RAJA? Serial should never use the RAJA path.

src/libs/conduit/conduit_execution.hpp
cyrush: Thinking of this at a high level, I think users will want sort ascending and sort descending bottled up - they won't need much flexibility for their own custom comparison function.  And if they do, they become the expert and should evaluate their choices. 


That said - I still think we need something like this implement that simpler interface.

siramok: Three more options could be:
1. Have the API be `sort_exec(OpenMPExec, Iterator begin, Iterator end, bool reverse=false)` which still does LT by default. Then passing true for reverse gives you GT.
2. Have an explicit `sort_exec_lt` and `sort_exec_gt`.
3. Keep what's in the PR now, remove the predicate intercept code, and rely on docs/examples to explain the situation.

scripts/build_conduit/2026_06_15_raja_sort_partition_ambiguity_fix.patch
siramok: I think the fix is good in any case, but yes I wondered if we would still want to use RAJA for the serial and openmp cases if it ends up being slower. In which case, this patch isn't necessary to have working CUDA/HIP sort.

src/libs/conduit/conduit_execution.hpp
siramok: Sounds good, should be fairly easy to implement.

siramok:
> > Testing with array sizes of 99,999,999 seems to show that RAJA is slower for the CPU cases (on my local workstation, but I'd be curious to see if RAJA does other optimizations at scale):
> > ```
> > (RAJA - serial)
> > Path   Min time/rank Max time/rank Avg time/rank
> > sort        2.128020      2.128020      2.128020
> > 
> > (no RAJA - serial)
> > Path   Min time/rank Max time/rank Avg time/rank
> > sort        0.488480      0.488480      0.488480
> > 
> > ------------------------------------------------------------------------------
> > (RAJA - openmp)
> > Path   Min time/rank Max time/rank Avg time/rank
> > sort        0.499935      0.499935      0.499935
> > 
> > (no RAJA - openmp)
> > Path   Min time/rank Max time/rank Avg time/rank
> > sort        0.185750      0.185750      0.185750
> > ```
> 
> I am a little confused by this - why the disparity in runtimes between serial and serial w/ RAJA? Serial should never use the RAJA path.

Ah, serial can go back to using `std::sort`. The patch that fixes RAJA also fixed it for the serial case, so I threw it in for no other reason than it worked.

siramok:
> I left some comments; in general this looks very good. Looking forward to chatting more and maybe getting some input from Cyrus.
> 
> It looks like some things shifted around in `src/libs/conduit/conduit_execution.hpp`, but I am guessing that things didn't change for forall and that we are only using our bespoke openMP sort when RAJA is off?

In regard to things shifting, I moved the openmp forall and sort to be near each other so that they could go in the same `#if defined(CONDUIT_USE_OPENMP)` block. I also moved the bespoke quicksort to be there as well, since it no longer needs to be used by both the RAJA and no-RAJA paths.

scripts/build_conduit/2026_06_15_raja_sort_partition_ambiguity_fix.patch
cyrush: @siramok  can you check with the RAJA folks -- see if they have already made a change like this, and if not maybe we can help get it in their next release (which is soon)

src/libs/conduit/conduit_execution.hpp
JustinPrivitera: FWIW @cyrush Brad had used sort already before and was using it with a custom comparator. But he is an expert and probably evaluated his choices.

scripts/build_conduit/2026_06_15_raja_sort_partition_ambiguity_fix.patch
siramok: This is still an issue in their develop branch, but I can let them know about it. Since it's a simple change, do you think would they'd be okay if I made a PR or is it better that I open an issue?

src/libs/conduit/conduit_execution.hpp
siramok: This has been addressed.

src/libs/conduit/conduit_execution.hpp
siramok: This has been addressed.

scripts/build_conduit/2026_06_15_raja_sort_partition_ambiguity_fix.patch
rhornung67: Please submit a RAJA issue or a PR if you wish. Thank you.

siramok:
- The serial case uses `std::sort` again, not RAJA.
- There's now a size threshold for the no-RAJA openmp case, below which `std::sort` gets used.
- ~Added more comments around `normalize_predicate` (I'm okay if we ultimately remove it as well).~ Removed in favor of getting rid of custom comparators.

Question:
- ~Do we still want openmp to use RAJA if its available, despite seeming to be slower? We could also add a TODO to come back later to do more rigorous testing.~ We do want to use RAJA + openmp if both of those are built together.

scripts/build_conduit/2026_06_15_raja_sort_partition_ambiguity_fix.patch
siramok: I opened a PR [here](https://github.com/llnl/RAJA/pull/2042).

src/libs/conduit/conduit_execution.hpp
JustinPrivitera: we will have `sort_ascending()` (afterwards in ascending order) and `sort_descending()` (afterwards in descending order) and then we don't need to do fancy things with predicates.

siramok: We decided to remove support for custom comparators, resolved.

scripts/build_conduit/2026_06_15_raja_sort_partition_ambiguity_fix.patch
siramok: The RAJA maintainers have accepted the fix, but we'll keep the patch in the meantime. A TODO was added to remove the patch when we eventually move to RAJA v2026.xx.x. Resolved.

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
```diff
@@ -975,13 +975,7 @@ class TopologyMetadata::Implementation : public TopologyMetadataBase
 #endif
 
         // Sort on ef to get back to a ef->unique mapping.
-        conduit::execution::sort(
-            policy, ef_to_unique.begin(), ef_to_unique.end(),
-            [&](const std::pair<uint64, uint64> &lhs, const std::pair<uint64, uint64> &rhs)
-            {
-                // Only sort using the ef value.
-                return lhs.first < rhs.first;
-            });
```
JustinPrivitera, line 984: How does this work? The custom comparator was trivial? Do all the tests pass?

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
```diff
@@ -1183,13 +1177,7 @@ class TopologyMetadata::Implementation : public TopologyMetadataBase
...
 
         // Sort on ef to get back to a ef->unique mapping.
-        conduit::execution::sort(
-            policy, ee_to_unique.begin(), ee_to_unique.end(),
-            [&](const std::pair<uint64, uint64> &lhs, const std::pair<uint64, uint64> &rhs)
-            {
-                // Only sort using the ee value.
-                return lhs.first < rhs.first;
-            });
+        conduit::execution::sort_ascending(policy, ee_to_unique.begin(), ee_to_unique.end());
```
JustinPrivitera, line 1189: same question with this one

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
```diff
@@ -2188,13 +2176,8 @@ TopologyMetadata::Implementation::build_edge_key_to_id(
...
     conduit::execution::ExecutionPolicy sort_policy = conduit::execution::ExecutionPolicy::host();
-    conduit::execution::sort(
-        sort_policy, edge_key_to_id.begin(), edge_key_to_id.end(),
-        [&](const std::pair<uint64, index_t> &lhs,
-            const std::pair<uint64, index_t> &rhs) 
-        {
-            return lhs.first < rhs.first;
-        });
+    conduit::execution::sort_ascending(
+        sort_policy, edge_key_to_id.begin(), edge_key_to_id.end());
```
JustinPrivitera: and this one

JustinPrivitera (commented):
Just those last couple questions

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
siramok: All tests do pass (at least the `t_blueprint_mesh_topology_metadata` tests do), without openmp+raja, with openmp but no raja, and with openmp+raja. :)

To answer your question, it reduces to the trivial case in this example (and the next 2) because the custom comparator was operating on the `.first` item of the pair which is coincidentally how comparison is already defined on `std::pair`. The one that I couldn't trivially convert was comparing on the `.second` item in the pair.

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
siramok: See above answer.

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
siramok: See above answer.

siramok:
All the trailing whitespace has been re-added to cleanup the PR. Sorry about that!

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
JustinPrivitera: gotcha, thanks for tracking down those details!

JustinPrivitera (approved):
Thanks for these changes! Looks great.

siramok:
Just rebased on the latest commit, no new changes.

PR #1624, opened 2026-06-16
Device Support: Tweaks to Global Execution Settings

JustinPrivitera:
The changes we discussed about adding fallback location and changing names from earlier today

### DIFF
--- a/src/libs/conduit/conduit_execution.cpp
+++ b/src/libs/conduit/conduit_execution.cpp
@@ -317,14 +317,17 @@ ExecutionPolicy::policy_id_to_name(const PolicyID policy_id)
 class ExecutionOptions
 {
 public:
-    // chosen execution policy
-    static std::string execution_policy;
+    // chosen execution location
+    static std::string execution_location;
 
-    // allocator to use
-    static std::string output_allocator;
+    // chosen output location
+    static std::string output_location;
 
     // "sync" or "assume"
     static std::string sync_strategy;
+
+    // "host" or "device"
+    static std::string fallback_location;
     
     // allocator ids that are available
     static index_t device_allocator;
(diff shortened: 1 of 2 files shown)


src/libs/conduit/conduit_execution.cpp
```diff
@@ -334,66 +337,71 @@ class ExecutionOptions
 public:
     //------------------------------------------------------------------------
     // opts node:
-    //   execution_policy: "host"|"device"|"input_location"
-    //     # choose host, device, or input location (use `use_with` to get a 
+    //   execution_location: "host"|"device"|"input"
```
siramok, line 340: I like "input" better.

src/libs/conduit/conduit_execution.cpp
```diff
@@ -578,6 +631,10 @@ index_t ExecutionOptions::host_allocator =
             return HostMemory::allocate(num_items, item_size);
         },
         HostMemory::deallocate);
+#else
+index_t ExecutionOptions::device_allocator = -1;
+index_t ExecutionOptions::host_allocator = 0;
```
siramok, line 636: I notice that `device_allocator` and `user_provided_allocator` are both initialized to -1, should `host_allocator` be also?

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -93,9 +93,14 @@ TEST(conduit_execution, execution_settings)
     // container for opts whenever we fetch them
     Node get_opts;
 
+#if defined(CONDUIT_USE_DEVICE)
     // if new allocators are added earlier, then these values could change
     const index_t DEVICE_ALLOC_ID = 1;
     const index_t HOST_ALLOC_ID = 2;
+#else
+    const index_t DEVICE_ALLOC_ID = -1;
+    const index_t HOST_ALLOC_ID = 0;
```
siramok, line 102: Related to the above question, just curious what these numbers mean.

src/libs/conduit/conduit_execution.cpp
JustinPrivitera: My understanding is that 0 is the default allocator for conduit, so setting the allocator here to 0 for the case when we do not have device should be legal. Having a negative number I assume is illegal. But maybe neither should be allowed in the case where we have no Umpire. Not sure.

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -257,31 +286,55 @@ TEST(conduit_execution, execution_settings)
...
+        EXPECT_TRUE(get_opts.has_child("fallback_location"));
+        EXPECT_EQ(get_opts["fallback_location"].as_string(), "device");
         execution::reset_execution_options();
     }
 
     // test user_provided output allocator
     {
         Node exec_opts;
-        exec_opts["output_allocator"] = 12345;
+        exec_opts["output_location"] = 12345;
```
siramok, line 337: Is this meant to be a fail case or is 12345 considered valid?

src/tests/conduit/t_conduit_execution.cpp
JustinPrivitera: Conduit has a list of allocator ids that can be used to tell nodes how to allocate memory. We can add new allocators so that nodes can automatically be allocated on the host or device. The ids are static so we can test for them like this.

siramok:
All looks good, just had some questions for my own curiosity.

src/tests/conduit/t_conduit_execution.cpp
JustinPrivitera: this is valid in theory, although to get here we would have had to have registered 12344 allocators before this. Unlikely, but there is nothing illegal about the number.

siramok (approved):

PR #1625, opened 2026-06-18
Device Support: Remove dispatch and simplify reducer/atomic API

JustinPrivitera:
developed using codex

### DIFF
--- a/src/libs/conduit/conduit_execution.hpp
+++ b/src/libs/conduit/conduit_execution.hpp
@@ -117,6 +117,28 @@ using DefaultReducePolicy = RAJA::seq_reduce;
 #endif
 
 //-----------------------------------------------------------------------------
+// Atomics also follow translation-unit capability instead of the runtime loop
+// policy. GPU-capable translation units use device atomics in device compile
+// paths, with host fallback selected once here for all runtime policies.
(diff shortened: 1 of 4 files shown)


cyrush:
The simplified templates look good -- Note: think we want to keep dispatch with the functor as a just in case. 
We know it works, even if the double dispatch wont.

JustinPrivitera:
> The simplified templates look good -- Note: think we want to keep dispatch with the functor as a just in case. We know it works, even if the double dispatch wont.

I'll keep polishing this... but can't we remove dispatch if we discover we don't need it? I think it complicates the API. We can always add back in later if we decide we need it again.

siramok (approved):
LGTM. This new approach both makes sense and makes the code much nicer to reason about, I am a fan.

siramok:
I verified that this PR does not fix the compilation errors with respect to CUDA and GTest (on both matrix and my personal PC). However, using this branch as a base, I was able to fix the remaining errors without too much trouble and verify that the tests do pass for the CUDA back-end. :)

I think the simplest path forward would be to merge this PR as-is, then I could open a separate PR with the CUDA fixes? If you'd rather handle it another way that's fine with me too.

cyrush (approved):
Looks great, nice we are able to simplify this!

PR #1626, opened 2026-06-18
Device Support: Performance Tests for Uniform to Rect Coordset Transform

JustinPrivitera:


src/tests/blueprint/t_blueprint_mesh_execution.cpp
```diff
@@ -0,0 +1,62 @@
...
+    Node mesh;
+    blueprint::mesh::examples::braid("uniform", 100, 100, 100, mesh);
+    const Node &uniform_coords = mesh["coordsets"]["coords"];
+
+    // create baseline result for correctness checking
+    Node baseline;
+    blueprint::mesh::examples::braid("rectilinear", 100, 100, 100, baseline);
+    
+    Node exec_opts;
+    // TODO for each choice in the matrix of choices...
```
JustinPrivitera, line 34: We may have a smaller matrix of choices for this transform, since there is no input data and the output is created during the transform instead of already living somewhere. We realistically need to test host & device for exec space against host & device for output location against sync & assume for sync strategy. Other transforms will require more thorough testing.

PR #1629, opened 2026-06-22
Various execution test fixes for different compile option combinations

siramok:
The tests were failing with various accelerator options enabled/disabled at compile time, so I took a pass at fixing all the issues I found.

This PR has been tested (and is verified passing) for the following combinations of compile options:
- no openmp, no RAJA (implicit no device)
- openmp, no RAJA (implicit no device)
- openmp, RAJA (with no device)
- openmp, RAJA (with device)


src/libs/conduit/conduit_execution.hpp
```diff
@@ -113,7 +113,9 @@ namespace detail
 using DefaultReducePolicy = RAJA::cuda_reduce;
 #elif defined(CONDUIT_TU_IS_HIP)
 using DefaultReducePolicy = RAJA::hip_reduce;
-#else
+#elif defined(CONDUIT_USE_OPENMP)
+using DefaultReducePolicy = RAJA::omp_reduce;
```
siramok, line 117: This change was needed to avoid data races (causing test failures) in the openmp + RAJA (but no device) case.

src/libs/conduit/conduit_execution.hpp
```diff
@@ -430,8 +432,17 @@ template <typename T>
 CONDUIT_EXEC T
 atomic_add_exec(T *acc, T value)
 {
-    T res = *acc;
-    *acc += value;
+    T res;
+#if defined(CONDUIT_USE_OPENMP)
+    // atomic capture does both the read and write atomically
+    #pragma omp atomic capture
```
siramok, line 438: This change (and subsequent atomic/reducer changes) were needed to avoid data races (causing test failures) in the openmp without RAJA case.

src/libs/conduit/conduit_execution.hpp
```diff
@@ -534,6 +573,11 @@ class ReduceMin
     //-----------------------------------------------------------------------------
     CONDUIT_EXEC void min(const T value) const
     {
+#if defined(CONDUIT_USE_OPENMP)
+        // A conditional update can't use omp atomic, so we use a critical section instead
+        // TODO: there may be a more efficient way to do this reduction
```
siramok, line 578: I added this (and subsequent) TODOs because critical sections are an inherent bottleneck, but there may be a more efficient (and likewise more complicated) way to perform these reductions with openmp.

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -114,19 +114,28 @@ TEST(conduit_execution, execution_settings)
     {
         const index_t host_alloc_id = execution::get_host_allocator_id();
         const index_t device_alloc_id = execution::get_device_allocator_id();
-        host_data.set_allocator(host_alloc_id);
-        device_data.set_allocator(device_alloc_id);
         const std::vector<float64> src_vals = make_execution_src_vals(EXECUTION_TEST_ARRAY_SIZE);
+
+        host_data.set_allocator(host_alloc_id);
```
siramok, line 118: These changes are to better guard against crashes/test failures in cases where RAJA is not built with device support. Previously this test would fail for these cases:

- No openmp, no RAJA
- openmp, no RAJA
- openmp, RAJA (without CUDA/HIP)

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -662,6 +934,94 @@ TEST(conduit_execution, test_atomics)
...
+        node["vals"].set(DataType::index_t(size));
+        {
+            index_t_array fill(node["vals"]);
+            for (index_t i = 0; i < size; i++)
+            {
+                fill[i] = init_vals[i];
+            }
+        }
+
+        // DataArray
```
siramok: The atomic functions require a pointer to the memory being operated on. We can exercise `DataArray` here because it returns a pointer to the underlying memory, while `DataAccessor` does not.

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -684,6 +1040,13 @@ TEST(conduit_execution, strawman_data_accessor)
     {
         for (const std::string &des_start : des_locations)
         {
+            // Skip device-memory combinations when no device is available.
+            if ((src_start == "device" || des_start == "device") &&
+                !ExecutionPolicy::is_device_enabled())
```
siramok, line 716: This (and the next similar instance) is another fix for tests failing when conduit is built without device support.

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -1042,6 +1405,13 @@ TEST(conduit_execution, strawman_data_array)
...
```
siramok, line 1078: This is another fix for tests failing when conduit is built without device support.

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -1442,6 +1812,60 @@ TEST(conduit_execution, test_sort_ascending)
...
+            // Fill with descending values so sort_ascending has work to do.
+            index_t_array fill(node["vals"]);
+            for (index_t i = 0; i < size; i++)
+            {
+                fill[i] = size - 1 - i;
+            }
+        }
+
+        // DataArray
+        index_t_array arr(node["vals"]);
```
siramok: Similar situation as the node-backed atomic test, sorting requires a pointer to memory we want to operate on, making it not compatible with `DataAccessor`.

siramok:
Per our conversation offline, the extra tests were removed since we already exercise the node-backed paths (`DataArray`/`DataAccessor`) in this file. The TODO asking about serial vs openmp was also removed per our discussion.

What remains in this PR are the fixes for tests failing due to 1) openmp data races and 2) assumptions about device availability.

src/tests/conduit/t_conduit_execution.cpp
JustinPrivitera: Down below (lines 727-730) we have another check for failure. We should combine these. Both ask for `!ExecutionPolicy::is_device_enabled())`

src/tests/conduit/t_conduit_execution.cpp
JustinPrivitera: same feedback here

JustinPrivitera (changes requested):
Looks good aside from minor tweak.

src/tests/conduit/t_conduit_execution.cpp
siramok: The other failure check occurs in a different test, both `strawman_data_array` (this instance) and `strawman_data_accessor` (instance you are referring to) do a double for loop over host/device to exercise the different combinations.

However, your point about combining them is well taken, although I expect doing so would require a slightly bigger refactor + re-review. E.g. the loop boilerplate could be extracted into a helper function and reused in both `strawman_data_array` and `strawman_data_accessor` ala the way `for_each_enabled_policy` does things.

For the sake of merging what you've already reviewed without first making substantial changes, I've added a `TODO` for that simplification above both instances.

siramok:
The duplicate instances of the failure check actually live within different tests. Despite that, It actually is possible to extract it into a helper to reduce the amount of duplicated boilerplate (ala the way `for_each_enabled_policy` works), but that would require a bit more refactoring than I think you were asking for.

Wanting to respect your verbal approval to merge once the feedback is addressed, but not wanting to change much without your review, I've added some `TODO`s as a compromise. I think it would be a good candidate for the cleanup pass we've discussed. :)

PR #1630, opened 2026-06-23
Test node-backed DataArray and DataAccessor construction

siramok:
This PR extends the existing `DataArray` and `DataAccessor` tests to exercise their respective node-backed constructors, i.e. various methods of constructing `DataArray` and `DataAccessor` instances directly from conduit `Node`s.

I didn't feel that there would be value in duplicating every test for the sake of creating node-backed versions. Furthermore, after duplicating some tests, I realized that there were some that could be combined back into the existing test. I tried to do that where possible. In cases where I did keep a duplicated test, I could look into creating some helper functions to reduce the boilerplate if you wish.

Despite my efforts, I still ended up in excess of 1k lines added. My apologies for presenting so much to review! If it helps, a vast majority is copy/pasted from existing tests.

### DIFF
--- a/src/tests/conduit/t_conduit_data_accessor.cpp
+++ b/src/tests/conduit/t_conduit_data_accessor.cpp
@@ -94,9 +94,9 @@ TEST(conduit_data_accessor, as_bitwidth_style)
 //-----------------------------------------------------------------------------
 TEST(conduit_data_accessor, summary_stats)
 {
-    std::vector<int64>   v_int64 ={-1,0,1};
-    std::vector<uint64>  v_uint64 = {1,2,3};
-    std::vector<float64> v_float64= {-1.0,0.0,1.0};
+    std::vector<int64>   v_int64   = {-1,0,1};
+    std::vector<uint64>  v_uint64  = {1,2,3};
+    std::vector<float64> v_float64 = {-1.0,0.0,1.0};
 
     // void* / DataType constructor variants
     {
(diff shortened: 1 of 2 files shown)


src/tests/conduit/t_conduit_data_accessor.cpp
```diff
@@ -98,33 +98,89 @@ TEST(conduit_data_accessor, summary_stats)
...
-    EXPECT_EQ(va_float64.sum(),0.0);
-    EXPECT_EQ(va_float64.count(0.0),1);
+        EXPECT_EQ(va_float64.min(),-1.0);
+        EXPECT_EQ(va_float64.max(),1.0);
+        EXPECT_EQ(va_float64.mean(),0.0);
+        EXPECT_EQ(va_float64.sum(),0.0);
+        EXPECT_EQ(va_float64.count(0.0),1);
+    }
 
+    // Node-backed constructor variants
```
JustinPrivitera, line 126: minor feedback but what do the variable names mean? In the original I assume the `va_type` vars meant "vector accessor type". Now there are no more vectors, so I wonder if the names should change? What does `nb_va_type` mean? Node-backed vector-accessor type?

src/tests/conduit/t_conduit_data_accessor.cpp
JustinPrivitera, line 95: summary stats was the only one it made sense to add tests to?

src/tests/conduit/t_conduit_data_array.cpp
```diff
@@ -489,13 +580,98 @@ TEST(conduit_data_array, set_single_element)
...
-    EXPECT_EQ(va_uint16[1],(int16)8);
-    EXPECT_EQ(va_uint32[1],(int32) 16);
-    EXPECT_EQ(va_uint64[1],(int64)32);
+    EXPECT_EQ(va_uint8[1],(uint8)4);
+    EXPECT_EQ(va_uint16[1],(uint16)8);
+    EXPECT_EQ(va_uint32[1],(uint32)16);
+    EXPECT_EQ(va_uint64[1],(uint64)32);
+
+    EXPECT_EQ(va_float32[1],(float32) 16.0);
+    EXPECT_EQ(va_float64[1],(float64) 32.0);
```
JustinPrivitera, line 589: fixing mistakes/missing cases?

src/tests/conduit/t_conduit_data_array.cpp
JustinPrivitera, line 523: is there no `set_single_element` test for data accessor? Maybe there should be plus `set_single_element_node_backed`

JustinPrivitera (changes requested):
Just a couple small comments. Otherwise looks great thanks for this careful work!

src/tests/conduit/t_conduit_data_accessor.cpp
siramok: Regarding `va_type`, that's essentially just copied from existing test variable names. I assumed it meant "value array" or "value accessor", although most (but not all) of the instances actually do still use a vector under the hood. There are a few spots that truly don't use a vector under the hood, but everything else follows the `va_type` pattern, so maybe it makes sense to just stick to convention?

Regarding `nb_va_type`, that's a result of getting AI assistance, and I don't think it's useful naming. So I simply changed all of those instances to be `va_type`. Thanks for catching that.

src/tests/conduit/t_conduit_data_accessor.cpp
siramok: There are a lot of usages of `n.value()` throughout `t_conduit_data_accessor`, but now that you mention it, that's not (I don't think) fully exercising the different constructor types. We can do more if you wish, I added another that I'll comment on in another conversation.

src/tests/conduit/t_conduit_data_array.cpp
siramok: There was a spot in the existing tests where we had `va_uintx`s on the left but on the right were casting values to `intx` (I guess this is it). Seemed like the compiler was ultimately doing the correct thing anyways, but I fixed those casts to be casting the right side value to/from the same type as the data array.

src/tests/conduit/t_conduit_data_array.cpp
siramok: There is, it's just called `set` in the data accessor tests. I went ahead and extended it to exercise all of the different node-based constructor types as well (effectively addressing this feedback).

siramok:
I believe that we discussed/resolved each of these comments offline. Please let me know if you want further changes. :)

src/tests/conduit/t_conduit_execution.cpp
JustinPrivitera, line 1: We need to rebase these changes away.

JustinPrivitera (changes requested):
Thanks for these changes. We need to rebase the changes to `src/tests/conduit/t_conduit_execution.cpp` away and then this is good to be merged.

src/tests/conduit/t_conduit_execution.cpp
siramok: I was able to get this done.

Issue #1614, opened 2026-05-26
Conduit Device Support Ongoing Development


JustinPrivitera:
## High-Level Goals
- Develop a device execution strategy for Conduit
- Port Conduit's algorithms to be able to run on device
    - **Our focus will be Blueprint transforms to start**
    - Performance analysis will be crucial
- Develop tests
    - For our execution model
    - For performance of ported algorithms
- Write comprehensive documentation for Device Support
   - Including code comments where they make sense.

## Features

- [x] Implement the strawman: #1151 - implemented [here](https://github.com/llnl/conduit/blob/dc6f9eff3e41b02506862e94f9bbdfa91ff72c1d/src/tests/conduit/execution_test_utils.hpp#L113): #1358
- [x] infrastructure to control where we run at a high level - so that convert, transform, etc can have a uniform way of users asking to run on device. - progress in #1358
- [x] Make sure CUDA and HIP are on par with one another (CUDA is likely behind)
- [x] data accessor/array bulk set/fill parallel acceleration using foralls #1671 
- [x] implement device sort using RAJA - @siramok #1623
- [x] make sync strategy an enum
- [x] openMP reduce + atomics are missing
- [ ] Port the algorithms in [Conduit Blueprint Mesh](https://github.com/llnl/conduit/blob/develop/src/libs/blueprint/conduit_blueprint_mesh.hpp) (as many as make sense) (start with conduit_blueprint_mesh transforms/algorithms). See #1637 for a list of algorithms to port.
   - [ ] partitioner
   - [x] topology/coordset transforms
   - [ ] generate transforms
   - [ ] matset transforms
   - [ ] polytopal transforms

## Testing

- [x] Execution tests that cover the strawman, reducers, atomics, policy choices - #1358
- [x] Performance tests for ported algorithms (tests for algorithms are going to be more oriented towards performance, since we already have lots of correctness tests. We want to test the full matrix of options for running on host/device based on source/destination data starting on host/device).
- [x] Test with both CUDA and HIP
- [x] Add node-backed tests for data accessor and data array - @siramok #1630

## Performance

- [x] Incorporate Caliper for measuring performance of algorithms
- [ ] Get hatchet/thicket support in to be able to compare runtimes
- [ ] discover how to profile compilation

## Documentation

- [x] Write documentation for our Device Support paradigm
- [x] include strawman examples in the documentation to demonstrate use
- [x] We should document how to ask algorithms to execute on the device
[...]

PR #1634, opened 2026-06-25
Device Support: Port benchmark to the device execution model + TU compilation fixes

siramok:
We decided that it would make more sense to backport the benchmarking infrastructure to `develop` then sync [#1358](https://github.com/llnl/conduit/pull/1358) with it. This makes it easy for us to directly compare API performance before vs after device support.

The purpose of this PR has subsequently changed, the remaining changes are to enable device support within the benchmarking infrastructure and `conduit_blueprint_mesh.cpp`. In practice, this means that the benchmark is extended to now loop over the matrix of possibilities (host/device for source, execution, and output locations) and `conduit_blueprint_mesh.cpp` has received fixes to be built as a CUDA/HIP TU.

Tested and works on both matrix and tuo.

OLD:
~This PR introduces benchmarking infrastructure to measure the performance of APIs that have been ported to use the new execution model. It currently benchmarks across a matrix of possibilities: host vs device execution, varying data sizes, and all possible execution location combinations (source, execute, and output).~

~Includes:~
- ~Benchmark infrastructure (flexible enough to add new benchmarks relatively easily).~
- ~Coordset conversion benchmarks.~
- ~Changes that enable `conduit_blueprint_mesh.cpp` to be built as a CUDA/HIP TU.~
- ~Caliper outputs designed to be consumed by hatchet or thicket.~
- ~A python script to plot the caliper outputs (thanks to AI).~


src/libs/conduit/conduit_benchmark.hpp
```diff
@@ -0,0 +1,316 @@
...
+inline std::vector<EP>
+get_enabled_policies()
+{
+    std::vector<EP> policies;
+    policies.reserve(7);
+
+    // Since asking for compile-time disabled policies will throw an error
+    // (and subsequently cause our benchmark "tests" to fail), we have to
+    // do extra work to avoid asking for disabled policies
+    detail::add_if_enabled(policies, EP::is_serial_enabled(),   [] { return EP::serial(); });
```
JustinPrivitera: couldn't we use `try`/`catch` for this purpose?

src/libs/conduit/conduit_benchmark.hpp
```diff
@@ -0,0 +1,316 @@
...
+            // Warm-up
+            for (index_t i = 0; i < warmup; i++)
+            {
+                fn(policy, mode);
+            }
+
+            std::cout << "\n" << policy.policy_name() << "," << warmup << "," << iterations << std::endl;
+            
+            // TODO: Do we care to have bespoke timing in the case that conduit was built
+            // without caliper?
```
JustinPrivitera: Probably not, but we want to use other performance metrics. Caliper doesn't tell us everything we would like to know. Maybe we could explore some of these tools:
 - LC Hosted tool for GPU profile vis: https://lc.llnl.gov/perfetto/
 - nsys: https://github.com/llnl/conduit/issues/1614#issuecomment-4743925297
 - rocprof for AMD
 - Hatchet: https://github.com/LLNL/hatchet
 - Thicket: https://github.com/llnl/thicket

I think ultimately some combination of tools will be able to give us more insightful metrics than just using Caliper. Caliper is still useful as a baseline since we have it set up for easy use in Conduit, but it may be useful to look at expanding our Caliper options since it can do much more than we use it for: https://github.com/Alpine-DAV/ascent/wiki/Caliper-How-To

src/libs/conduit/conduit_execution.cpp
```diff
@@ -709,6 +709,11 @@ get_host_allocator_id()
 }
 
 //---------------------------------------------------------------------------//
+// TODO: All of our tests call this function unconditionally (regardless of
+// whether conduit was built with device support or not). Does this function
+// need to exist? Seems like we could reduce boilerplate by hiding this from the
+// user and just doing it automatically upon init or lazily upon first call of
+// any execution:: func
```
JustinPrivitera: There is a note about this under https://github.com/llnl/conduit/issues/1614 _Investigations_ subsection

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -0,0 +1,96 @@
...
+                                     BENCHMARK_ARRAY_SIZE,
+                                     mesh);
+
+    const Node &coordset = mesh["coordsets"].child(0);
+    Node dst;
+    blueprint::mesh::coordset::rectilinear::to_explicit(coordset, dst);
+}
+
+//-----------------------------------------------------------------------------
+TEST(blueprint_mesh_transform, coordset_to_explicit)
```
JustinPrivitera:
```suggestion
TEST(blueprint_mesh_transform_execution, coordset_to_explicit)
```
All other Blueprint mesh tests have a further qualified name.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -0,0 +1,96 @@
...
+using namespace conduit;
+using EP = execution::ExecutionPolicy;
+
+index_t BENCHMARK_ARRAY_SIZE            = 64;
+index_t BENCHMARK_NUM_WARMUP_ITERATIONS = 10;
+index_t BENCHMARK_NUM_ITERATIONS        = 100;
+
+//-----------------------------------------------------------------------------
+void
+coordset_uniform_to_explicit(EP /*policy*/,
```
JustinPrivitera: We are missing `coordset_uniform_to_rectilinear`

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+    CONDUIT_ANNOTATE_MARK_FUNCTION;
+    benchmark::exec(coordset_uniform_to_explicit,
+                    BENCHMARK_NUM_WARMUP_ITERATIONS,
+                    BENCHMARK_NUM_ITERATIONS,
+                    {benchmark::FillMode::None});
+    benchmark::exec(coordset_rectilinear_to_explicit,
+                    BENCHMARK_NUM_WARMUP_ITERATIONS,
+                    BENCHMARK_NUM_ITERATIONS,
+                    {benchmark::FillMode::None});
```
JustinPrivitera: I don't understand; where are the execution options set? If we want to run the ported algorithms with different configurations, we need to set the execution options and then iteratively loop over the available execution options for execution location, output location, fallback location, and sync strategy. We will also need to experiment with having source data start on the host versus device. For these coordset transforms, we create destination data so we don't have to worry about where destination data starts.

siramok: Not implemented yet, but on the todo list.

src/tests/conduit/t_conduit_execution_benchmark.cpp
```diff
@@ -0,0 +1,347 @@
...
+    CONDUIT_ANNOTATE_MARK_FUNCTION;
+    benchmark::exec(atomic_max<int32>, BENCHMARK_NUM_WARMUP_ITERATIONS, BENCHMARK_NUM_ITERATIONS);
+    benchmark::exec(atomic_max<int64>, BENCHMARK_NUM_WARMUP_ITERATIONS, BENCHMARK_NUM_ITERATIONS);
+}
+
+//-----------------------------------------------------------------------------
+// Reducer benchmark tests
+//
+// TODO: We are curious if we gain anything by using the bonus reducer policies:
+// https://raja.readthedocs.io/en/main/sphinx/user_guide/cook_book/reduction.html
```
JustinPrivitera: This is true, but I don't think this is an experiment we would necessarily want to put in our test suite and have it live forever. Playing with the different execution policy choices would help us to settle on the right implementation choices, so I imagine it would be a one-time experiment.

src/tests/conduit/t_conduit_execution_benchmark.cpp
```diff
@@ -0,0 +1,347 @@
...
+    });
+}
+
+//-----------------------------------------------------------------------------
+// Atomic benchmark tests
+//
+// TODO: OpenMP w/ no RAJA seems incredibly slow (most obvious when doing many
+// iterations). We should look at what RAJA does differently to improve
+// performance. Maybe limiting the number of omp threads to 1-4 before atomics
+// would help?
```
JustinPrivitera: This gets to a deeper question about how well we want to support that case. What are our design goals for each case? I think we want people to use RAJA and we want to ensure that algorithms still work if they don't have RAJA, but we may not be promising excellent performance without RAJA. We probably don't need to reinvent the wheel if RAJA can do these same things better.

siramok: Good point and I agree, it would create less work for us (and whoever maintains this into the future) to not have to support it. I guess it's ultimately up to Cyrus?

src/tests/blueprint/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ set(BLUEPRINT_TESTS t_blueprint_smoke
                     t_blueprint_mesh_verify
                     t_blueprint_mesh_convert
                     t_blueprint_mesh_transform
+                    t_blueprint_mesh_transform_benchmark
```
JustinPrivitera, line 21: we will need to add this
```cmake
set(conduit_device_test_sources
    t_blueprint_mesh_transform_benchmark.cpp)

conduit_set_source_device_properties("${conduit_device_test_sources}")
```
somewhere in this cmake file so that our test is compiled as a CUDA/HIP TU.

src/tests/conduit/t_conduit_execution_benchmark.cpp
JustinPrivitera: Ultimately all things are up to Cyrus. :) I think we want the algorithms to work but it is an open question how much we try to improve things for those who want openMP without RAJA. I don't know that we've settled on an answer yet.

src/libs/conduit/conduit_benchmark.hpp
siramok: I've switched to using `try` / `catch` and was able to simplify things.

src/libs/conduit/conduit_benchmark.hpp
siramok: I removed this TODO.

Looks like caliper can be configured to output a file that hatchet can consume. Furthermore, it looks like thicket builds on hatchet, and therefore consumes the same data. I've changed the caliper config to output this file, but I think the file is currently getting overwritten for each benchmark combo, so I need to investigate a fix.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: This has been added.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: This has been added for execution location, however it revealed a problem that will need discussion on how to proceed:

Essentially, it's not good enough for `t_blueprint_mesh_transform_benchmark.cpp` to be compiled with CUDA/HIP TUs. We also have to do that for `conduit_blueprint_mesh.cpp` to get device support in the coordset transforms.

The problem is that attempting to do so on the entire file is currently not possible. NVCC throws several compilation errors due to the `polyhedral_*` functions (which have not been ported).

src/libs/conduit/conduit_benchmark.hpp
siramok: I fixed this by creating one .cali file with all of the profiler data. It works, however the purpose of thicket seems to be for making it easy to consume many .cali files. I added a new TODO to investigate creating a separate .cali per benchmark to see what can be gained by going that route.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: I was able to build `conduit_blueprint_mesh.cpp` as a TU after making several fixes.

siramok:
Still need to do:

- ~Add support for iteratively ramping up data sizes. We want to understand how the algorithms perform at varying input sizes ranging from very small to very large.~
- ~Validate that all of the caliper annotations are in correct places. There doesn't seem to be a need to produce multiple .cali files per Cyrus.~
- ~I discovered that the changes to `conduit_blueprint_mesh.cpp` for GPU TU compilation only make NVCC happy, but HIP still has compilation errors. Those will need to be fixed.~
- ~Create a python script for making scaling plots out of these .cali files. It can live next to the other test files per Cyrus. Ascent has a python script like this that gets called automatically via cmake, so that could perhaps give some inspiration.~

siramok:
Mark brought up a good point during today's Visit meeting: are we interested in getting benchmark data for the APIs as they existed before they were ported to the execution model (to compare old vs new)?

After discussing this, it sounds like yes, we want to try to get this working.

siramok:
At this point:
- The benchmarking code seems to work well and has most (if not all) of the features implemented that we've discussed. Of course, it may need refinement as we create new benchmarks and have to accommodate their differences.
- Fixes are in place to get both CUDA/HIP compiling `conduit_blueprint_mesh.cpp` as a TU. This was done so that we could properly perform host vs device benchmarking, and I've been able to collect/plot host vs device results from both matrix and tuolumne.
- I've included a python script to get us started on plotting the benchmark data that this produces (AI did most of the work but I reviewed it and refined it).

What seems to remain:
- Get feedback about how the plots look. Get feedback about what other types of plots would be useful to have. Improve the plotting script based on that feedback.
- Do more empirical testing to see what good defaults should be for warmup iterations, benchmark iterations, and data sizes that we iterate over.
- Address feedback about the PR itself.

siramok:
I took a pass at porting the topology conversions to the execution model in #1640. It depends on this PR, which I only mention because it should give you a rough idea for how future API ports + benchmarks would look using the current infra.

siramok:
The ascent team gave me verbal feedback with respect to improving the plots, so I'll be making some changes to address their feedback.

Not related to this PR (probably), but there's currently a big discrepancy between CUDA and HIP performance. We came to the conclusion that it would help to print some basic system information, e.g. the number of GPUs and OpenMP threads being used in a run, to help assure us that the comparisons are apples to apples.

siramok:
I've updated the plots based on feedback:
- Changes to make it easier to see trends/reduce overlapping lines.
- More plot variants (e.g. device only, host only).
- The number of openmp threads is now included in the caliper output (if built with openmp), and the plots now include that number if available.

siramok:
``` #include "conduit.hpp"
  #include "conduit_blueprint_mesh_examples.hpp"

  using namespace conduit;

  static void copy_numeric_arrays_to_device(const Node &src,
                                            Node &dst,
                                            index_t device_alloc)
  {
      if(src.dtype().is_object())
      {
          NodeConstIterator itr = src.children();
          while(itr.has_next())
          {
              const Node &src_child = itr.next();
              copy_numeric_arrays_to_device(src_child, dst[itr.name()], device_alloc);
          }
      }
      else if(src.dtype().is_list())
      {
          NodeConstIterator itr = src.children();
          while(itr.has_next())
          {
              const Node &src_child = itr.next();
              copy_numeric_arrays_to_device(src_child, dst.append(), device_alloc);
          }
      }
      else if(src.dtype().is_number() && src.dtype().number_of_elements() > 1)
      {
          dst.set_allocator(device_alloc);
          dst.set(src); // requires a device-aware memcpy handler
      }
      else
      {
          dst.set(src); // keep strings, dims, small metadata on host
      }
  }

  Node braid_host;
  blueprint::mesh::examples::braid("hexs", 32, 32, 32, braid_host);

  Node braid_dev;
  copy_numeric_arrays_to_device(braid_host,
                                braid_dev,
                                execution::get_device_allocator_id());
 ```

File: src/libs/conduit/conduit_benchmark.hpp [hunk 10]
```diff
@@ -19,11 +19,7 @@
 
 #include "conduit_annotations.hpp"
 #include "conduit_data_type.hpp"
-
-// This is not needed after device support is added back
-#if defined(CONDUIT_USE_OPENMP)
-#include <omp.h>
-#endif
```
JustinPrivitera, line 26: did we decide that we might need this for the non-RAJA openMP case?

src/libs/conduit/conduit_benchmark.hpp
```diff
@@ -62,10 +60,17 @@ inline
...
 get_exec_configs()
 {
-    // In the pre-device execution model world, we don't have the concept of
-    // host vs device execution
-    //                                src       exec      dest
-    std::vector<ExecConfig> configs{{"host", "host", "host"}};
+    // src, exec, dest
+    std::vector<ExecConfig> configs{
+        {"host",   "host",   "host"},
+        {"host",   "host",   "device"},
```
JustinPrivitera: maybe this could be annotated a little more cleanly... instead of `// src, exec, dest` you could be more explicit about what each column means.

src/libs/conduit/conduit_benchmark.hpp
```diff
@@ -87,6 +95,13 @@ exec(const char *name,
         {
             config.dim_size = dim_size;
 
+            // Set all execution options for this configuration
+            Node exec_opts;
+            exec_opts["execution_location"].set(config.exec_location);
+            exec_opts["output_location"].set(config.output_location);
+            exec_opts["sync_strategy"].set("sync");
```
JustinPrivitera: do we always want the sync strategy to be "sync"? I think we will want to "assume" sometimes also, depending on where we want the output data to be.

src/libs/conduit/conduit_fixed_size_map.hpp
JustinPrivitera, line 1: Why do these files need to be device annotated?

src/tests/blueprint/CMakeLists.txt
```diff
@@ -50,6 +50,11 @@ foreach(TEST ${BLUEPRINT_TESTS})
                  FOLDER tests/blueprint)
 endforeach()
 
+set(conduit_device_test_sources
+    t_blueprint_mesh_transform_benchmark.cpp)
+
+conduit_set_source_device_properties("${conduit_device_test_sources}")
+
```
JustinPrivitera:
```suggestion
set(conduit_blueprint_device_test_sources
    t_blueprint_mesh_transform_benchmark.cpp)

conduit_set_source_device_properties("${conduit_blueprint_device_test_sources}")

```

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -17,21 +17,71 @@
...
+        {
+            const Node &src_child = itr.next();
+            copy_numeric_arrays_to_device(src_child, dst.append(), device_alloc);
+        }
+    }
+    else if(src.dtype().is_number() && src.dtype().number_of_elements() > 1)
+    {
+        dst.set_allocator(device_alloc);
+        dst.set(src);
+    }
```
JustinPrivitera, line 59: I wonder if this is still too strong and if we lose things like origin and strides. I'm not sure where those should live.

JustinPrivitera (commented):
I am mainly wondering about the conduit headers that were device annotated as part of this.

src/libs/conduit/conduit_benchmark.hpp
siramok: We do still need `omp.h`, but `conduit_execution_policy.hpp` includes it and `conduit_benchmark.hpp` now includes `conduit_execution_policy.hpp` for the `execution::` stuff.

siramok: In other words, having it here doesn't help or hurt, as we'll end up getting it either way.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
JustinPrivitera: Maybe it would be an interesting test of things to be aggressive in moving these things to device and seeing where things break. Then we can fix as we go to make sure our tools are robust to those cases.

src/libs/conduit/conduit_benchmark.hpp
JustinPrivitera: sounds good; let's leave it removed then.

src/libs/conduit/conduit_benchmark.hpp
siramok: I took a stab at being more explicit all around, maybe overly so. Let me know if this is what you were envisioning.

JustinPrivitera (commented):
These comments are nice :)

src/libs/conduit/conduit_benchmark.hpp
siramok: I did think about this question, and convinced myself that if `sync` is always a no-op in cases where memory doesn't need to move (maybe it's not always a no-op), then defaulting to `sync` should always do the correct thing since we always explicitly set the execution and output locations. My understanding is that `assume` is mainly for when you want the execution model to determine those locations implicitly, which does not apply in this case. I'd be curious to hear your thoughts.

src/libs/conduit/conduit_fixed_size_map.hpp
siramok: IIRC, not device annotating them was causing HIP to have compilation errors, I guess because `conduit_blueprint_mesh` uses them somewhere and TU compilation is all-or-nothing for the entire file.

I could dig deeper and give you a more specific answer if you wish.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: I agree. It seems to work fine for now, but I went ahead and added a TODO to investigate. Whether you want that investigation to happen now or in the future, I'm fine either way.

src/libs/conduit/conduit_fixed_size_map.hpp
siramok: I did a quick and dirty test where I restored `conduit_fixed_size_vector.hpp` to the pre-device annotated version and tried to compile. HIP complains about there being references to `__host__` functions in a `__host__ __device__` function. It specifically complains about lines 8509 and 8519, which appear to be normal `std::vectors` but apparently are using this bespoke type under the hood.

src/libs/conduit/conduit_benchmark.hpp
siramok: We agreed that there's value in exploring this, I added a TODO.

src/libs/conduit/conduit_benchmark.hpp
siramok: We discussed it verbally, and decided it was good.

src/libs/conduit/conduit_benchmark.hpp
siramok: I dug into this more and added support for looping over both sync strategies. I understand how they differ a bit better: seems like they only come into play if the execution location and output location are different, and should be identical no-ops otherwise. I extended the configs to have "assume" variants for cases where the execution and output locations are different.

What I found most convincing is that "assume" is the default strategy, so it does make sense to be benchmarking it. What I think this gives us is the ability to understand how much it costs to guarantee that the output lives where we tell it to ("sync") vs if we're fine with getting it in whatever memory space the execution was performed in (faster due to no memcpy, but could be in the wrong location if the user needs it somewhere specific).

src/libs/conduit/conduit_fixed_size_map.hpp
siramok: We spoke verbally and agreed that these changes are ok.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: I investigated. My understanding was that the single value accessors (e.g. origin or stride values) just directly dereference a pointer to get the value back out, and don't account for device memory. I verified that copying single elements to device memory would lead to a segfault in our benchmarks due to host code eventually dereferencing a pointer to device memory. I've updated the TODO to confirm that we're doing the correct thing by only moving numeric arrays to device memory.

src/libs/conduit/conduit_benchmark.hpp
```diff
@@ -66,33 +67,51 @@ get_exec_configs()
...
+        {"host",           "device",           "host",          "assume"},
+        {"host",           "device",           "device",        "sync"},
+        {"device",         "host",             "host",          "sync"},
+        {"device",         "host",             "device",        "sync"},
+        {"device",         "host",             "device",        "assume"},
+        {"device",         "device",           "host",          "sync"},
+        {"device",         "device",           "host",          "assume"},
+        {"device",         "device",           "device",        "sync"},
+#endif
     };
```
JustinPrivitera, line 115: We need to ask @cyrush when assume is going to make sense. Now that we are porting algorithms I want to better understand the intended use cases, because it feels redundant at this point.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
JustinPrivitera: Cool, thanks for this.

src/libs/conduit/conduit_fixed_size_map.hpp
```diff
@@ -13,12 +13,23 @@
...
+
 //-----------------------------------------------------------------------------
 // -- begin conduit:: --
 //-----------------------------------------------------------------------------
 namespace conduit
 {
 
+// Workaround for TU compilation, since CONDUIT_ERROR does not work on device
+#if defined(CONDUIT_DEVICE_COMPILE)
+#define CONDUIT_ERROR_COMPAT_WRAPPER(MSG) assert(false && MSG)
```
JustinPrivitera:
```suggestion
#define CONDUIT_ERROR_COMPAT_WRAPPER(MSG) assert(false)
```
I think this is brittle with message and not sure we need error messages on device perhaps.

JustinPrivitera (changes requested):
Final little things

src/libs/conduit/conduit_benchmark.hpp
siramok: I think that's a good discussion to have. I can envision a world in which the sync strategies go away and the behavior is instead controlled in some other way, like whether the user sets an explicit output location or not.

If the execution and output locations are the same, I'm not sure what currently happens, but I would hope that we always avoid doing an extra memcpy by simply swapping out a pointer behind the scenes (essentially what "assume" already does).

If the execution and output locations are different, then to me, that implies the user is intentionally asking for a memcpy such that the result lives in a specific memory space. Because if you're not asking, it implies that you don't want to have to care (or need a reason to care), in which case it's almost always going to be better to leave the output in the same memory space as execution (without seeing a bigger picture to tell me otherwise).

But in that scenario, would it be just as good and clear (if not more so) to expose a `conduit::execution::move_to_host(Node&)` and `conduit::execution::move_to_device(Node&)` that makes it extra explicit?

JustinPrivitera (approved):
🦾

PR #1635, opened 2026-06-26
Device Support: Port `coordset::to_explicit` to use execution model

JustinPrivitera:


src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1004,7 +1004,7 @@ convert_coordset_to_explicit(const std::string &base_type,
         // rectilinear transform case.
         const Node &src_cvals_node = coordset.has_child("values") ?
             coordset["values"][csys_axis] : info;
-        float64_accessor src_cvals_acc = src_cvals_node.value();
+        float64_accessor src_cvals_acc(src_cvals_node);
```
siramok, line 1007: Nice, I know what this means.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1019,12 +1019,27 @@ convert_coordset_to_explicit(const std::string &base_type,
             dim_block_count *= (i < j) ? dim_lens[j] : 1;
         }
 
+        // execution setup
+        conduit::execution::ExecutionPolicy policy = is_base_rectilinear ?
+            conduit::execution::get_execution_policy(src_cvals_node) :
+            conduit::execution::get_execution_policy();
```
siramok, line 1025: Had to stare for a few minutes and look at the full file, but I think I understand what's going on now:

If we're converting a rectilinear coordset -> explicit, the incoming `Node` will already have a buffer of values that we have to access and convert from. So we access those directly (whether they live on the host or device) and make sure to allocate `dst` memory in the same location. The only other case is uniform -> explicit, in which case the incoming `Node` does not already have a buffer of values to access and convert from. So we could end up allocating in either place, depending on if the user overrode the fallback location or not.

src/libs/conduit/conduit_execution.cpp
```diff
@@ -475,7 +475,7 @@ class ExecutionOptions
     }
 
     //------------------------------------------------------------------------
-    static ExecutionPolicy get_execution_policy_helper(Node *src_node)
+    static ExecutionPolicy get_execution_policy_helper(const Node *src_node)
```
siramok, line 478: Nice.

siramok (approved):
LGTM, no feedback to address.

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: Exactly. The function could probably use more comments...

PR #1639, opened 2026-07-01
Device Support: Fix RAJA smoke test when OpenMP + device are both enabled

siramok:
I noticed that `t_raja_smoke` was segfaulting when RAJA was built with both OpenMP and CUDA. Changing the order in which `ExecPolicy` is evaluated fixes the segfault.


JustinPrivitera:
Hmm, this is odd. Any idea why this was segfaulting to begin with?

siramok:
IIRC, it was happening because the OpenMP policy gets selected (due to coming first in the if/else chain) as the execution policy for a device-based execute.

JustinPrivitera (approved):

PR #1640, opened 2026-07-02
Device Support: Port topology::*::to_unstructured to the new execution model

siramok:
Depends on #1634 for changes that make TU compilation and benchmarking possible.

Includes:
- Port of the `topology::*::to_unstructured` transforms to the new execution model.


src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1114,14 +1116,15 @@ convert_topology_to_rectilinear(const std::string &/*base_type*/,
                                 conduit::Node &cdest)
 {
     CONDUIT_ANNOTATE_MARK_FUNCTION;
-    // bool is_base_uniform = true;
```
JustinPrivitera, line 1120: why remove this guy? I think the intent is that this function could be for any kind of topo, but the only kind it makes sense to support right now is from uniform to rectilinear. So it doesn't hurt to have the comment signalling intent.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1114,14 +1116,15 @@ convert_topology_to_rectilinear(const std::string &/*base_type*/,
...
 
     dest.reset();
     cdest.reset();
 
     const Node *coordset = bputils::find_reference_node(topo, "coordset");
     blueprint::mesh::coordset::uniform::to_rectilinear(*coordset, cdest);
 
+    const index_t allocator_id = conduit::execution::get_output_allocator_id();
+    dest.set_allocator(allocator_id);
```
JustinPrivitera: I am not sure we want to put rectilinear topo data on device. We don't want strings to live there: https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#implicit-topology

We should probably leave everything for the rectilinear topology on host and not use any custom allocators.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1151,18 +1154,24 @@ convert_topology_to_structured(const std::string &base_type,
...
+        conduit::execution::get_output_allocator_id((*coordset)["values"][csys_axes[0]]) :
+        conduit::execution::get_output_allocator_id();
+
     dest["type"].set("structured");
     dest["coordset"].set(cdest.name());
     if(topo.has_child("origin"))
     {
+        dest["origin"].set_allocator(allocator_id);
         dest["origin"].set(topo["origin"]);
     }
```
JustinPrivitera: Again, I am not sure we want this to depend on the allocator. I think the only pieces we care about using the allocators for are big data arrays. The origin and other 2-3 element arrays might have direct access in different helpers so I think moving them over to device is potentially dangerous and probably unnecessary.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1174,8 +1183,11 @@ convert_topology_to_structured(const std::string &base_type,
         // than the number of points along each dimension.
         src_dlen_node.set(src_dlen_node.to_int64() - 1);
 
-        Node &dst_dlen_node = dest["elements/dims"][logical_axes[i]];
-        src_dlen_node.to_data_type(int_dtype.id(), dst_dlen_node);
+        Node host_dlen_node;
+        src_dlen_node.to_data_type(int_dtype.id(), host_dlen_node);
+
+        dest["elements/dims"][logical_axes[i]].set_allocator(allocator_id);
+        dest["elements/dims"][logical_axes[i]].set(host_dlen_node);
```
JustinPrivitera: Same thing for these. I think we can leave these transforms alone and focus on the ones that have to generate big arrays. These are just tiny arrays of 3 elements at most.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: Readded.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: Removed.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: Removed.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: Removed.

siramok:
Execution model stuff has been removed in places where the computation being performed was trivial. Therefore, `to_unstructured` is now the only transform that this PR ports over.

We discussed earlier today about how it would be good to consider further optimizations on these existing APIs. While I was cleaning things up, for fun, I asked Claude about whether the contents of the `forall` could be optimized at all.  Its suggestions seemed to make sense, compiled fine, and the correctness tests still pass.

They are:
- Instead of first writing all vertex indices in raw bit order and then doing a second pass to fix the ordering, the indices are now directly written out in the correct order.
- `grid_ijk_to_id`/`grid_id_to_ijk` were previously implemented as generic N-dimensional algorithms, but since they were only ever operating on fixed size 3-element arrays, it turned out that the math could be simplified for both.

I was able to measure a minor improvement even on small input meshes, so I suspect it'll end up scaling nicely.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1196,9 +1185,28 @@ convert_topology_to_unstructured(const std::string &base_type,
...
     if(is_base_structured)
     {
-        cdest.set(*coordset);
+        cdest["type"].set((*coordset)["type"]);
+        for(index_t i = 0; i < num_axes; i++)
+        {
+            Node &dst_cvals_node = cdest["values"][csys_axes[i]];
+            dst_cvals_node.set_allocator(allocator_id);
+            dst_cvals_node.set((*coordset)["values"][csys_axes[i]]);
+        }
```
JustinPrivitera: no need for loop....

trust... but verify!

siramok: We discussed that it should be sufficient to set the parent allocator, and new children under that parent should inherit its allocator as well. In other words, it's not necessary to loop through the children and set the allocator explicitly.

JustinPrivitera:
```c++
execution::DeviceMemory::is_device_ptr(ptr)
```

siramok:
Yesterday Justin and I spent time validating that this new/optimized approach for `to_unstructured` produces an identical output to the old approach. It indeed does, as verified by saving the mesh outputs of old vs. new and diffing them.

Furthermore, since the optimization suggestions came from AI, we agreed that it's necessary to understand well what the changes do before proceeding. It's my next task to spend time understanding what's going on well enough to create detailed documentation.

We agreed that `const index_t out_i = (i & 2) ? (i ^ 1) : i;` is confusing, and that an attempt should be made to achieve the same but written more clearly.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: This still needs to be done.

siramok:
The clearer writing + comments are in. Just need to fix the looping feedback, then this is ready for a new review.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: This has been addressed. You were correct, I verified that it's good enough to set a parent's allocator and then just do `Node::set()` for the entire tree that you want to copy over.

siramok:
All feedback has been addressed, this PR is ready for a re-review.

This PR already changes `::to_unstructured` substantially, so I figured I may as well go all the way and make it as nice as I could. I spaced things out better, improved variable names, made more things `const`, and took a final pass to make sure that it was as optimized as I could make it.

siramok:
On my local machine I compared the `develop` branch vs this PR using a 64x64x64 braid mesh, 10 warmups, and 100 iterations. Only the `to_unstructured` tests were measured. This is just the raw topology test time as reported by gtest:

develop (serial): 15765 ms
This PR (serial, only host->host->host): 2985 ms
This PR (cuda, only device->device->device): 546 ms

That's a ~5.2x improvement in `develop` serial vs this PR serial and a ~28.8x improvement in `develop` serial vs this PR in the best case scenario (all data originating, executing, and ending up on device). This is of course just an average of all base types, it could be broken down to understand exact improvements for particular base types -> unstructured.

(Am not sure what the improvements will look like once we move to the new infrastructure in #1653, which significantly optimizes the benchmarks themselves.)

JustinPrivitera (approved):
Thanks for this!

PR #1642, opened 2026-07-06
Node::swap add info on ownership of schema

cyrush:


JustinPrivitera (approved):

cyrush:
not sure this should be a pure swap, the ownership semantics are complicated. 
we should write out a table with all the possible cases to understand.

cyrush:
This is not a good change, the ownership semantics have to do with how the schema tree is owned, and a simple swap will not work.

PR #1643, opened 2026-07-07
Add a pre-device execution model version of the benchmark infrastructure

siramok:
This PR reproduces the benchmarking infrastructure + benchmarks for coordset and topology conversion + plotting script (largely AI) from #1634 and #1640, minus all of the extra `execution::` and TU compilation changes necessary for device support. We decided that this would ultimately be the best route forward, so that we can have an easy way to benchmark the Conduit APIs pre- and post-device support. It has the added benefit of splitting PRs to make the TU changes easier to review.

Initial results seem to confirm that adding device support does indeed have performance benefits (good news for us). The main practical difference is that this version doesn't loop over the different host/device combinations (since that wouldn't make sense), so `host->host->host` is effectively all it's doing here. Since we no longer have the concept of execution backends either, the plotting script was updated to treat that parameter as optional, so it should create similar plots from both pre- and post-device support .cali files but avoid creating the breakdown plots with `use_with`, `forall`, and `sync`.

Tested and working on both matrix and tuo.


src/libs/conduit/conduit_benchmark.hpp
```diff
@@ -0,0 +1,163 @@
...
+    return locations;
+}
+
+//-----------------------------------------------------------------------------
+inline
+std::vector<ExecConfig>
+get_exec_configs()
+{
+    std::vector<ExecConfig> configs;
+    const auto locations = get_enabled_locations();
```
JustinPrivitera: Why not fold this in like this?

```suggestion
//-----------------------------------------------------------------------------
inline
std::vector<ExecConfig>
get_exec_configs()
{
    std::vector<ExecConfig> configs;
    const std::vector<std::string> locations = {"host"};
```

It doesn't look to me like `get_enabled_locations()` is ever used anywhere else.

JustinPrivitera: Going further, why not explicitly enumerate the execution configs (removing `get_exec_configs()` entirely)? Are we ever going to have additional locations other than host and device? If we only have host/device then there aren't too many, right?

Check out these tests elsewhere in Conduit: https://github.com/llnl/conduit/blob/025d5d8a30e30242b7916adb84d14824ad671439/src/tests/relay/t_relay_io_silo.cpp#L3170-L3200

I am biased because I wrote them, but I like that all the cases are explicitly enumerated. It is easy to comment out lines when you only want to debug certain pieces.

src/libs/conduit/conduit_benchmark.hpp
```diff
@@ -0,0 +1,163 @@
...
+#include <string>
+#include <vector>
+
+#include "conduit_annotations.hpp"
+#include "conduit_data_type.hpp"
+
+// This is not needed after device support is added back
+#if defined(CONDUIT_USE_OPENMP)
+#include <omp.h>
+#endif
```
JustinPrivitera, line 26: Will we need this for the non-RAJA openMP case?

JustinPrivitera (commented):
I looked at the first few files.

src/libs/conduit/conduit_benchmark.hpp
siramok: Good question, I believe the answer is yes. This header is only used for querying the number of omp threads available at benchmark time so that it can be plumbed through to the caliper output. Then the thread count gets added to the plots as an extra piece of info about the run.

Is OpenMP relevant to us in a pre-device execution model world? Even if none of the currently benchmarked transforms use OpenMP, are there any pre-DEM APIs that do leverage it? If yes, then I think it's worth leaving in. If no, then I could remove it from this PR and leave it as a change that comes back in the other PR.

src/libs/conduit/conduit_benchmark.hpp
siramok: I'm not opposed. There's only host + device and 3 execution locations, so 9 unique combinations total.

src/libs/conduit/conduit_benchmark.hpp
JustinPrivitera: OpenMP is relevant to us in the pre-device execution model world. There are blueprint mesh transforms that use OpenMP and use an older, simpler execution model developed by Brad Whitlock.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include "conduit.hpp"
+#include "conduit_annotations.hpp"
+#include "conduit_benchmark.hpp"
+#include "conduit_blueprint.hpp"
+
+#include "gtest/gtest.h"
+
+using namespace conduit;
+
+std::vector<index_t> BENCHMARK_DIM_SIZES = {2, 4, 8, 16, 32, 64};
```
JustinPrivitera, line 20: We may want to do some very large runs that take a while to execute to see how things are at scale. It would be good to do some tests with meshes that are at least 1 million elements.

siramok: I agree. The default sizes are relatively small keeping in mind that CI will have to re-run these benchmarks for every future PR/commit. Plus the runtime will get worse as we add more benchmarks. So truthfully, even these sizes may be too much to leave as default.

siramok: But for the serious data collection runs, we will indeed need to crank the numbers.

src/libs/conduit/conduit_benchmark.hpp
siramok: I've fixed it for this PR. I'll have to finish the job in the in follow-up PR.

siramok:
As best can I can tell, the CI failure is unrelated to the most recent commit.

src/libs/conduit/conduit_benchmark.hpp
```diff
@@ -0,0 +1,138 @@
...
+//-----------------------------------------------------------------------------
+inline
+std::vector<ExecConfig>
+get_exec_configs()
+{
+    // In the pre-device execution model world, we don't have the concept of
+    // host vs device execution
+    std::vector<ExecConfig> configs{{"host", "host", "host"}};
+    return configs;
+}
```
JustinPrivitera, line 70: 1. Do we still need `ExecConfig`?
2. One thing I liked about the example I shared was that each column in the table was annotated so it was clear what we were looking at. Maybe something like this? But perhaps better formatted :)


```suggestion
//-----------------------------------------------------------------------------
inline
std::vector<ExecConfig>
get_exec_configs()
{
    // In the pre-device execution model world, we don't have the concept of
    // host vs device execution
    //                                src       exec      dest
    std::vector<ExecConfig> configs{{"host", "host", "host"}};
    return configs;
}
```

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
JustinPrivitera: 64^3 = 262144, so maybe we should try at least 128 for the large dim size.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -0,0 +1,176 @@
...
+    // Coordset setup
+    const Node &host_coordset = mesh["coordsets"].child(0);
+    Node &src_coordset = src["coordsets"][host_coordset.name()];
+    src_coordset.set(host_coordset);
+
+    // Topology setup
+    const Node &host_topo = mesh["topologies"].child(0);
+    Node &src_topo = src["topologies"][host_topo.name()];
+    src_topo.set(host_topo);
+}
```
JustinPrivitera, line 35: Why copy the created pieces into the `src` node? Why not just use `src` to hold the mesh and pass it to `blueprint::mesh::examples::braid` directly?

src/libs/conduit/conduit_benchmark.hpp
siramok: 1) For this PR, `ExecConfig` is useless. For the follow-up PR, it's still not strictly necessary, but mainly exists for the sake of organization and ease of plumbing. I have no strong preference whether to have it be introduced here or there. I could make a more serious pass at only leaving in necessary bits for this PR if you prefer.

2) Whether `ExecConfig` gets introduced here or in the next PR, I like the idea of adding column annotations.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -0,0 +1,176 @@
...
+        BENCHMARK_NUM_ITERATIONS = static_cast<index_t>(atoll(argv[2]));
+    }
+    if (argc >= 4)
+    {
+        BENCHMARK_DIM_SIZES.clear();
+        for (int i = 3; i < argc; i++)
+        {
+            BENCHMARK_DIM_SIZES.push_back(static_cast<index_t>(atoll(argv[i])));
+        }
+    }
```
JustinPrivitera, line 148: I didn't see this when commenting above.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+    // TODO: Look at Caliper options related to OpenMP/GPU profiling
+    const std::string timestamp = benchmark::get_timestamp();
+    Node cali_opts;
+    cali_opts["config"] = "hatchet-region-profile(output=" + timestamp + ".cali)";
```
JustinPrivitera, line 153: 😍

src/tests/blueprint/plot_benchmark_output.py
```diff
@@ -0,0 +1,435 @@
...
+    "uniform_to_rectilinear",
+]
+
+TOPOLOGY_CONVERSIONS = [
+    "topology_uniform_to_rectilinear",
+    "topology_uniform_to_structured",
+    "topology_uniform_to_unstructured",
+    "topology_rectilinear_to_structured",
+    "topology_rectilinear_to_unstructured",
+    "topology_structured_to_unstructured",
```
JustinPrivitera, line 30: why do topo conversions have a "topology_" prefix but coordset conversions have no prefix?

src/tests/blueprint/plot_benchmark_output.py
JustinPrivitera, line 1: Can you post images of the plots this script generates?

src/libs/conduit/conduit_benchmark.hpp
JustinPrivitera: Sorry, what I was asking was more along the lines of will we need `ExecConfig` in the future? It adds complexity but bundles things nicely, so the question is if you are happy with that tradeoff now and in the future. There are no wrong answers!

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: It's not necessary in the context of this PR.

In the context of the next PR, `blueprint::mesh::examples::braid` allocates everything in host memory, but we sometimes want the `src` fields to originate in device memory at the start of a benchmark. The thought is that the above approach would allow us to easily control whether `src` data lives in host (do the `.set` as-is) or device memory (first set the device allocator, then call `.set` on the exact fields we need for a transform).

In that scenario, is it possible to move existing fields of a node from host to device memory in-place (keeping the same node structure)? I'd be curious to get your feedback: https://github.com/llnl/conduit/blob/a658d615fb92998a38fbf53bc80fd8209c62f747/src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp#L42

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: I left documentation links here: https://github.com/llnl/conduit/pull/1358#issuecomment-4849799733

But essentially, it seems like caliper can be built with extra options that enable us to collect openmp/cuda/hip-specific info.

src/tests/blueprint/plot_benchmark_output.py
siramok: Good point, will change.

src/tests/blueprint/plot_benchmark_output.py
siramok: Yes, it just now occurred to me that GH comments allow attaching images.

siramok: Here's a dump of everything that gets generated for a pre-device execution model .cali file (the post-DEM .cali files are more colorful, and include the breakdown plots of `use_with`, `forall`, and `sync`:
end_to_end_combined_coordset
end_to_end_combined_coordset_per_element
end_to_end_combined_topology
end_to_end_combined_topology_per_element
end_to_end_rectilinear_to_explicit
end_to_end_rectilinear_to_explicit_per_element
end_to_end_topology_rectilinear_to_structured
end_to_end_topology_rectilinear_to_structured_per_element
end_to_end_topology_rectilinear_to_unstructured
end_to_end_topology_rectilinear_to_unstructured_per_element
end_to_end_topology_structured_to_unstructured
end_to_end_topology_structured_to_unstructured_per_element
end_to_end_topology_uniform_to_rectilinear
end_to_end_topology_uniform_to_rectilinear_per_element
end_to_end_topology_uniform_to_structured
end_to_end_topology_uniform_to_structured_per_element
end_to_end_topology_uniform_to_unstructured
end_to_end_topology_uniform_to_unstructured_per_element
end_to_end_uniform_to_explicit
end_to_end_uniform_to_explicit_per_element
end_to_end_uniform_to_rectilinear
end_to_end_uniform_to_rectilinear_per_element

src/tests/blueprint/plot_benchmark_output.py
siramok: Fixed.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: Do you mean for the default sizes? Or for the real runs?

src/libs/conduit/conduit_benchmark.hpp
siramok: Need is a strong word. For this PR, it adds nothing of value. I do like its utility for the next PR, but I can re-evaluate whether there's a better way once we're working on it.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
JustinPrivitera: I see. I think it will be cleaner to just

```c++
    conduit::execution::init_device_memory_handlers();
    Node mesh;
    blueprint::mesh::examples::braid(src_type,
                                     config.dim_size,
                                     config.dim_size,
                                     config.dim_size,
                                     mesh);
    conduit::index_t d_alloc = conduit::execution::get_device_allocator_id();
    src.set_allocator(d_alloc);
    src.set(mesh);
```
instead of grabbing the pieces for coordsets and topos each?

siramok: It does look cleaner. Will that copy everything to device memory, including trivial fields and metadata (e.g. strings, or other things that may need to stay on host)? Maybe for our purposes it doesn't matter.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
JustinPrivitera: Oh sorry, the heart eyes were for `cali_opts["config"] = "hatchet-region-profile(output=" + timestamp + ".cali)";` you discovering the magic options to create hatchet input.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
JustinPrivitera: I did not see that you could append a larger value to this list down below, my bad.

src/libs/conduit/conduit_benchmark.hpp
JustinPrivitera: ok sounds good

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
JustinPrivitera: I think everything gets copied over. You might be right that we don't want that. The tree structure will live on host but the string leaves will not, which is going to cause problems.

AI came up with this for me:

```c++
  #include "conduit.hpp"
  #include "conduit_blueprint_mesh_examples.hpp"

  using namespace conduit;

  static void copy_numeric_arrays_to_device(const Node &src,
                                            Node &dst,
                                            index_t device_alloc)
  {
      if(src.dtype().is_object())
      {
          NodeConstIterator itr = src.children();
          while(itr.has_next())
          {
              const Node &src_child = itr.next();
              copy_numeric_arrays_to_device(src_child, dst[itr.name()], device_alloc);
          }
      }
      else if(src.dtype().is_list())
      {
          NodeConstIterator itr = src.children();
          while(itr.has_next())
          {
              const Node &src_child = itr.next();
              copy_numeric_arrays_to_device(src_child, dst.append(), device_alloc);
          }
      }
      else if(src.dtype().is_number() && src.dtype().number_of_elements() > 1)
      {
          dst.set_allocator(device_alloc);
          dst.set(src); // requires a device-aware memcpy handler
      }
      else
      {
          dst.set(src); // keep strings, dims, small metadata on host
      }
  }

  Node braid_host;
  blueprint::mesh::examples::braid("hexs", 32, 32, 32, braid_host);

  Node braid_dev;
  copy_numeric_arrays_to_device(braid_host,
                                braid_dev,
                                execution::get_device_allocator_id());
```

This could be useful for copying arbitrary meshes to device in the way that we want - only leaf data arrays. We will soon have field transforms so being able to move everything over simply could be useful, rather than moving each piece by hand.

siramok: Makes sense. This will have to be on the other PR, so I'll just copy this to have it there for inspiration.

siramok: For this PR, I changed it to just be the braid example.

JustinPrivitera (approved):
We just need to fix CI now

siramok:
CI looks fixed.

PR #1645, opened 2026-07-10
Change implementation for determining MPI tag limits

BradWhitlock:
The previous implementation for determining MPI tag upper limits failed in a `MPI_Waitall` to self on one platform. The implementation was changed to try and first query the value from the communicator attributes and failing that determine the number using messages, similar to before but with a `MPI_Sendrecv` to self.

* Changed implementation
* Tags above the limit are wrapped now with mod instead of using clamping
* Improved tests


cyrush (approved):
Looks good, thanks for the updated logic!

BradWhitlock:
Hey @cyrush , I see that you approved this PR already so I'm merging it.

Issue #1644, opened 2026-07-09
to_unstructured removes coordset name from topology


JustinPrivitera:
input mesh:

```
coordsets: 
  coords: 
    type: "uniform"
    dims: 
      i: 3
      j: 3
      k: 3
    origin: 
      x: -10.0
      y: -10.0
      z: -10.0
    spacing: 
      dx: 10.0
      dy: 10.0
      dz: 10.0
topologies: 
  mesh: 
    type: "uniform"
    coordset: "coords"
fields: 
  field: 
    association: "element"
    topology: "mesh"
    volume_dependent: "false"
    values: [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0]
```
after to_unstructured

we are missing coordset for topology
```
coordsets: 
  coords: 
    type: "explicit"
    values: 
      x: [-10.0, 0.0, 10.0, -10.0, 0.0, 10.0, -10.0, 0.0, 10.0, -10.0, 0.0, 10.0, -10.0, 0.0, 10.0, -10.0, 0.0, 10.0, -10.0, 0.0, 10.0, -10.0, 0.0, 10.0, -10.0, 0.0, 10.0]
      y: [-10.0, -10.0, -10.0, 0.0, 0.0, 0.0, 10.0, 10.0, 10.0, -10.0, -10.0, -10.0, 0.0, 0.0, 0.0, 10.0, 10.0, 10.0, -10.0, -10.0, -10.0, 0.0, 0.0, 0.0, 10.0, 10.0, 10.0]
      z: [-10.0, -10.0, -10.0, -10.0, -10.0, -10.0, -10.0, -10.0, -10.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0, 10.0]
topologies: 
  mesh: 
    type: "unstructured"
    coordset: ""
    elements: 
      shape: "hex"
[...]

PR #1646, opened 2026-07-10
Make mesh transform outputs inherit the source's coordset name if none was explicitly set

siramok:
Addresses #1644.

Justin and I discovered that performing mesh transforms using empty output `Node`s was resulting in no coordset name being set in the output topology. Existing behavior expects that explicitly setting a coordset name on an output `Node` before performing a mesh transform will preserve that coordset name.

This PR preserves existing behavior when setting an explicit coordset name, but now makes the output `Node`s inherit the input `Node`'s coordset name if one was not explicitly set. The existing topology transform tests were extended to exercise this new behavior.


siramok:
We decided that existing behavior was important to keep.

~If you're curious to look at the initial commit, it has a version of this PR that tries to both keep the existing behavior while also fixing the issue. Although I'm not sure if 1) the existing behavior was ever intended or 2) whether it is important to preserve.~

JustinPrivitera:
@cyrush can you have a look at this? We discovered that the blueprint transforms can give wrong results if you call them with empty nodes.

This is the way things are set up to work:
```c++
    // CASE 1
    Node &new_mesh_coords = new_mesh["coordsets"][coordset_name];
    Node &new_mesh_topo = new_mesh["topologies"][topo_name];
    blueprint::mesh::topology::*****::to_*****(original_topo, new_mesh_topo, new_mesh_coords);
```
But the transforms don't work if you create brand new nodes for the topology and coordset:
```c++
    // CASE 2
    Node new_mesh_coords;
    Node new_mesh_topo;
    blueprint::mesh::topology::*****::to_*****(original_topo, new_mesh_topo, new_mesh_coords);
```

Because the topology transforms set the coordset name for the topology to be the name of the destination coordset node.
i.e.
```c++
new_mesh_topo["coordset"].set(new_mesh_coords.name());
```

In the first case up above, the name exists because we created it manually. In the second case, there is no name and so we are left with bad results. `blueprint::mesh::convert` invokes the transforms using the first case pattern, but I feel that because the transforms are in the API, they ought to work for the case where the destination nodes passed in have no names. I think we should fallback and use the coordset name stored in the source topology, but I am open to hearing your thoughts. Please let us know what you think about this problem and this proposed solution.

Thanks!

siramok:
Justin commented to me offline that it would be better to query `topo["coordset"]` instead of `coordset.name()` for the fall back, so I changed the helper to take that approach instead.

siramok:
When we spoke last week, we decided that we don't want to have a fallback if `cdest.name().empty()` (the fallback was to just use `topo["coordset"].name()`) and that we should instead throw a `CONDUIT_ERROR` in these cases. The thought was that this would preserve existing behavior and enforce calling these APIs in a way that would always contain a coordset name in the output.

I've made that change, but it was somewhat complicated to implement, because it revealed several places where this wasn't actually the case:

- Both `generate_corners` and `generate_sides` needed a fix.
- `Partitioner` had several instances.
- The mesh utils had two instances.

It also revealed a bug in the `topology_transform_dtypes` test that's unrelated to the changes made in this PR. The test was creating `jtopology`/`jcoordset` as plain `Node` copies instead of references (there was an old `FIXME(JRC)` that seemed to be aware of this). It seems that because they were `Node` copies and not references, `jmesh` itself was always empty at the end of the loop, and so the two `verify_node_data` assertions were basically checking nothing (and passing). Now that `jmesh` isn't empty, the test fails due to what appear to be deeper issues.

siramok:
~Seems that I may have missed some MPI-specific instances.~

I was, fixed.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok, line 1225: Pointing this out because it seems to be possibly relevant to the broken dtype test.

JRC was leaving us the breadcrumbs all along. :thinking:

siramok:
We discussed this again and it seems like the internal APIs that make use of `to_unstructured` are using it in a way that don't actually use the coordset info that gets created. We think adding an overload for `to_unstructured` that just takes a coordset name directly as a string will be an improvement, because then these internal APIs can be made slightly more efficient and also avoid the new pattern that this PR forces. We also agreed that it would be good to continue erroring in the case that an empty string is passed.

src/tests/blueprint/t_blueprint_mesh_transform.cpp
```diff
@@ -381,19 +381,26 @@ TEST(conduit_blueprint_mesh_transform, topology_transform_dtypes)
                     Node &itopology = imesh["topologies"].child(0);
                     Node &icoordset = imesh["coordsets"].child(0);
 
+                    set_node_data(imesh, bputils::INT_DTYPES[ii]);
+                    set_node_data(imesh, bputils::FLOAT_DTYPES[fi]);
+
```
JustinPrivitera: how come this got moved?

src/tests/blueprint/t_blueprint_mesh_transform.cpp
```diff
@@ -381,19 +381,26 @@ TEST(conduit_blueprint_mesh_transform, topology_transform_dtypes)
...
                     // FIXME(JRC): I think these should be references! i.e. Node &jtopology
                     // Changing them in this way causes this test to fail.
+                    //
+                    // The above comment was correct: jtopology/jcoordset do need to be
+                    // references for the coordset name to come through correctly.
+                    // However, doing so reveals that 'find_widest_dtype' doesn't
+                    // seem to propagate the source's dtype into new fields. Fixing
+                    // that is probably a big job, so I've disabled the
+                    // 'verify_node_data' calls for now.
```
JustinPrivitera, line 392: Can you add a ticket referencing this problem with a link to the code?

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -48,7 +48,8 @@ benchmark_coordset_transform(const char *name,
     // Perform a transform on the source Node
     auto run = [=](const Node &src) {
         Node dst;
-        transform(src["coordsets"].child(0), dst);
+        Node &coords_dst = dst["coordsets/coords"];
+        transform(src["coordsets"].child(0), coords_dst);
     };
```
JustinPrivitera: I think we should transition to using `mesh::convert` instead of calling the individual transforms. In practice we will be reasoning about meshes, so testing topo transforms and coordset transforms separately is not as important. If we go to `mesh::convert` then we won't have to worry about this business with named coordsets.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -73,7 +74,9 @@ benchmark_topology_transform(const char *name,
 
     // Perform a transform on the source Node
     auto run = [=](const Node &src) {
-        Node topo_dst, coords_dst;
+        Node dst;
+        Node &topo_dst = dst["topologies/mesh"];
+        Node &coords_dst = dst["coordsets/coords"];
```
JustinPrivitera: same comment here about using `convert`

JustinPrivitera (changes requested):
Additionally, can you add another ticket about investigating if we should have another signature for all of the topology transformations that, instead of taking a destination coordset, takes a coordset name and does not bother creating a coordset? Along with instructions to audit the topo transform uses around the code to determine if they should use the simplified signature.

src/tests/blueprint/t_blueprint_mesh_transform.cpp
siramok: I was attempting to fix the test, but moving these around didn't make a difference.

src/tests/blueprint/t_blueprint_mesh_transform.cpp
siramok: Yes I can work on that.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: That makes sense to me. Do you want that to be a new ticket also?

JustinPrivitera: If this is already handled in the benchmarking PR then no need to do that here or open a new ticket.

src/tests/blueprint/t_blueprint_mesh_transform.cpp
siramok: This has been done: #1654.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: Yes it is handled there, I'll mark this as resolved.

src/tests/blueprint/t_blueprint_mesh_transform.cpp
siramok: Moved it back.

siramok:
In reference to your most recent comment, I created #1655 to track it.

JustinPrivitera:
What did we decide for this PR? We were going to put most ideas into tickets and then clean up a few things here, do you remember what they are @siramok?

siramok:
There are 3 separate issues here, but the first two might be fixable by the same solution/PR:
1. The topology transforms currently assume that the output node is getting passed in the "correct" way (i.e., by reference to a named coordset node). The problems with that are:
    1. The correct way is non-obvious (maybe a non-issue if we expect people to just use `mesh::convert`)
    2. The solution presented in this PR fixes the problem by making the incorrect way throw an error, which forces calling code to use an ugly/unintuitive pattern to avoid the error (there's an argument to be made that this is still the right thing to do while we figure out how to make the pattern nicer)

2. Forcing the ugly pattern revealed that we have internal `_mesh_partition` APIs which only care about the transformed topology, but `to_<transform>` always generates a coordset (even if the calling code doesn't end up using it). Exposing a way to ask for the topology conversion only would effectively be an optimization. The problems with that are:
    1. We discussed overloading the topology transforms to do this, e.g. by passing a coordset name, but we won't know if this is a viable fix until we try it first (similar to the current state of this PR)
    2. Fixing this doesn't solve the fundamental problem with the first issue (and ideally both could be fixed at once with a clever enough solution)

3. Exploring the above issue revealed that one of the mesh transform tests has been broken all along.
    1. Fixing it seems unrelated to the above 2 issues, and will require its own deep dive that may reveal more issues

siramok:
> What did we decide for this PR?

If you'd like, since I've already spent time investigating these things, I can make another attempt to solve the first two issues within this PR. It got put on the backburner while I handled other stuff, but I've got time again.

src/tests/blueprint/t_blueprint_mesh_transform.cpp
```diff
@@ -383,17 +383,24 @@ TEST(conduit_blueprint_mesh_transform, topology_transform_dtypes)
 
                     // FIXME(JRC): I think these should be references! i.e. Node &jtopology
                     // Changing them in this way causes this test to fail.
+
+                    // The above comment was correct: jtopology/jcoordset do need to be
+                    // references for the coordset name to come through correctly.
+                    // However, doing so reveals that 'find_widest_dtype' doesn't
+                    // seem to propagate the source's dtype into new fields. Fixing
+                    // that is probably a big job, so I've disabled the
+                    // 'verify_node_data' calls for now.
```
JustinPrivitera, line 392: let's leave a comment with the URL to the issue we opened: https://github.com/llnl/conduit/issues/1654

JustinPrivitera:
I think since we have the ticket to investigate making a better way to call the transforms that does not make a coordset, we can leave the ugly pattern as is for now with regards to (1).

You are welcome to attack that but maybe we can do that in a follow-on PR for (1) and (2).

I think all we need to do for (3) is add a note linking to the ticket we opened.

I so I think this is good to go after you do that.

src/tests/blueprint/t_blueprint_mesh_transform.cpp
siramok: I have added another line to this comment linking to the issue.

siramok:
I verified that the tests pass (on tuo anyways) after the merge and have added the comment you asked for. Once CI passes, I think this is ready to merge.

siramok:
We said that this is good to go as-is, but then I thought about it more. If there exists a user whose workflow depends on the existing "bad" behavior that this PR changes, they'll eventually update conduit and discover that their previously working code now throws an error at runtime. The current error message says:
```c++
CONDUIT_ERROR("The destination coordset Node has no name. Pass a named Node "
                      "(e.g. output[\"coordsets/mycoordset\"]) instead of an anonymous Node.")
 ```

Would it be helpful to add something to the effect of `"If you're attempting to convert a mesh, consider using the newer mesh::convert API instead which will do the correct thing on your behalf"`?

JustinPrivitera (approved):

siramok:
We decided that if users have issues, we can cross that bridge when we get there.

PR #1647, opened 2026-07-10
Device Support: Make Sync Strategy an Enum Class

JustinPrivitera:
Before we get too far, I wanted to clean up sync strategy a bit.

### DIFF
--- a/src/libs/conduit/conduit_data_array.hpp
+++ b/src/libs/conduit/conduit_data_array.hpp
@@ -236,7 +236,7 @@ public:
 
     void                                assume();
 
-    void                                data_movement(const std::string &strategy);
+    void                                data_movement(conduit::execution::SyncStrategy strategy);
 
     conduit::execution::ExecutionPolicy active_space();
 
--- a/src/libs/conduit/conduit_data_array.hpp
+++ b/src/libs/conduit/conduit_data_array.hpp
@@ -236,7 +236,7 @@ public:
 
     void                                assume();
 
-    void                                data_movement(conduit::execution::SyncStrategy strategy);
+    void                                data_movement(const conduit::execution::SyncStrategy strategy);
 
     conduit::execution::ExecutionPolicy active_space();
 
--- a/src/libs/conduit/conduit_execution.cpp
+++ b/src/libs/conduit/conduit_execution.cpp
@@ -36,39 +36,6 @@ namespace execution
 {
 
 //-----------------------------------------------------------------------------
-const char *
-sync_strategy_to_string(SyncStrategy strategy)
-{
-    switch(strategy)
-    {
-        case SyncStrategy::Sync:
-            return "sync";
-        case SyncStrategy::Assume:
-            return "assume";
-    }
-
-    return "<invalid sync strategy>";
-}
-
-//-----------------------------------------------------------------------------
-SyncStrategy
-sync_strategy_from_string(const std::string &strategy)
-{
-    if(strategy == "sync")
-    {
-        return SyncStrategy::Sync;
-    }
-
-    if(strategy == "assume")
-    {
-        return SyncStrategy::Assume;
-    }
-
-    CONDUIT_ERROR("ExecutionOptions: invalid sync_strategy option.");
-    return SyncStrategy::Assume;
-}
-
-//-----------------------------------------------------------------------------
 /// policy constructor helpers
 //-----------------------------------------------------------------------------
 ExecutionPolicy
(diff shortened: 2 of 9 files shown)


src/libs/conduit/conduit_execution.cpp
```diff
@@ -584,7 +576,7 @@ class ExecutionOptions
     }
 
     //------------------------------------------------------------------------
-    static const std::string& get_sync_strategy()
+    static SyncStrategy get_sync_strategy()
```
siramok, line 579: Does it no longer need to be `const`?

src/libs/conduit/conduit_data_array.hpp
```diff
@@ -236,7 +236,7 @@ class CONDUIT_API DataArray
 
     void                                assume();
 
-    void                                data_movement(const std::string &strategy);
+    void                                data_movement(conduit::execution::SyncStrategy strategy);
```
siramok: Does it no longer need to be `const`?

src/libs/conduit/conduit_execution.cpp
```diff
@@ -688,7 +680,7 @@ get_output_allocator_id()
 }
 
 //-----------------------------------------------------------------------------
-const std::string&
+SyncStrategy
```
siramok, line 683: Same question about `const`.

src/libs/conduit/conduit_execution.cpp
```diff
@@ -708,6 +700,45 @@ get_host_allocator_id()
...
+    else if (SyncStrategy::Assume == strategy)
+    {
+        return "assume";
+    }
+    else
+    {
+        CONDUIT_ERROR("Unknown data movement strategy: "
+                      << conduit::execution::sync_strategy_to_string(strategy)
+                      << " (" << static_cast<int>(strategy) << ").");
+        return "";
```
siramok, line 720: If we're passed an invalid sync strategy, is it reasonable to return an answer that's also invalid (`""`)? I believe that `CONDUIT_ERROR` will cause the program to exit before that actually happens, and I assume that this is only here because the compiler requires you to return something in all possible code paths.

A pattern along these lines could be an alternative:
```c++
std::string result = "assume";

if (SyncStrategy::Sync == strategy)
{
    result = "sync";
}
else if (SyncStrategy::Assume != strategy)
{
        CONDUIT_ERROR("Unknown data movement strategy: "
                      << conduit::execution::sync_strategy_to_string(strategy)
                      << " (" << static_cast<int>(strategy) << ").");
}

return result;
```

(I don't feel strongly that this should be changed, just thinking out loud in case it's helpful. 😄 )

src/libs/conduit/conduit_execution.cpp
```diff
@@ -708,6 +700,45 @@ get_host_allocator_id()
...
+//-----------------------------------------------------------------------------
+SyncStrategy
+sync_strategy_from_string(const std::string &strategy)
+{
+    if (strategy == "sync")
+    {
+        return SyncStrategy::Sync;
+    }
+
+    if (strategy == "assume")
```
siramok: I noticed that the other function uses `else if`, but this one does two `if`s back to back. Since they both live side-by-side and are essentially performing the same pattern, maybe it would be nice to pick your preferred approach and use that in both places?

src/libs/conduit/conduit_execution.cpp
```diff
@@ -708,6 +700,45 @@ get_host_allocator_id()
...
+    {
+        return SyncStrategy::Assume;
+    }
+
+    CONDUIT_ERROR("ExecutionOptions: invalid sync_strategy option.");
+    return SyncStrategy::Assume;
```
siramok: Since we're switching to an enum, does it make sense to have `SyncStrategy::Invalid` for this case? Even if the `CONDUIT_ERROR` never allows the program to actually get this far, perhaps it better signals the intent?

Alternatively, in the above function `""` is returned in the invalid case, but returning `"assume"` there would be consistent with what gets returned in the invalid case here.

src/libs/conduit/conduit_execution.hpp
```diff
@@ -40,6 +40,25 @@ namespace conduit
 namespace execution
 {
 
+//-----------------------------------------------------------------------------
+/// Internal sync/data movement strategy for execution helpers.
+//-----------------------------------------------------------------------------
+enum class SyncStrategy
```
siramok, line 46: Beautiful. 🚀

siramok (commented):
I think this is a good change! I just had some minor feedback for you to consider.

siramok:
I forgot that I could drag to include more of the context in my comments, sorry about that!

src/libs/conduit/conduit_data_array.hpp
JustinPrivitera: good call const is best.

src/libs/conduit/conduit_execution.cpp
JustinPrivitera: It is just returning an enum by value, so making the return type const does nothing.

src/libs/conduit/conduit_execution.cpp
JustinPrivitera: same answer :)

src/libs/conduit/conduit_execution.cpp
JustinPrivitera: The compiler likes when we return things along all code paths, yes. CONDUIT_ERROR will halt execution, as you noted, so it really is just about making the compiler happy.

I think returning an invalid sync strategy is actually a good idea because this will likely force things to go wrong down the road if for some reason the error was removed. We want future developers to make sure they handle all the cases instead of silently giving them a workable solution.

That's my opinion, in any case.

src/libs/conduit/conduit_execution.cpp
JustinPrivitera: Thanks for catching. I blame AI 🙃. This is why it is good to have a second set of eyes.

src/libs/conduit/conduit_execution.cpp
JustinPrivitera: I have to return a Sync Strategy. I think adding an invalid strategy is probably more complicated than we need this to be. I guess I will contradict what I said before a little bit and say that I think it is ok if we return a useful value here, as long as we have the Conduit error.

PR #1648, opened 2026-07-16
Device Support: Expose explicit policy choices in global execution options

JustinPrivitera:


src/libs/conduit/conduit_execution.cpp
```diff
@@ -326,7 +360,7 @@ class ExecutionOptions
     // "sync" or "assume"
     static SyncStrategy sync_strategy;
 
-    // "host" or "device"
+    // fallback execution policy if input-based resolution has no input node
```
siramok: I'm in favor of replacing the old comment with something more descriptive, but I think what's here would benefit from plainer language. Specifically, `fallback execution policy` makes perfect sense but it's not immediately clear when you'd encounter `if input-based resolution has no input node`.

My assumption would be that `fallback_location` comes into play when an incoming options `Node` specified "input" for the `execution_location`, but didn't posses a child node named "input" containing a valid policy name?

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -362,6 +592,96 @@ TEST(conduit_execution, execution_settings)
         execution::reset_execution_options();
     }
```
siramok, line 588: This block is causing the test to fail for builds that don't have CUDA or HIP enabled:
```c++
[ RUN      ] conduit_execution.execution_settings
unknown file: Failure
C++ exception with description "
file: /workspaces/conduit/src/libs/conduit/conduit_execution.cpp
line: 74
message:
Conduit was built with neither CUDA nor HIP.
" thrown in the test body.

[  FAILED  ] conduit_execution.execution_settings (0 ms)
```
It's because `fallback_location` is being unconditionally set to "device", which previously would end up returning `device_allocator` (equal to -1), but now executes `resolve_execution_option_policy` which executes `else if (policy_name == "device")   return ExecutionPolicy::device();` and throws an error.

I think it would be fine to just guard the entire block with:
```c++
#if defined(CONDUIT_USE_DEVICE)
...
#endif // defined(CONDUIT_USE_DEVICE)
```
Or since it seems like a reasonable error, you could make the test expect it:
```c++
#if defined(CONDUIT_USE_DEVICE)
        index_t fallback_alloc_id = execution::get_output_allocator_id();
        EXPECT_EQ(fallback_alloc_id, DEVICE_ALLOC_ID);
#else // !defined(CONDUIT_USE_DEVICE)
        EXPECT_THROW(execution::get_output_allocator_id(), conduit::Error);
#endif // !defined(CONDUIT_USE_DEVICE)
```

siramok (changes requested):

src/tests/conduit/t_conduit_execution.cpp
siramok: I guess github doesn't feel like showing the full context I marked, but the block I'm referring to is on line 569 starting with `// test input output allocator & w/ device fallback`.

src/libs/conduit/conduit_execution.cpp
JustinPrivitera: added new comment

src/tests/conduit/t_conduit_execution.cpp
JustinPrivitera: I encased the whole test block in the ifdef device block

siramok:
I'm building again locally to verify that the failing test is fixed, then it's good to merge.

siramok:
It's good to go!

siramok (approved):

Issue #1637, opened 2026-06-26
Device Support: Port Blueprint Transforms to the GPU


JustinPrivitera:
We will list transforms that we are porting and check them off here.

### Coordset Transforms:
From/To | uniform | rectilinear | explicit
-- | -- | -- | --
**uniform** | n/a | ✅ #1358 | ✅ #1635
**rectilinear** | no xform | n/a | ✅ #1635
**explicit** | no xform | no xform | n/a

- [ ] what about `generate_strip`? Investigate ❌ TODO

#### Transforms between topology types:
From/To | points | uniform | rectilinear | structured | unstructured
-- | -- | -- | -- | -- | --
**points** | n/a | no xform | no xform | no xform | no xform
**uniform** | no xform | n/a | trivial xform | trivial xform | ✅ #1640
**rectilinear** | no xform | no xform | n/a | trivial xform | ✅ #1640
**structured** | no xform | no xform | no xform | n/a | ✅ #1640
**unstructured** | no xform | no xform | no xform | no xform | n/a

- [ ] what about `generate_strip`? Investigate ❌ TODO

#### Unstructured Transforms/Generations:
- [x] `generate_centroids` - ✅ #1673
- [ ] `to_polytopal` - ❌ TODO
- [ ] `generate_points` - ❌ TODO
- [ ] `generate_lines` - ❌ TODO
- [ ] `generate_faces` - ❌ TODO
- [ ] `generate_sides` - ❌ TODO
- [ ] `generate_corners` - ❌ TODO
- [ ] `generate_offsets` - ❌ TODO
- [ ] `generate_offsets_inline` - ❌ TODO

TODO additional transforms will be added to this list, but this is a good start for now.

PR #1653, opened 2026-07-22
Add the rest of the mesh transform benchmarks and improve the benchmarking infrastructure

siramok:
This PR adds benchmarks for the rest of the APIs mentioned in #1637. It also makes improvements to the benchmarking infrastructure itself, backporting changes from the device-execution side as well as implementing new ideas I've been meaning to explore. This PR targets the `develop` branch so that its changes can be later merged into the device-support branch too.

Output nodes are now only created once and are `.reset()` instead of getting re-allocated many times. This seems to be a big performance improvement, as the overall benchmark time with all of the new benchmarks is now about the same as what it took to perform just the coordset and topology benchmarks before.

Everything (that can) is now using the `mesh::convert` API. AI helped me find other simplifications + improvements with respect to naming and reducing boilerplate.

I still need to review the plotting script changes, as they are purely AI-generated without me making sure they are sensible.

### DIFF
--- a/src/libs/conduit/conduit_benchmark.hpp
+++ b/src/libs/conduit/conduit_benchmark.hpp
@@ -19,6 +19,7 @@
 
 #include "conduit_annotations.hpp"
 #include "conduit_data_type.hpp"
+#include "conduit_node.hpp"
 
 // This is not needed after device support is added back
 #if defined(CONDUIT_USE_OPENMP)
@@ -64,32 +65,40 @@ get_exec_configs()
 {
     // In the pre-device execution model world, we don't have the concept of
     // host vs device execution
-    //                                src       exec      dest
-    std::vector<ExecConfig> configs{{"host", "host", "host"}};
+    std::vector<ExecConfig> configs{
+        //source location, execution location, output location
+        {"host",           "host",             "host"},
+    };
     return configs;
 }
 
 //-----------------------------------------------------------------------------
 template <typename SetupFn, typename RunFn>
 void
-exec(const char *name,
+exec(const std::string &name,
      SetupFn &&setup,
      RunFn &&run,
      const index_t warmup,
      const index_t iterations,
      const std::vector<index_t> &dim_sizes)
 {
+    const std::vector<ExecConfig> configs = get_exec_configs();
+
     // Benchmark each data size
     for (const auto &dim_size : dim_sizes)
     {
         // Benchmark each possible configuration
-        for (auto &config : get_exec_configs())
+        for (ExecConfig config : configs)
         {
             config.dim_size = dim_size;
 
             // Build the input once, outside the timed regions
             Node input;
             setup(config, input);
+
+            // Reused across every iteration. It gets reset() immediately
+            // before each call, so run() always sees it empty.
+            Node dst;
 
             // Execute `run` `warmup` times
             {
@@ -98,7 +107,8 @@ exec(const char *name,
                 CONDUIT_ANNOTATE_MARK_SCOPE("warmup");
                 for (index_t i = 0; i < warmup; i++)
                 {
-                    run(input);
+                    dst.reset();
+                    run(input, dst);
                 }
             }
 
@@ -107,7 +117,7 @@ exec(const char *name,
                 // This scope name is used to identify specific benchmarks in
                 // the Caliper output and identify their attributes
                 // (dim size, policy, etc.)
-                const std::string scope_name = std::string(name)
+                const std::string scope_name = name
                     + "_dim-"  + std::to_string(dim_size)
                     + "_src-"  + config.src_location
                     + "_exec-" + config.exec_location
@@ -119,7 +129,8 @@ exec(const char *name,
                 CONDUIT_ANNOTATE_MARK_SCOPE(scope_name.c_str());
                 for (index_t i = 0; i < iterations; i++)
                 {
-                    run(input);
+                    dst.reset();
+                    run(input, dst);
                 }
             }
         }
--- a/src/libs/conduit/conduit_benchmark.hpp
+++ b/src/libs/conduit/conduit_benchmark.hpp
@@ -88,7 +88,7 @@ exec(const std::string &name,
     for (const auto &dim_size : dim_sizes)
     {
         // Benchmark each possible configuration
-        for (ExecConfig &config : configs)
+        for (ExecConfig config : configs)
         {
             config.dim_size = dim_size;
 
--- a/src/libs/conduit/conduit_benchmark.hpp
+++ b/src/libs/conduit/conduit_benchmark.hpp
@@ -19,7 +19,6 @@
 
 #include "conduit_annotations.hpp"
 #include "conduit_data_type.hpp"
-#include "conduit_node.hpp"
 
 // This is not needed after device support is added back
 #if defined(CONDUIT_USE_OPENMP)
--- a/src/libs/conduit/conduit_benchmark.hpp
+++ b/src/libs/conduit/conduit_benchmark.hpp
@@ -38,6 +38,17 @@ namespace benchmark
 {
 
 //-----------------------------------------------------------------------------
+inline
+std::string
+get_timestamp()
+{
+    std::time_t t = std::time(nullptr);
+    std::ostringstream oss;
+    oss << std::put_time(std::localtime(&t), "%Y%m%d_%H%M%S");
+    return oss.str();
+}
+
+//-----------------------------------------------------------------------------
 struct ExecConfig
 {
     std::string src_location;
--- a/src/libs/conduit/conduit_benchmark.hpp
+++ b/src/libs/conduit/conduit_benchmark.hpp
@@ -86,9 +86,13 @@ exec(const std::string &name,
     // Capture input/output data sizes to include in the scope name below.
     // This is a function of (name, npts) and never changes between
     // iterations.
-    run(input, output);
-    const std::string sizes = size_info(input, output);
-    output.reset();
+    std::string sizes;
+    {
+        CONDUIT_ANNOTATE_MARK_SCOPE("size_probe");
+        run(input, output);
+        sizes = size_info(input, output);
+        output.reset();
+    }
 
     // Execute `run` `warmup` times
     {
--- a/src/libs/conduit/conduit_benchmark.hpp
+++ b/src/libs/conduit/conduit_benchmark.hpp
@@ -55,7 +55,6 @@ struct ExecConfig
     std::string src_location;
     std::string exec_location;
     std::string output_location;
-    index_t dim_size = 1;
 };
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/libs/conduit/conduit_benchmark.hpp
```diff
@@ -64,41 +65,50 @@ get_exec_configs()
...
      const std::vector<index_t> &dim_sizes)
 {
+    const std::vector<ExecConfig> configs = get_exec_configs();
+
     // Benchmark each data size
     for (const auto &dim_size : dim_sizes)
     {
         // Benchmark each possible configuration
-        for (auto &config : get_exec_configs())
+        for (ExecConfig config : configs)
```
JustinPrivitera: why not
```suggestion
        for (ExecConfig &config : configs)
```
?

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -46,90 +92,186 @@ benchmark_coordset_transform(const char *name,
...
+        blueprint::mesh::topology::unstructured::generate_offsets_inline(topo);
+    };
+
+    run_benchmark(name, setup, run);
 }
 
 //-----------------------------------------------------------------------------
 TEST(blueprint_mesh_transform_benchmark, coordset_transforms)
 {
+    // This is not an exhaustive benchmark of the coordset transforms;
```
JustinPrivitera: do we need this anymore now that we are using convert?

JustinPrivitera: we can combine topo + coordset benchmarking to just do mesh benchmarking

JustinPrivitera (changes requested):
A couple questions

src/libs/conduit/conduit_benchmark.hpp
siramok: A minor optimization, the old version creates/returns a new (but identical every time) `std::vector<ExecConfig>` for each dimension size that we benchmark. This new version just says lets create it once and reuse it, since the possibilities never change.

siramok: Ah, I misunderstood your comment to be asking about why make this change at all. I accepted your suggestion, I agree with it!

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: From the perspective of measuring performance I don't think it has an effect either way. Fewer tests is technically more efficient, since you have to pay a very minor gtest cost per test, but that's true for any test file. The reason to have separation is mostly for the sake of organization. If anything, I was thinking about further splitting things up so that you can see the gtest report about the total test time it takes per-benchmark. For example: having 3 tests for coordsets and 3 for topology that have one benchmark each.

src/libs/conduit/conduit_benchmark.hpp
siramok: Reverted for breaking CI.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: To clarify, when running the benchmark manually, it might be nice if each benchmark were its own test. That way you could get a nice summarized log of how long each conversion takes on its own.

JustinPrivitera: image

I think if we do full mesh conversions then we only lose the separation for to_unstructured between coordset and topo transform, but in all those cases coordset transform happens under the hood so topo transform is measuring coordset+topo transform time anyway.

For the other cases, the topo transform is trivial but the coordset transform is meaningful. Does that make sense?

src/libs/conduit/conduit_benchmark.hpp
JustinPrivitera: sorry :/ I wonder why that change broke things. It is probably not too hard to discover in a local checkout

siramok: Actually, CI seems to be broken for unrelated reasons. I'll try your change again locally and if it compiles fine, I'll add it back in.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: After discussing offline, my understanding is that full-mesh benchmarks at the `mesh::convert` level would be most useful to us. Since `mesh::convert` does both topology and coordset conversion, it'll be more representative of real-world improvements. Caliper measures each distinct piece anyways, so it should still be possible to see a breakdown of topo vs coordset conversion time.

Furthermore, since we're interested in benchmarking `mesh::convert` itself which does several things internally, it seems to make sense to benchmark trivial conversions (or ones that didn't make sense to port to the device execution model) as well since they may still be benefiting indirectly at a high level. At the very least, it would give us data on which conversions did and didn't improve after moving to DEM.

siramok:
My apologies, I changed pretty much everything about this PR after you last reviewed it (but for good reasons).

- Every benchmark now leverages `mesh::convert`; there are no more exceptions. This enabled me to make large simplifications. In the case of the two `generate_offset` benchmarks that used to be here, I determined that the other benchmarks will exercise them indirectly, so I removed them.
- The input and output vertex and element counts are now captured and included in the .cali files.
- The benchmarks have changed entirely, so the plotting script is also entirely different (with respect to what's in `develop`).
- I added support for benchmarking the `generate_*` APIs with different shape types, since 2D vs 3D vs poly meshes should theoretically touch different code paths. In the process, I discovered that the `generate_*` APIs are slow for small dim_size values regardless of shape type, but that the slowness compounds when using pyramids as the braid mesh type. My testing suggests pyramids scale worse than what would be explained by a larger vertex/element count per primitive alone. I also discovered that mixed and mixed_2D shape types trigger a segfault.
- This inspired me to look for further optimization opportunities within the benchmark infrastructure itself. The different braid meshes were previously being recreated for every benchmark that used them. I've restructured things so a single braid mesh gets reused across every benchmark that needs it, and then I grouped benchmarks by which mesh they depend on, so we only ever hold the meshes currently in use in memory. To do that required pulling stuff out of `::benchmark` and into the test, but the ability to do more tests in less time is worth it IMO.

siramok:
Here are examples of some plots:
mesh_scaling
mesh_scaling_per_element
mesh_conversion_heatmap_dim-8
mesh_conversion_heatmap_per_element_dim-8
generate_scaling
generate_scaling_per_element
generate_heatmap_dim-8
generate_heatmap_per_element_dim-8
generate_growth_heatmap_dim-8

siramok:
CI failure looks unrelated to the latest commit.

JustinPrivitera:
> Here are examples of some plots: 

These are cool; thanks for investing the time in making these.

src/tests/blueprint/plot_benchmark_output.py
```diff
@@ -2,7 +2,11 @@
 # Project developers. See top-level LICENSE AND COPYRIGHT files for dates and
 # other details. No copyright assignment is required to contribute to Conduit.
 
-import argparse
+# Plots .cali output from t_blueprint_mesh_transform_benchmark. Run with no
+# arguments to plot the most recently written .cali file in the current
+# directory, or pass a path to a specific one.
+
```
JustinPrivitera: Any chance you can add some usage disambiguation/examples here? Then it will be easier for future developers to examine results.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -12,153 +12,239 @@
...
-                                 "structured",
-                                 blueprint::mesh::topology::structured::to_unstructured);
+    
+    const std::vector<std::string> shapes = {
+        "quads",
+        "hexs",
+        "pyramids",    // particularly slow pre-device execution
+        // "mixed_2d", // TODO: investigate why this segfaults
+        // "mixed"     // TODO: investigate why this segfaults
+    };
```
JustinPrivitera, line 200: Why are we limited to these 3 (5) shape types?

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -12,153 +12,239 @@
...
+
+    // Building this list programatically makes it easy to benchmark with
```
JustinPrivitera:
```suggestion
    // Building this list programmatically makes it easy to benchmark with
```

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -12,153 +12,239 @@
...
+        configs.push_back({"generate_corners_" + shape,   shape, "generate_corners"});
+    }
+
+    run_benchmarks(configs);
 }
 
 //-----------------------------------------------------------------------------
-int main(int argc, char *argv[])
+int
+main(int argc, char *argv[])
```
JustinPrivitera, line 221: can you add a comment either here or at the top of the file showing how to run with all the options?

JustinPrivitera (changes requested):
Very nice work; I left a few requests for small things.

JustinPrivitera:
We will need to figure out the CI failure.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: There's not a technical reason that we're limited. Each shape that gets added to this list slows the benchmark down considerably, so what's here are the shapes that seemed to touch the most code-paths with the least amount of overlap.

`mixed_2d` and `mixed` are still here as a reminder that we should either investigate why we're segfaulting or make a separate issue about it to investigate later.

If you want the full list of possible shapes here, maybe just commented out, I think that's fine.

src/tests/blueprint/plot_benchmark_output.py
siramok: This has been added.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: This has been added.

siramok:
Just one remaining follow-up question to resolve.

siramok:
After the caliper tutorial, there are a few small low-hanging fruit things that I think are worth changing:
- ~Seems we don't need our own helper for giving .cali files unique names, caliper does it for us.~ Only true of the spot profile, and switching to it is out of scope for this PR.
    - ~Requires a corresponding change to the plotting script.~
- ~We can print runtime statistics to stdout while also getting a .cali file, I think printing something more than we currently do (nothing) would be helpful.~ This was doable.
- ~There are specific annotations for loops to get per-iteration info.~ Can't be done yet, requires adding these macros to the conduit annotations.

This would be a larger effort, but it didn't occur to me that too much caliper annotation could hurt performance. It would be interesting to quantify our existing caliper overhead. If it happened to be high, caliper supports a sampling profiler which would be much lighter-weight.

siramok:
I'm finished with investigating/making the changes that I wanted to make.

~CI is unfortunately failing due to mpich.~

JustinPrivitera (approved):

PR #1656, opened 2026-07-24
Port new benchmark PR (#1653) to device execution model

siramok:
I suggest focusing on merging #1653 first then #1657 second before reviewing this PR, because any requested changes there will have to be replicated here.

With that being said, this PR works well as-is and serves as a starting point for porting the `generate_*` APIs to DEM while I wait for #1653 to be merged.

TODO:

- Once #1648 is merged, we'll want to update the benchmark to exercise serial, openmp, and device within one benchmark run.
- ~Need to make sure that backend-separated plots are getting created again (I don't think they are).~
- ~Make it possible to set `host_only` per config and not globally for the whole test, because we'll only be porting 1 API at a time and will want to selectively allow device benchmarks.~

### DIFF
--- a/src/libs/conduit/conduit_benchmark.hpp
+++ b/src/libs/conduit/conduit_benchmark.hpp
@@ -147,6 +147,16 @@ exec(const std::string &name,
         }
     }
 
+    std::string backend_name;
+    if (config.exec_location == "host")
+    {
+        backend_name = execution::ExecutionPolicy::host().policy_name();
+    }
+    else // if (config.exec_location == "device")
+    {
+        backend_name = execution::ExecutionPolicy::device().policy_name();
+    }
+
     // Execute `run` `iterations` times
     {
         // This scope name is used to identify specific benchmarks in the
@@ -157,6 +167,8 @@ exec(const std::string &name,
             + "_src-"  + config.src_location
             + "_exec-" + config.exec_location
             + "_out-"  + config.output_location
+            + "_sync-" + config.sync_strategy
+            + "_backend-" + backend_name
 #if defined(CONDUIT_USE_OPENMP)
             + "_threads-" + std::to_string(omp_get_max_threads())
 #endif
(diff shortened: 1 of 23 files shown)


siramok:
Updated this PR to depend on #1657 to reduce limitations with testing as much as possible.

siramok:
It occurs to me that once #1653 is merged, we should consider merging it into #1358, completely overwriting `conduit_benchmark.hpp`, `t_blueprint_mesh_transform_benchmark.cpp`, and `plot_benchmark_output.py`.

This would technically be "wrong", but since this PR was developed in parallel with #1653, 1) ultimately nothing is lost and 2) the resulting diff will be much easier to review.

JustinPrivitera:
> It occurs to me that once #1653 is merged, we should consider merging it into #1358, completely overwriting `conduit_benchmark.hpp`, `t_blueprint_mesh_transform_benchmark.cpp`, and `plot_benchmark_output.py`.
> 
> This would technically be "wrong", but since this PR was developed in parallel with #1653, 1) ultimately nothing is lost and 2) the resulting diff will be much easier to review.

@siramok I am interested in ways to reduce the diff size

siramok:
@JustinPrivitera Diff reduction is complete.

src/libs/conduit/conduit_benchmark.hpp
```diff
@@ -52,46 +50,91 @@ struct ExecConfig
...
+        {"device",         "host",             "host",          "sync"},
+        {"device",         "host",             "device",        "sync"},
+        {"device",         "host",             "device",        "assume"},
+        {"device",         "device",           "host",          "sync"},
+        {"device",         "device",           "host",          "assume"},
+        {"device",         "device",           "device",        "sync"},
+#endif // defined(CONDUIT_USE_DEVICE)
     };
     return configs;
 }
```
siramok, line 123: To further help reduce the diff, this comment + function (~100 lines) are identical to what you already reviewed + approved here: https://github.com/llnl/conduit/blob/4c473cd4a8c3c4d40a69bcc73df078d2d6c8db5c/src/libs/conduit/conduit_benchmark.hpp#L62

JustinPrivitera (approved):

PR #1657, opened 2026-07-25
Fix memory leak in assume()

siramok:
This PR fixes a memory leak within the `assume()` implementation for both `DataAccessor` and `DataArray`.

After creating #1656, I was performing device-based test runs on my local PC to see how quick the new implementation was running relative to the old implementation. I noticed that I wasn't able to go above 256 dim test meshes on my GPU due to CUDA trying to allocate too much memory.

Maybe a 256 dim braid mesh is just too much for my 12GB of VRAM to handle, I thought. But by random chance, I simultaneously scaled the dim sizes down but increased the iteration count, and to my surprise, this also caused CUDA to ask for too much memory and crash. I did some digging to try to understand why device memory wasn't getting freed and believe I found the issue:

Previously, `assume()` was handing off a pointer to its execution location buffer without handing off ownership (i.e. letting the receiving node know that it was now holding allocated memory). This resulted in `.reset()` never actually freeing that memory. The fix is to set the necessary fields in the receiving node (in addition to the `data_ptr` itself) so that ownership of the buffer is effectively transferred.

I included a test that demonstrates the memory leak, which fails if the changes in this PR are reverted. I initially believed that `sync()` was leaking memory as well (which is why those tests are here), but splitting them into 4 separate tests proved that `sync()` was correctly cleaning up after itself.

I can now do hundreds of large mesh conversion tests without running out of device memory.

### DIFF
--- a/src/libs/conduit/conduit_data_accessor.cpp
+++ b/src/libs/conduit/conduit_data_accessor.cpp
@@ -597,6 +597,29 @@ DataAccessor<T>::sync()
                                                m_node_ptr->dtype().stride(),
                                                m_data,
                                                m_stride);
+
+        // By this point, m_data has been copied into m_node_ptr->data_ptr().
+        // We need to free our local copy of m_data, lest we leak memory.
+        if (m_do_i_own_it)
+        {
+            if (execution::DeviceMemory::is_device_ptr(m_data))
+            {
+                execution::DeviceMemory::deallocate(m_data);
+            }
+            else // if (!execution::DeviceMemory::is_device_ptr(m_data))
+            {
+                execution::HostMemory::deallocate(m_data);
+            }
+
+            m_do_i_own_it = false;
+            m_other_ptr = nullptr;
+            m_other_dtype = DataType::empty();
+        }
+
+        // Set m_data to the node's data and update offset and stride
+        m_data = m_node_ptr->data_ptr();
+        m_offset = dtype().offset();
+        m_stride = dtype().stride();
     }
 }
 
@@ -622,7 +645,16 @@ DataAccessor<T>::assume()
         // reset will deallocate the data the node points to
         m_node_ptr->reset();
         m_node_ptr->schema_ptr()->set(dtype());
-        m_node_ptr->set_data_ptr(m_data);
+
+        // Allow m_node_ptr to take ownership of m_data so that future
+        // release()/reset() calls will free it, lest we leak memory.
+        const index_t owning_allocator_id =
+            execution::DeviceMemory::is_device_ptr(m_data)
+                ? execution::get_device_allocator_id()
+                : execution::get_host_allocator_id();
+        m_node_ptr->take_data_ptr(m_data,
+                                  dtype().element_bytes() * number_of_elements(),
+                                  owning_allocator_id);
 
         // the assumed data is now the accessor's new original backing storage
         m_orig_data_ptr = m_data;
--- a/src/libs/conduit/conduit_data_accessor.cpp
+++ b/src/libs/conduit/conduit_data_accessor.cpp
@@ -597,29 +597,6 @@ DataAccessor<T>::sync()
                                                m_node_ptr->dtype().stride(),
                                                m_data,
                                                m_stride);
-
-        // By this point, m_data has been copied into m_node_ptr->data_ptr().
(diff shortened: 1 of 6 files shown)


src/libs/conduit/conduit_node.cpp
```diff
@@ -17595,6 +17595,19 @@ Node::set_data_ptr(void *data)
     m_data    = data;
 }
 
+//---------------------------------------------------------------------------//
+void
+Node::take_data_ptr(void *data_ptr,
```
cyrush: should we call this `assume_data_ptr`?

siramok: Sure thing, I went ahead and made this change.

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -1421,6 +1425,133 @@ TEST(conduit_execution, strawman_data_array)
...
+    return umpire::ResourceManager::getInstance()
+               .getAllocator(pool_name)
+               .getCurrentSize();
+}
+
+//-----------------------------------------------------------------------------
+static void
+expect_no_leak(void (*run_fn)(Node &, ExecutionPolicy),
+               index_t node_alloc_id,
+               ExecutionPolicy policy)
```
JustinPrivitera: should these live in the execution test utils instead?

src/tests/conduit/t_conduit_execution.cpp
JustinPrivitera, line 1: these tests are very cool

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -622,7 +622,16 @@ DataAccessor<T>::assume()
...
+
+        // Allow m_node_ptr to take ownership of m_data so that future
+        // release()/reset() calls will free it, lest we leak memory.
+        const index_t owning_allocator_id =
+            execution::DeviceMemory::is_device_ptr(m_data)
+                ? execution::get_device_allocator_id()
+                : execution::get_host_allocator_id();
+        m_node_ptr->assume_data_ptr(m_data,
+                                    dtype().element_bytes() * number_of_elements(),
+                                    owning_allocator_id);
```
JustinPrivitera, line 634: I just wonder if this is bulletproof to other kinds of allocators? Like if someone had their own device/host allocator I wonder if this would matter that we use the default ones? I imagine this will be fine in all or almost all cases.

JustinPrivitera (approved):
Thanks for catching this and being so thorough in fixing/testing! Great work 🦾

src/tests/conduit/t_conduit_execution.cpp
siramok: These have been moved to the execution test utils.

src/libs/conduit/conduit_data_accessor.cpp
siramok: My understanding, at least with respect to how things currently exist, is that `DataArray` and `DataAccessor` will only ever allocate their working buffers (execution location) in host or device memory using the default host/device allocators. A custom allocator can be set to use for the output location, but `DataArray` and `DataAccessor` don't check for the existence of a custom allocator when creating their working buffers.

So in the `assume()` case where we transfer ownership of memory in whatever memory space we executed from (which we understand can be different than the source location), it seems that when we `assume` the working buffer away, the output node would end up with data allocated by one of the default allocators and not the custom allocator.

If that ever changes (`DataArray` and `DataAccessor`  start using custom allocators when provided), then yes, this implementation would have to change as well. As the codebase currently exists, it should never create a problem.

Separate from what this PR does, I agree that it would perhaps be a good idea to make `DataArray` and `DataAccessor`  take custom allocators into consideration.

siramok:
The remaining piece of open feedback could perhaps become a comment for future TODO under #1358. I'll leave this PR open to make sure it doesn't get missed, but I otherwise consider this PR ready to merge.

JustinPrivitera:
> The remaining piece of open feedback could perhaps become a comment for future TODO under #1358. I'll leave this PR open to make sure it doesn't get missed, but I otherwise consider this PR ready to merge.

could you add a TODO there?

JustinPrivitera (approved):
good to merge as long as you add the TODO

siramok:
The TODO has been added.

PR #1658, opened 2026-07-28
Add sphinx documentation for the data-parallel execution model

siramok:
I drafted a few pages of outline for what I imagined these docs looking like, put it all into an .rst file, and then asked an LLM to format it correctly for sphinx and make it look nice. Specifically, I included all of the relevant code files as context and then asked it to follow my outline but improve the content where it made sense to do. I took a pass over the changes and think it did a pretty good job. I built an offline copy of the docs for myself, and the layout of the new page seemed quite good to me (sphinx is a good tool).

Despite that, I'm labeling this PR as WIP because 1) I want to make further edits to the text and 2) I want to personally double check its accuracy before requesting a review.


siramok:
I took a pass and made some tweaks. The remaining problem is that there's some repetition across examples, e.g. sync strategies. There also may be a more logical order to present the information, e.g. explain the boilerplate first and then reducers, sorts, and atomics last.

Depending on preference maybe you'd rather not abbreviate to DEM, which is fine. I assume there are other preference/consistency tweaks that make sense to implement as well.

src/docs/sphinx/device_execution_model.rst
```diff
@@ -0,0 +1,320 @@
+.. # Copyright (c) Lawrence Livermore National Security, LLC and other Conduit
+.. # Project developers. See top-level LICENSE AND COPYRIGHT files for dates and
+.. # other details. No copyright assignment is required to contribute to Conduit.
+
+.. _device_execution_model:
+
+======================
+Device Execution Model
```
JustinPrivitera: I don't know that I would call it a _Device Execution Model_ necessarily, as it does more than just execute on the device (serial + openmp are options). _Host/Device Execution Model_ is more accurate but is not a great name. Maybe we can workshop but I think we shouldn't call it a _Device Execution Model_.

src/docs/sphinx/device_execution_model.rst
```diff
@@ -0,0 +1,320 @@
...
+======================
+
+.. note::
+    The Device Execution Model APIs and docs are under active development.
+
+Conduit is a library that provides an intuitive model for describing hierarchical scientific data. Conduit Blueprint is a set of methods for building, verifying, and transforming scientific datasets according to known conventions called protocols. In practice, this frequently involves representing 2D and 3D meshes along with corresponding topologies, coordsets, and field data.
```
JustinPrivitera: I'm not sure it makes sense to summarize Conduit and Conduit Blueprint here in this manner. We could link to the documentation for both of these sections.

On the other hand, this paragraph is good content for your poster.

src/docs/sphinx/device_execution_model.rst
```diff
@@ -0,0 +1,320 @@
...
+
+Previously, these APIs only executed on the CPU. The Device Execution Model (DEM) is an API layer that provides tools for building accelerator-agnostic implementations of these operations. It leverages `RAJA <https://github.com/LLNL/RAJA>`_ when Conduit is built with RAJA support and provides CPU-based fallback implementations otherwise. This allows Conduit to preserve existing behavior while enabling new GPU-accelerated workflows through existing APIs.
```
JustinPrivitera: What APIs are being referred to in this first sentence?

It would be good to mention that we are using Umpire for memory management as well.

JustinPrivitera:
> Depending on preference maybe you'd rather not abbreviate to DEM, which is fine. I assume there are other preference/consistency tweaks that make sense to implement as well.

I think I would rather not abbreviate to DEM. I'm also not sure if we want to call it a **Device** Execution Model at all, since it does more than just execute on the device. It's really a _Host/Device Execution Model_ or just an _Execution Model_.

src/docs/sphinx/device_execution_model.rst
```diff
@@ -0,0 +1,320 @@
...
+
+The DEM lives in the ``conduit::execution`` namespace and is declared in ``conduit_execution.hpp`` and ``conduit_execution_policy.hpp``.
+
+Goals
+-----
+
+Modern high-performance computers are heterogeneous, meaning they include multiple types of processors within a single node. The most common pairing is one or two CPUs alongside several GPUs. Simulation codes are increasingly GPU-accelerated, which means Conduit Blueprint data may originate in device memory. Previously, this data always had to be copied to host memory before Blueprint APIs could operate on it. The goal of the DEM is to give users flexibility in where these Blueprint operations execute. Porting existing APIs to the DEM generally involves better parallelization, and in some cases the introduction of parallelization where none existed previously. This has led to performance improvements in both CPU- and GPU-accelerated workloads.
```
JustinPrivitera: It would be good to mention what the Blueprint APIs are.

src/docs/sphinx/device_execution_model.rst
```diff
@@ -0,0 +1,320 @@
...
+
+   * - ``sync_strategy``
+     - ``"sync"``, ``"assume"``
+     - ``"assume"``
+     - How results move back to the output node. See `Sync Strategies`_.
+
+   * - ``fallback_location``
+     - ``"host"``, ``"device"``
+     - ``"host"``
+     - Where to execute and allocate when ``"input"`` is selected but there is no input node to reason about.
```
JustinPrivitera: Let's get some newlines into the table so that the whole table can render without needing a scroll bar.

Image

src/docs/sphinx/device_execution_model.rst
```diff
@@ -0,0 +1,320 @@
...
+Helper methods resolve the current options, optionally against an input node:
+
+.. code:: cpp
+
+    namespace execution = conduit::execution;
+
+    ExecutionPolicy         policy     = execution::get_execution_policy(src_node);
+    index_t                 alloc_id   = execution::get_output_allocator_id(src_node);
+    execution::SyncStrategy strategy   = execution::get_sync_strategy();
+    index_t                 dev_alloc  = execution::get_device_allocator_id();
```
JustinPrivitera: why is it  `ExecutionPolicy` and `execution::SyncStrategy`?

Why not `execution::ExecutionPolicy` and `execution::SyncStrategy` or  `ExecutionPolicy` and `SyncStrategy`?

src/docs/sphinx/device_execution_model.rst
```diff
@@ -0,0 +1,320 @@
...
+   * - ``ExecutionPolicy::device()``
+     - ``cuda()`` or ``hip()``, depending on the build. Errors with neither.
+
+   * - ``ExecutionPolicy::parallel()``
+     - First available of ``cuda()``, ``hip()``, ``openmp()``, then ``serial()``.
+
+   * - ``ExecutionPolicy::empty()``
+     - No policy. DEM calls made with an empty policy raise an error.
+
+Policies can also be constructed from a name: ``ExecutionPolicy("device")`` accepts ``"empty"``, ``"serial"``, ``"cuda"``, ``"hip"``, ``"openmp"``, and the aliases ``"host"``, ``"device"``, and ``"parallel"``.
```
JustinPrivitera:
```suggestion
Policies can also be constructed from a (string) name: ``ExecutionPolicy(<policy name>)`` accepts ``"empty"``, ``"serial"``, ``"cuda"``, ``"hip"``, ``"openmp"``, and the aliases ``"host"``, ``"device"``, and ``"parallel"``.
```
or something like this

src/docs/sphinx/device_execution_model.rst
```diff
@@ -0,0 +1,320 @@
...
+
+Atomic updates are safe to call from any kernel:
+
+.. code:: cpp
+
+    conduit::execution::atomic_add(&acc, value);
+    conduit::execution::atomic_min(&acc, value);
+    conduit::execution::atomic_max(&acc, value);
+
+Each returns the value held at ``acc`` before the update.
```
JustinPrivitera: Might be useful to show how these are used. Can we copy anything from tests?

src/docs/sphinx/device_execution_model.rst
```diff
@@ -0,0 +1,320 @@
...
+
+* ``use_with(policy)``: make the data accessible in the memory space of ``policy``, moving it if it is not already there.
+* ``sync()``: copy the working buffer back to the node's original memory space. A no-op if the data never moved.
+* ``assume()``: the node takes ownership of the working buffer in whatever memory space it currently occupies.
+* ``active_space()``: returns an ``ExecutionPolicy`` describing the memory space where the data currently lives. This lets you execute wherever the input already resides:
+
+.. code:: cpp
+
+    float64_accessor acc_src(node["src"]);
+    float64_accessor acc_des(node["des"]);
```
JustinPrivitera: It would be useful to explain that accessors/arrays can wrap node leaf data, and constructing them in this special way gives us access to the memory movement methods.

src/docs/sphinx/device_execution_model.rst
```diff
@@ -0,0 +1,320 @@
...
+
+    conduit::execution::sort_ascending(policy, begin, end);
+    conduit::execution::sort_descending(policy, begin, end);
+
+Moving Data Between Memory Spaces
+---------------------------------
+
+``DataAccessor`` (``float64_accessor``, ``index_t_accessor``, ...) and ``DataArray`` (``float64_array``, ``index_t_array``, ...) wrappers expose the data movement methods used to make node leaf data available to kernels:
+
+* ``use_with(policy)``: make the data accessible in the memory space of ``policy``, moving it if it is not already there.
```
JustinPrivitera: I think use_with could use more explanation of what is happening with the data. What happens to the original data? The accessor/array straddles the two memory spaces, so how is ownership affected?

src/docs/sphinx/device_execution_model.rst
```diff
@@ -0,0 +1,320 @@
...
+
+    acc_des.sync();
+
+Sync Strategies
+~~~~~~~~~~~~~~~
+
+DEM-aware transforms stage output in a working buffer sized for the execution memory space, then reconcile it with the output node according to the global ``sync_strategy`` option:
+
+* ``"sync"``: the working buffer is copied to the final destination, which can live in a different memory space (for example, back to the host).
+* ``"assume"`` (default): the output node takes ownership of the working buffer in whatever memory space it occupies, avoiding a final copy.
```
JustinPrivitera: Do we need this section? We just discussed sync and assume above. It might be best to put more info in the Execution Options section about this and then remove this section entirely.

src/docs/sphinx/device_execution_model.rst
```diff
@@ -0,0 +1,320 @@
...
+
+Conduit Nodes can hold data that lives in host or device memory. The DEM provides the fundamental tools for writing accelerator-agnostic transformations on this data.
+
+When Conduit is built with CUDA or HIP, call ``init_device_memory_handlers()`` once during setup. It initializes memory-space-aware copy and set handlers so that Node operations (``set()``, assignment, and copies between Nodes) work transparently across host and device memory:
+
+.. code:: cpp
+
+    conduit::execution::init_device_memory_handlers();
+
+A minimal example that doubles the values of an array using device execution:
```
JustinPrivitera: This example is a little strange.

I think it is odd that node["src"] gets allocated in the output memory space.

Then later use_with() is not doing anything because the allocation location is already the execution location. This can happen, but is not as effective as an example of how these things work.

It might be useful to use one or multiple of the strawman examples. It is possible to link code blocks in docs directly without copying the code.

JustinPrivitera (changes requested):
Thanks for putting this together. I think we can massage this and continue thinking about how best to tell the story. There is good content here we can put on your poster as well.

siramok:
Thanks for your feedback! I'll work on addressing these comments.

src/docs/sphinx/device_execution_model.rst
siramok: This has been addressed. I went with "Data-Parallel Execution Model" as the formal name and refer to it as just "Execution Model" throughout.

src/docs/sphinx/device_execution_model.rst
siramok: Addressed this by linking to Conduit and Conduit Blueprint, and not summarizing them here.

src/docs/sphinx/device_execution_model.rst
siramok: Addressed this by being more specific that we've ported over `mesh::convert` and the underlying topology and coordset conversions. Added mentions of Umpire as well.

src/docs/sphinx/device_execution_model.rst
siramok: Do you mean outline them specifically (e.g., `to_*`) or give a high-level explanation? I think the latest changes have a high-level explanation, but not a specific outline.

src/docs/sphinx/device_execution_model.rst
siramok: Changed this to not be a table, because I couldn't get it to not be too wide to avoid scrolling.

src/docs/sphinx/device_execution_model.rst
siramok: Addressed this by being consistent and using `ExecutionPolicy` and `SyncStrategy`.

src/docs/sphinx/device_execution_model.rst
siramok: I have included this change.

src/docs/sphinx/device_execution_model.rst
siramok: Added something about this.

src/docs/sphinx/device_execution_model.rst
siramok: The explanation for what `use_with` does and how it works has been expanded.

src/docs/sphinx/device_execution_model.rst
siramok: Addressed by getting rid of this section and having a `Moving Data Between Memory Spaces` section instead, which covers the accessors, `use_with`, etc.

src/docs/sphinx/device_execution_model.rst
siramok: Changed the example around to start in host memory and execute in device memory, and then borrowed an example from our tests. Did you mean that sphinx itself has a way of linking to code, or that I should use a permanent hyperlink from GitHub? Either way, what's currently there is not linked, so it would theoretically go out of data if the test ever changed.

src/docs/sphinx/device_execution_model.rst
siramok: Addressed this by adding more of a real example of using these within a `forall`.

siramok:
> 

> I think I would rather not abbreviate to DEM. I'm also not sure if we want to call it a **Device** Execution Model at all, since it does more than just execute on the device. It's really a _Host/Device Execution Model_ or just an _Execution Model_.

Addressed this, seems like we liked "Data-Parallel Execution Model" so I went with that. Instead of abbreviating, I just say "Execution Model" to be less verbose.

siramok:
I did a full read-through of the rendered version of the new docs page and think it looks and sounds good. There are no more elements that are too wide/require horizontal scrolling. I verified that all of the links work as well. Besides the comments that I left unresolved (due to me asking follow-up questions), I would consider this PR to no longer be WIP and ready for re-review.

src/docs/sphinx/device_execution_model.rst
JustinPrivitera: sphinx itself has a way of linking to code: https://github.com/llnl/conduit/blob/529e7d7aaa8f45fe881e649b051d366f2c7c2a14/src/docs/sphinx/blueprint_mesh.rst?plain=1#L930

JustinPrivitera: see the lines around this as well

src/docs/sphinx/data_parallel_execution_model.rst
```diff
@@ -0,0 +1,332 @@
...
+
+Conduit Nodes can hold data that lives in host or device memory. The execution model provides the fundamental tools for writing transformations that operate on this data regardless of where it lives.
+
+When Conduit is built with CUDA or HIP, call ``init_device_memory_handlers()`` once during setup. It initializes memory-space-aware copy and set handlers so that Node operations (``set()``, assignment, and copies between Nodes) work transparently across host and device memory:
+
+.. code:: cpp
+
+    conduit::execution::init_device_memory_handlers();
+
+A minimal example that doubles the values of an array on the device, starting from data that lives on the host. First we initialize the memory handlers, select device execution, and allocate the input and output leaf data with the host allocator, which is also the default:
```
JustinPrivitera:
```suggestion
Below is a minimal example that doubles the values of an array on the device, starting from data that lives on the host. First we initialize the memory handlers, select device execution, and allocate the input and output leaf data with the host allocator, which is also the default:
```

let's make it a complete sentence

src/docs/sphinx/data_parallel_execution_model.rst
```diff
@@ -0,0 +1,332 @@
...
+    node["src"].set(src_vals);
+    node["des"].set(des_vals);
+
+    // wrap the leaf data in accessors, execute with the selected policy,
+    // and sync the results back to node["des"]
+    run_data_accessor_policy_and_sync(node, policy);
+
+The work itself happens in ``run_data_accessor_policy_and_sync()``, borrowed from Conduit's execution tests. It wraps the node leaf data in accessors, moves that data to the memory space of the policy, doubles each value, and copies the results back to where ``node["des"]`` started:
+
+.. code:: cpp
```
JustinPrivitera: let's literalinclude this so it reflects any changes that are made in the future

src/docs/sphinx/data_parallel_execution_model.rst
```diff
@@ -0,0 +1,332 @@
...
+    namespace execution = conduit::execution;
+
+    // setup: install memory handlers and select device execution
+    execution::init_device_memory_handlers();
+
+    Node exec_opts;
+    exec_opts["execution_location"] = "device";
+    execution::execution_set_options(exec_opts);
+
+    ExecutionPolicy policy = execution::get_execution_policy();
```
JustinPrivitera, line 69: maybe we could add a comment above this line saying that the policy will be a device policy because we requested the execution location to be the device

src/docs/sphinx/device_execution_model.rst
siramok: I see, I have made this change.

src/docs/sphinx/data_parallel_execution_model.rst
```diff
@@ -0,0 +1,332 @@
...
+.. code:: cpp
+
+    Node exec_opts;
+    exec_opts["execution_location"] = "device";
+    exec_opts["output_location"]    = "device";
+    exec_opts["sync_strategy"]      = "assume";
+    conduit::execution::execution_set_options(exec_opts);
+
+    Node cur_opts;
+    conduit::execution::execution_options(cur_opts);
```
JustinPrivitera, line 109:
```suggestion
    Node cur_opts;
    conduit::execution::execution_options(cur_opts);
    cur_opts.print();
```

src/docs/sphinx/data_parallel_execution_model.rst
```diff
@@ -0,0 +1,306 @@
...
+            // kernel body
+        });
+    CONDUIT_DEVICE_ERROR_CHECK(policy);
+
+Two macros make kernels portable:
+
+* ``CONDUIT_EXEC`` marks a function or lambda so it compiles for both host and device in CUDA/HIP translation units. In host-only builds it expands to nothing.
+* ``CONDUIT_DEVICE_ERROR_CHECK(policy)`` reports CUDA/HIP errors from the last kernel launch by printing to ``stderr``. It does not raise a ``conduit::Error`` or abort. It is a no-op for host policies.
+
+Requesting a backend that Conduit was not built with (for example, a CUDA policy without CUDA support) raises a ``conduit::Error``.
```
JustinPrivitera: should this line be in another section?

JustinPrivitera (changes requested):
Minor feedback; otherwise, this looks great!

src/docs/sphinx/data_parallel_execution_model.rst
siramok: I believe this change has been implemented.

src/docs/sphinx/data_parallel_execution_model.rst
siramok: I added a comment that says this.

src/docs/sphinx/data_parallel_execution_model.rst
siramok: Seems reasonable, I moved this sentence up so that it's in the Execution Policies section.

JustinPrivitera (approved):

JustinPrivitera:
Looks great, thanks for adding this!

PR #1661, opened 2026-07-30
Silo Reader: handle material dependent variables with different choices on different domains

### LINKED ISSUE #1652: Silo Reader doesn't handle material dependent variables with different choices on different domains
In Silo, if you have a material dependent variable split across multiple domains, if the underlying material set has no mixing on a particular domain, then the variable has no mixvals.

When we read such a variable into Blueprint, we have no way of knowing that it is supposed to be a material dependent variable based on the Silo representation, so we end up with some domains where it is material dependent and others where it is not.

This behavior breaks downstream tools like the partitioner. We need to handle this case at the Silo Reader.

JustinPrivitera:
Resolves #1652

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
@@ -2598,8 +2598,7 @@ get_material_names(const conduit::Node &matset,
 
 //-----------------------------------------------------------------------------
 bool
-has_mixed_elements(const conduit::Node &matset,
-                   const float64 epsilon = CONDUIT_EPSILON)
+has_mixed_elements(const conduit::Node &matset, const float64 epsilon)
 {
     // extra seat belt here
     if (! matset.dtype().is_object())
@@ -3544,7 +3543,7 @@ to_uni_buffer_by_material(const conduit::Node &src_matset,
 void
 create_field_matset_values_from_unmixed_matset(const conduit::Node &matset,
                                                conduit::Node &field,
-                                               const float64 epsilon = CONDUIT_EPSILON)
+                                               const float64 epsilon)
 {
     // extra seat belt here
     if (! matset.dtype().is_object())
@@ -3555,9 +3554,9 @@ create_field_matset_values_from_unmixed_matset(const conduit::Node &matset,
 
     MatsetAccessor m_acc = MatsetAccessor(matset);
 
-    if (is_uni_buffer(matset))
+    if (matset::is_uni_buffer(matset))
     {
-        if (is_element_dominant(matset))
+        if (matset::is_element_dominant(matset))
         {
             // sparse by element
 
@@ -3569,16 +3568,15 @@ create_field_matset_values_from_unmixed_matset(const conduit::Node &matset,
             // unsupported uni-buffer by material
             CONDUIT_ERROR("conduit::blueprint::mesh::matset::create_field_matset_values_from_unmixed_matset() "
                           "material-dominant uni-buffer material set is unsupported.");
-            return false;
         }
     }
     else // multi-buffer
     {
-        if (is_element_dominant(matset))
+        if (matset::is_element_dominant(matset))
         {
             // full
             
-            const index_t num_elems = field["values"].number_of_elements();
+            const index_t num_elems = field["values"].dtype().number_of_elements();
             float64_accessor field_values = field["values"].value();
             const Node &vol_fracs = matset["volume_fractions"];
             const std::vector<std::string> matnames = vol_fracs.child_names();
@@ -3594,7 +3592,7 @@ create_field_matset_values_from_unmixed_matset(const conduit::Node &matset,
                     const float64 vol_frac = vol_fracs_for_mat[elem_id];
                     if (1.0 - epsilon < vol_frac && vol_frac < 1.0 + epsilon)
                     {
-                        mset_vals_for_mat[elem_id].set(field_values[elem_id]);
+                        mset_vals_for_mat.set(elem_id, field_values[elem_id]);
                     }
                 }
             }
@@ -3619,7 +3617,7 @@ create_field_matset_values_from_unmixed_matset(const conduit::Node &matset,
                      index_of_elem_id ++)
                 {
                     const index_t elem_id = mat_elem_ids_vals[index_of_elem_id];
-                    mset_vals_for_mat[index_of_elem_id].set(field_values[elem_id]);
+                    mset_vals_for_mat.set(index_of_elem_id, field_values[elem_id]);
                 }
             }
         }
--- a/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
@@ -2614,6 +2614,8 @@ has_mixed_elements(const conduit::Node &matset,
     {
         if (is_element_dominant(matset))
         {
+            // sparse by element
+
             const index_t num_elems = m_acc.num_elems();
             for (index_t elem_idx = 0; elem_idx < num_elems; elem_idx ++)
             {
@@ -3536,6 +3538,92 @@ to_uni_buffer_by_material(const conduit::Node &src_matset,
     (void) epsilon;
     CONDUIT_ERROR("blueprint::mesh::field::to_uni_buffer_by_material() "
                   "converting from a material-dominant uni-buffer material set/field is unsupported.");
+}
+
+//-----------------------------------------------------------------------------
+void
+create_field_matset_values_from_unmixed_matset(const conduit::Node &matset,
+                                               conduit::Node &field,
+                                               const float64 epsilon = CONDUIT_EPSILON)
+{
+    // extra seat belt here
+    if (! matset.dtype().is_object())
+    {
+        CONDUIT_ERROR("blueprint::mesh::matset::create_field_matset_values_from_unmixed_matset"
+                      " passed matset node must be a valid matset tree.");
+    }
+
+    MatsetAccessor m_acc = MatsetAccessor(matset);
+
+    if (is_uni_buffer(matset))
+    {
+        if (is_element_dominant(matset))
+        {
+            // sparse by element
+
+            // all we need to do is copy
+            field["matset_values"].set(field["values"]);
+        }
+        else // material-dominant
+        {
+            // unsupported uni-buffer by material
+            CONDUIT_ERROR("conduit::blueprint::mesh::matset::create_field_matset_values_from_unmixed_matset() "
+                          "material-dominant uni-buffer material set is unsupported.");
+            return false;
+        }
+    }
+    else // multi-buffer
+    {
+        if (is_element_dominant(matset))
+        {
+            // full
+            
+            const index_t num_elems = field["values"].number_of_elements();
+            float64_accessor field_values = field["values"].value();
+            const Node &vol_fracs = matset["volume_fractions"];
+            const std::vector<std::string> matnames = vol_fracs.child_names();
+            for (const std::string &matname : matnames)
+            {
+                float64_accessor vol_fracs_for_mat = vol_fracs[matname].value();
+                field["matset_values"][matname].set(DataType::float64(num_elems));
+                float64_array mset_vals_for_mat = field["matset_values"][matname].as_float64_array();
+                mset_vals_for_mat.fill(0.0);
+
+                for (index_t elem_id = 0; elem_id < num_elems; elem_id ++)
+                {
+                    const float64 vol_frac = vol_fracs_for_mat[elem_id];
+                    if (1.0 - epsilon < vol_frac && vol_frac < 1.0 + epsilon)
+                    {
+                        mset_vals_for_mat[elem_id].set(field_values[elem_id]);
+                    }
+                }
+            }
+        }
+        else // material-dominant
+        {
+            // sparse_by_material
+            
+            float64_accessor field_values = field["values"].value();
+            const Node &elem_ids = matset["element_ids"];
+            const std::vector<std::string> matnames = elem_ids.child_names();
+            for (const std::string &matname : matnames)
+            {
+                index_t_accessor mat_elem_ids_vals = elem_ids[matname].value();
+                const index_t num_elems_for_mat = mat_elem_ids_vals.number_of_elements();
+                field["matset_values"][matname].set(DataType::float64(num_elems_for_mat));
+                float64_array mset_vals_for_mat = field["matset_values"][matname].value();
+
+                // iterate over element ids
+                for (index_t index_of_elem_id = 0;
+                     index_of_elem_id < num_elems_for_mat;
+                     index_of_elem_id ++)
+                {
+                    const index_t elem_id = mat_elem_ids_vals[index_of_elem_id];
+                    mset_vals_for_mat[index_of_elem_id].set(field_values[elem_id]);
+                }
+            }
+        }
+    }
 }
 
 //-----------------------------------------------------------------------------
--- a/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
@@ -2597,6 +2597,94 @@ get_material_names(const conduit::Node &matset,
 }
 
 //-----------------------------------------------------------------------------
+bool
+has_mixed_elements(const conduit::Node &matset,
+                   const float64 epsilon = CONDUIT_EPSILON)
+{
+    // extra seat belt here
+    if (! matset.dtype().is_object())
+    {
+        CONDUIT_ERROR("blueprint::mesh::matset::has_mixed_elements"
+                      " passed matset node must be a valid matset tree.");
+    }
+
+    MatsetAccessor m_acc = MatsetAccessor(matset);
+
+    if (is_uni_buffer(matset))
+    {
+        if (is_element_dominant(matset))
+        {
+            const index_t num_elems = m_acc.num_elems();
+            for (index_t elem_idx = 0; elem_idx < num_elems; elem_idx ++)
+            {
+                if (m_acc.num_mats_for_elem(elem_idx) != index_t{1})
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+        else // material-dominant
+        {
+            // unsupported uni-buffer by material
+            CONDUIT_ERROR("conduit::blueprint::mesh::matset::has_mixed_elements() "
+                          "material-dominant uni-buffer material set is unsupported.");
+            return false;
+        }
+    }
+    else // multi-buffer
+    {
+        if (is_element_dominant(matset))
+        {
+            // full
+            const index_t num_elems = m_acc.num_elems();
+            const index_t num_mats  = m_acc.num_mats();
+
+            for (index_t elem_idx = 0; elem_idx < num_elems; elem_idx ++)
+            {
+                for (index_t mat_idx = 0; mat_idx < num_mats; mat_idx ++)
+                {
+                    const float64 vol_frac = m_acc.get_vol_frac(elem_idx, mat_idx);
+                    const float64 diff = 1.0f - vol_frac;
+                    // volume fraction must be neither 1.0 nor 0.0 for this to 
+                    // be clean
+                    // if vol_frac is greater than 0 and
+                    // 1 - vol_frac is greater than 0 (1 is greater than vol_frac)
+                    if (vol_frac > epsilon && diff > epsilon)
+                    {
+                        // then we are mixed
+                        return true;
+                    }
+                }
+            }
+
+            return false;
+        }
+        else // material-dominant
+        {
+            // sparse_by_material
+            const index_t num_elems = m_acc.num_elems();
+            const index_t num_mats  = m_acc.num_mats();
+
+            index_t running_sum = 0;
+            for (index_t mat_idx = 0; mat_idx < num_mats; mat_idx ++)
+            {
+                running_sum += m_acc.num_elems_for_mat(mat_idx);
+            }
+
+            // if any of the elements are double counted
+            if (running_sum > num_elems)
+            {
+                return true;
+            }
+
+            return false;
+        }
+    }
+}
+
+//-----------------------------------------------------------------------------
 void
 to_multi_buffer_by_element(const conduit::Node &src_matset,
                            conduit::Node &dest_matset)
--- a/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
@@ -2598,7 +2598,8 @@ get_material_names(const conduit::Node &matset,
 
 //-----------------------------------------------------------------------------
 bool
-has_mixed_elements(const conduit::Node &matset, const float64 epsilon)
+has_mixed_elements(const conduit::Node &matset,
+                   const float64 epsilon)
 {
     // extra seat belt here
     if (! matset.dtype().is_object())
--- a/src/libs/relay/conduit_relay_io_silo.cpp
+++ b/src/libs/relay/conduit_relay_io_silo.cpp
@@ -4121,6 +4121,9 @@ honor_material_dependent_fields(const int domain_start,
     // mesh_radius_b: "no"
     // mesh_radius_c: "no"
 
+    int error = 0;
+    std::ostringstream error_oss;
+
     for (int domain_id = domain_start; domain_id < domain_end; domain_id ++)
     {
         const std::string domain_path = conduit_fmt::format("domain_{:06d}", domain_id);
@@ -4161,20 +4164,53 @@ honor_material_dependent_fields(const int domain_start,
             }
             else
             {
-                // TODO parallel error checking
-                CONDUIT_ERROR("field " << fieldname << " was not read correctly "
-                              "from Silo.");
+                error = 1;
+                error_oss << "Field " << fieldname << " on domain " << domain_id << 
+                             " was not read correctly from Silo.";
+                break;
             }
+        }
+
+        // break out of the domain loop if an error was encountered
+        if (error == 1)
+        {
+            break;
         }
     }
 
+    // parallel error handling & collect global data
     Node global_collected_fields_material_status;
 #ifdef CONDUIT_RELAY_IO_MPI_ENABLED
-    relay::mpi::all_gather_using_schema(local_fields_material_status,
-                                        global_collected_fields_material_status,
-                                        mpi_comm);
+    Node n_local, n_global;
+    n_local.set((int)error);
+    relay::mpi::sum_all_reduce(n_local, n_global, mpi_comm);
+    error = n_global.as_int();
+
+    if (1 == error)
+    {
+        // we have a problem, broadcast string messages
+        // to all ranks so they can throw
+        n_local.set(error_oss.str());
+        relay::mpi::all_gather_using_schema(n_local, n_global, mpi_comm);
+        CONDUIT_ERROR(n_global.as_string());
+    }
+    else
+    {
+        relay::mpi::all_gather_using_schema(local_fields_material_status,
+                                            global_collected_fields_material_status,
+                                            mpi_comm);
+    }
+
 #else
-    global_collected_fields_material_status.append().set_external(local_fields_material_status);
+    // non MPI case, throw error
+    if (1 == error)
+    {
+        CONDUIT_ERROR(error_oss.str());
+    }
+    else
+    {
+        global_collected_fields_material_status.append().set_external(local_fields_material_status);
+    }
 #endif
 
     // after the communication, we have a structure like this:
@@ -4247,6 +4283,7 @@ honor_material_dependent_fields(const int domain_start,
     // now for the fun part - we must ensure that all of the fields that are globally
     // material dependent have matset_values.
 
+
     for (int domain_id = domain_start; domain_id < domain_end; domain_id ++)
     {
         // fetch the domain path for this domain_id
@@ -4283,9 +4320,10 @@ honor_material_dependent_fields(const int domain_start,
                 }
                 else if (! globally_matdep && locally_matdep)
                 {
-                    // TODO parallel error checking
-                    CONDUIT_ERROR("Material dependence could not be established for field "
-                                  << fieldname);
+                    error = 1;
+                    error_oss << "Material dependence could not be established for field "
+                                  << fieldname << " on domain " << domain_id;
+                    return false;
                 }
                 else // (! globally_matdep && ! locally_matdep)
                 {
@@ -4295,6 +4333,12 @@ honor_material_dependent_fields(const int domain_start,
 
             return changes_needed;
         }();
+
+        // break out of the domain loop if an error was encountered
+        if (error == 1)
+        {
+            break;
+        }
 
         // If there are not any material dependent fields on this domain, we can skip
         if (! any_matdep_fields_requiring_changes)
@@ -4310,16 +4354,18 @@ honor_material_dependent_fields(const int domain_start,
         // check for the existence of matsets
         if (! mesh.has_path(domain_path + "/matsets"))
         {
-            // TODO parallel error checking
-            CONDUIT_ERROR("Missing matsets for domain " << domain_path);
+            error = 1;
+            error_oss << "Missing matsets for domain " << domain_path;
+            break; // break out of the domain loop if an error was encountered
         }
 
         // make sure there is only one matset
         const Node &mesh_matsets = mesh[domain_path]["matsets"];
         if (mesh_matsets.number_of_children() != 1)
         {
-            // TODO parallel error checking
-            CONDUIT_ERROR("Ambiguous matsets for domain " << domain_path);
+            error = 1;
+            error_oss << "Ambiguous matsets for domain " << domain_path;
+            break; // break out of the domain loop if an error was encountered
         }
 
         // fetch the matset
@@ -4331,10 +4377,11 @@ honor_material_dependent_fields(const int domain_start,
         const bool matset_has_mixed_elems = conduit::blueprint::mesh::matset::has_mixed_elements(matset);
         if (matset_has_mixed_elems)
         {
-            // TODO parallel error checking
-            CONDUIT_ERROR("Matset has mixed elements so we cannot infer "
-                          "field matset values for material dependent "
-                          "fields on domain " << domain_path);
+            error = 1;
+            error_oss << "Matset has mixed elements so we cannot infer "
+                         "field matset values for material dependent "
+                         "fields on domain " << domain_path;
+            break; // break out of the domain loop if an error was encountered
         }
 
         // finally we can fetch the fields and iterate over them
@@ -4366,7 +4413,29 @@ honor_material_dependent_fields(const int domain_start,
                 // we have already errored.
             }
         }
+    } // end loop over domains
+
+    // parallel error handling
+#ifdef CONDUIT_RELAY_IO_MPI_ENABLED
+    n_local.set((int)error);
+    relay::mpi::sum_all_reduce(n_local, n_global, mpi_comm);
+    error = n_global.as_int();
+
+    if (1 == error)
+    {
+        // we have a problem, broadcast string messages
+        // to all ranks so they can throw
+        n_local.set(error_oss.str());
+        relay::mpi::all_gather_using_schema(n_local, n_global, mpi_comm);
+        CONDUIT_ERROR(n_global.as_string());
     }
+#else
+    // non MPI case, throw error
+    if (1 == error)
+    {
+        CONDUIT_ERROR(error_oss.str());
+    }
+#endif
 }
 
 //-----------------------------------------------------------------------------
@@ -5218,7 +5287,7 @@ read_mesh(const std::string &root_file_path,
                 file_map_itr ++;
             }
         }
-    }
+    } // end loop over domains
 
     //
     // Post-processing
--- a/src/libs/relay/conduit_relay_io_silo.cpp
+++ b/src/libs/relay/conduit_relay_io_silo.cpp
@@ -4956,6 +4956,20 @@ read_mesh(const std::string &root_file_path,
 
     Node local_fields_material_status;
 
+    // we create a structure like this:
+    //
+    // mesh_area: "yes"
+    // mesh_background: "no"
+    // mesh_circle_a: "no"
+    // mesh_circle_b: "no"
+    // mesh_circle_c: "no"
+    // mesh_importance: "yes"
+    // mesh_mat_check: "yes"
+    // mesh_overlap: "no"
+    // mesh_radius_a: "no"
+    // mesh_radius_b: "no"
+    // mesh_radius_c: "no"
+
     for (int domain_id = domain_start; domain_id < domain_end; domain_id ++)
     {
         const std::string domain_path = conduit_fmt::format("domain_{:06d}", domain_id);
@@ -4971,16 +4985,15 @@ read_mesh(const std::string &root_file_path,
         {
             const Node &field = field_itr.next();
             const std::string fieldname = field_itr.name();
-
             const bool has_matset      = field.has_child("matset");
-            const bool has_matset_vals = field.has_child("matset_vals");
+            const bool has_matset_vals = field.has_child("matset_values");
 
             if (has_matset && has_matset_vals)
             {
                 // Whether we have a note about this field or not,
                 // if we are material dependent on this domain then
                 // we want to mark this field as material dependent.
-                local_fields_material_status["fieldname"].set("yes");
+                local_fields_material_status[fieldname].set("yes");
             }
             else if (! has_matset && ! has_matset_vals)
             {
@@ -4992,7 +5005,7 @@ read_mesh(const std::string &root_file_path,
                 // that down.
                 if (! local_fields_material_status.has_child(fieldname))
                 {
-                    local_fields_material_status["fieldname"].set("no");
+                    local_fields_material_status[fieldname].set("no");
                 }
             }
             else
@@ -5003,21 +5016,165 @@ read_mesh(const std::string &root_file_path,
         }
     }
 
-    Node global_fields_material_status;
+    Node global_collected_fields_material_status;
 #ifdef CONDUIT_RELAY_IO_MPI_ENABLED
-    relay::mpi::gather_using_schema(local_fields_material_status,
-                                    global_fields_material_status,
-                                    0, // rank 0
-                                    mpi_comm);
+    relay::mpi::all_gather_using_schema(local_fields_material_status,
+                                        global_collected_fields_material_status,
+                                        mpi_comm);
 #else
-    global_fields_material_status.append().set_external(local_type_domain_info);
+    global_collected_fields_material_status.append().set_external(local_fields_material_status);
 #endif
 
-    if (0 == par_rank)
+    global_collected_fields_material_status.print();
+
+    // after the communication, we have a structure like this:
+    // 
+    // - 
+    //   mesh_area: "yes"
+    //   mesh_background: "no"
+    //   mesh_circle_a: "no"
+    //   mesh_circle_b: "no"
+    //   mesh_circle_c: "no"
+    //   mesh_importance: "yes"
+    //   mesh_mat_check: "yes"
+    //   mesh_overlap: "no"
+    //   mesh_radius_a: "no"
+    //   mesh_radius_b: "no"
+    //   mesh_radius_c: "no"
+    // -
+    //   mesh_area: "yes"
+    //   mesh_background: "no"
+    //   mesh_circle_a: "no"
+    //   ...
+
+    // now we must unify the sources of truth across ranks
+    Node fields_material_status;
+    auto rank_data_itr = global_collected_fields_material_status.children();
+    while (rank_data_itr.has_next())
     {
-        // TODO left off here
+        const Node &rank_data = rank_data_itr.next();
+        auto field_status_itr = rank_data.children();
+        while (field_status_itr.has_next())
+        {
+            const Node &field_status = field_status_itr.next();
+            const std::string fieldname = field_status_itr.name();
+            
+            // if we already have an entry for this field
+            if (fields_material_status.has_child(fieldname))
+            {
+                // then we choose "yes" over "no"
+                if (field_status.as_string() == "yes" &&
+                    fields_material_status[fieldname].as_string() == "no")
+                {
+                    fields_material_status[fieldname].set(field_status);
+                }
+            }
+            // if we do not yet have an entry for this field
+            else
+            {
+                // then add the one we have found
+                fields_material_status[fieldname].set(field_status);
+            }
+        }
     }
 
+    // finally, we are back to a structure like this, but unified over all ranks:
+    //
+    // mesh_area: "yes"
+    // mesh_background: "no"
+    // mesh_circle_a: "no"
+    // mesh_circle_b: "no"
+    // mesh_circle_c: "no"
+    // mesh_importance: "yes"
+    // mesh_mat_check: "yes"
+    // mesh_overlap: "no"
+    // mesh_radius_a: "no"
+    // mesh_radius_b: "no"
+    // mesh_radius_c: "no"
+    //
+    // stored in `fields_material_status`
+
+    // now for the fun part - we must ensure that all of the fields that are globally
+    // material dependent have matset_values.
+
+    if (any_matdep_fields)
+    {
+        for (int domain_id = domain_start; domain_id < domain_end; domain_id ++)
+        {
+            const std::string domain_path = conduit_fmt::format("domain_{:06d}", domain_id);
+            
+            if (! mesh.has_path(domain_path + "/fields"))
+            {
+                continue;
+            }
+
+            // are there any material dependent fields on this domain?
+            const bool any_matdep_fields = [fields_material_status]()
+            {
+                bool are_there_any_matdep_fields = false;
+
+                Node &mesh_fields = mesh[domain_path]["fields"];
+                auto field_itr = mesh_fields.children();
+                while (field_itr.has_next())
+                {
+                    field_itr.next();
+                    const std::string fieldname = field_itr.name();
+                    if (fields_material_status[fieldname].as_string() == "yes")
+                    {
+                        are_there_any_matdep_fields = true;
+                    }
+                }
+            };
+
+            // if there are not any material dependent fields on this domain, we can skip
+            if (! any_matdep_fields)
+            {
+                continue;
+            }
+
+            // we have at least one material dependent field on this domain, so we must
+            // find its matset
+
+            if (! mesh.has_path(domain_path + "/matsets"))
+            {
+                CONDUIT_ERROR("TODO");
+                // TODO parallel error checking
+            }
+
+            const Node &mesh_matsets = mesh[domain_path]["matsets"];
+            const std::vector<std::string> matset_names = mesh_matsets.child_names();
+            if (matset_names.size() != 1)
+            {
+                CONDUIT_ERROR("TODO");
+                // TODO parallel error checking
+            }
+
+            const Node &matset = mesh_matsets[matset_names[0]];
+
+            // TODO JUSTIN
+            // we only need to check if matsets are clean if we have a 
+            // globally material dependent field while the local field
+            // is material-independent - but ideally we could only do this once for all fields
+
+            Node &mesh_fields = mesh[domain_path]["fields"];
+            auto field_itr = mesh_fields.children();
+            while (field_itr.has_next())
+            {
+                const Node &field = field_itr.next();
+                const std::string fieldname = field_itr.name();
+
+                const bool globally_matdep = fields_material_status[fieldname].as_string() == "yes";
+                const bool locally_matdep  = field.has_child("matset");
+
+                if (globally_matdep && ! locally_matdep)
+                {
+                    // TODO JUSTIN
+                    // I wrote a new helper just for you: has_mixed_elements
+                    // but I want to cache the result
+                }
+            }
+        }
+    }
 
     //
     // Cleanup
--- a/src/libs/relay/conduit_relay_io_silo.cpp
+++ b/src/libs/relay/conduit_relay_io_silo.cpp
@@ -4090,6 +4090,286 @@ read_var_attributes(DBfile *dbfile,
 }
 
 //-----------------------------------------------------------------------------
+// We need an additional step to examine fields we have read from Silo.
+// If a material is clean (unmixed) for an entire domain, then any
+// material-dependent fields on that domain will not present as such.
+// We have no way of knowing until after we have read in the entire mesh
+// to check for this case. Blueprint requires material-dependent fields
+// to have "matset" and "matset_vals" children, even if there is no
+// material mixing for a particular domain. To keep Blueprint happy,
+// it is necessary to perform post-processing for fields to ensure that
+// material dependence is consistent across all domains.
+void
+honor_material_dependent_fields(const int domain_start,
+                                const int domain_end,
+                                Node &mesh
+                                CONDUIT_RELAY_COMMUNICATOR_ARG(MPI_Comm mpi_comm))
+{
+    Node local_fields_material_status;
+
+    // we create a structure like this:
+    //
+    // mesh_area: "yes"
+    // mesh_background: "no"
+    // mesh_circle_a: "no"
+    // mesh_circle_b: "no"
+    // mesh_circle_c: "no"
+    // mesh_importance: "yes"
+    // mesh_mat_check: "yes"
+    // mesh_overlap: "no"
+    // mesh_radius_a: "no"
+    // mesh_radius_b: "no"
+    // mesh_radius_c: "no"
+
+    for (int domain_id = domain_start; domain_id < domain_end; domain_id ++)
+    {
+        const std::string domain_path = conduit_fmt::format("domain_{:06d}", domain_id);
+        
+        if (! mesh.has_path(domain_path + "/fields"))
+        {
+            continue;
+        }
+        Node &mesh_fields = mesh[domain_path]["fields"];
+
+        auto field_itr = mesh_fields.children();
+        while (field_itr.has_next())
+        {
+            const Node &field = field_itr.next();
+            const std::string fieldname = field_itr.name();
+            const bool has_matset      = field.has_child("matset");
+            const bool has_matset_vals = field.has_child("matset_values");
+
+            if (has_matset && has_matset_vals)
+            {
+                // Whether we have a note about this field or not,
+                // if we are material dependent on this domain then
+                // we want to mark this field as material dependent.
+                local_fields_material_status[fieldname].set("yes");
+            }
+            else if (! has_matset && ! has_matset_vals)
+            {
+                // If we have already looked at the field, and on this 
+                // domain there is no matset association, then we defer 
+                // to other domains.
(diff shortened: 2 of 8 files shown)


rhornung67:
@JustinPrivitera to be clear, is this PR a draft?!?  😆

JustinPrivitera:
> @JustinPrivitera to be clear, is this PR a draft?!? 😆

I don't know what you're talking about 😉

JustinPrivitera:
~~TODO add new function to API with a note and then write tests for it that trigger the errors.~~

JustinPrivitera:
@cyrush I added a changelog entry as well as tests for error checking.

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -2596,6 +2596,95 @@ get_material_names(const conduit::Node &matset,
     }
 }
 
+//-----------------------------------------------------------------------------
+bool
+has_mixed_elements(const conduit::Node &matset, const float64 epsilon)
```
siramok:
```suggestion
has_mixed_elements(const conduit::Node &matset,
                   const float64 epsilon)
```
For consistency with the rest of the file

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -2596,6 +2596,95 @@ get_material_names(const conduit::Node &matset,
...
+            // unsupported uni-buffer by material
+            CONDUIT_ERROR("conduit::blueprint::mesh::matset::has_mixed_elements() "
+                          "material-dominant uni-buffer material set is unsupported.");
+            return false;
+        }
+    }
+    else // multi-buffer
+    {
+        if (is_element_dominant(matset))
+        {
```
siramok, line 2641: Could this become:
```
else if (is_element_dominant(matset))
{
```
and remove a layer of indentation?

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -2596,6 +2596,95 @@ get_material_names(const conduit::Node &matset,
...
+
+            // if any of the elements are double counted
+            if (running_sum > num_elems)
+            {
+                return true;
+            }
+
+            return false;
+        }
+    }
```
siramok, line 2686: With the above change, this wouldn't need to be in an `else` clause anymore, and could be in the function body itself (further reducing indentation). This should be okay because you already have early returns everywhere that matter.

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -3450,6 +3539,91 @@ to_uni_buffer_by_material(const conduit::Node &src_matset,
                   "converting from a material-dominant uni-buffer material set/field is unsupported.");
 }
 
+//-----------------------------------------------------------------------------
+void
+create_field_matset_values_from_unmixed_matset(const conduit::Node &matset,
+                                               conduit::Node &field,
+                                               const float64 epsilon)
+{
+    // extra seat belt here
```
siramok, line 3549: I like seat belts

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -3450,6 +3539,91 @@ to_uni_buffer_by_material(const conduit::Node &src_matset,
...
+        {
+            // unsupported uni-buffer by material
+            CONDUIT_ERROR("conduit::blueprint::mesh::matset::create_field_matset_values_from_unmixed_matset() "
+                          "material-dominant uni-buffer material set is unsupported.");
+        }
+    }
+    else // multi-buffer
+    {
+        if (matset::is_element_dominant(matset))
+        {
```
siramok, line 3577: Same feedback.

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -3450,6 +3539,91 @@ to_uni_buffer_by_material(const conduit::Node &src_matset,
...
+                    if (1.0 - epsilon < vol_frac && vol_frac < 1.0 + epsilon)
+                    {
+                        mset_vals_for_mat.set(elem_id, field_values[elem_id]);
+                    }
+                }
+            }
+        }
+        else // material-dominant
+        {
+            // sparse_by_material
```
siramok, line 3603: Same feedback.

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
```diff
@@ -3450,6 +3539,91 @@ to_uni_buffer_by_material(const conduit::Node &src_matset,
...
+                index_t_accessor mat_elem_ids_vals = elem_ids[matname].value();
+                const index_t num_elems_for_mat = mat_elem_ids_vals.number_of_elements();
+                field["matset_values"][matname].set(DataType::float64(num_elems_for_mat));
+                float64_array mset_vals_for_mat = field["matset_values"][matname].value();
+
+                // iterate over element ids
+                for (index_t index_of_elem_id = 0;
+                     index_of_elem_id < num_elems_for_mat;
+                     index_of_elem_id ++)
+                {
```
siramok, line 3619: Not sure if I've seen a for loop organized like this before (but I like it). I'll have to keep it in mind next time I've got a loop with long variable names.

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+        }
+        const Node &mesh_fields = mesh[domain_path]["fields"];
+
+        auto field_itr = mesh_fields.children();
+        while (field_itr.has_next())
+        {
+            const Node &field = field_itr.next();
+            const std::string fieldname = field_itr.name();
+            const bool has_matset      = field.has_child("matset");
+            const bool has_matset_vals = field.has_child("matset_values");
```
siramok:
```suggestion
            const std::string fieldname = field_itr.name();
            const bool has_matset       = field.has_child("matset");
            const bool has_matset_vals  = field.has_child("matset_values");
```
Finishing the job

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+
+            if (has_matset && has_matset_vals)
+            {
+                // Whether we have a note about this field or not,
+                // if we are material dependent on this domain then
+                // we want to mark this field as material dependent.
+                local_fields_material_status[fieldname].set("yes");
+            }
+            else if (! has_matset && ! has_matset_vals)
```
siramok: Or ignore the previous feedback, and I think this minor simplification would be logically equivalent to what you have:
```c++
const bool has_matset_fields = field.has_child("matset") && field.has_child("matset_values");
if (has_matset_fields)
{
...
}
else if (! has_matset_fields)
{
...
}

```

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+            {
+                error = 1;
+                error_oss << "Field " << fieldname << " on domain " << domain_id << 
+                             " was not read correctly from Silo.";
+                break;
+            }
+        }
+
+        // break out of the domain loop if an error was encountered
+        if (error == 1)
```
siramok:
```suggestion
        if (1 == error)
```
Consistency with your other usage

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+        {
+            break;
+        }
+    }
+
+    // parallel error handling & collect global data
+    Node global_collected_fields_material_status;
+#ifdef CONDUIT_RELAY_IO_MPI_ENABLED
+    Node n_local, n_global;
+    n_local.set((int)error);
```
siramok:
```suggestion
    n_local.set(std::static_cast<int>(error));
```
Up to you

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+        {
+            const Node &field_status = field_status_itr.next();
+            const std::string fieldname = field_status_itr.name();
+            
+            // if we already have an entry for this field
+            if (fields_material_status.has_child(fieldname))
+            {
+                // then we choose "yes" over "no"
+                if (field_status.as_string() == "yes" &&
+                    fields_material_status[fieldname].as_string() == "no")
```
siramok:
```suggestion
                if ("yes" == field_status.as_string() &&
                    "no" == fields_material_status[fieldname].as_string())
```
This could use the inverted style that you like

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+    // stored in `fields_material_status`
+
+    // now for the fun part - we must ensure that all of the fields that are globally
+    // material dependent have matset_values.
+
+
+    for (int domain_id = domain_start; domain_id < domain_end; domain_id ++)
+    {
+        // fetch the domain path for this domain_id
+        const std::string domain_path = conduit_fmt::format("domain_{:06d}", domain_id);
```
siramok, line 4290: I saw that you already pay the cost to compute these domain paths up on line 4129. If they're going to be identical (it seems like they are), you could have a `std::vector<std::string> domain_paths` somewhere at the beginning of the function to cache the strings and reuse them here (i.e., avoid paying the cost again).

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+
+        // ensure we have fields on this domain
+        if (! mesh.has_path(domain_path + "/fields"))
+        {
+            continue;
+        }
+
+        // are there any material dependent fields requiring changes on this domain?
+        const bool any_matdep_fields_requiring_changes = [&]() -> bool
+        {
```
siramok, line 4300: This is blowing my mind. I have no serious opinion about when to use a big inline lambda like this. Usually I'd think to have it be a named helper function to improve organization, but I think this is a case where that would make things more complicated for no benefit.

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+            bool changes_needed = false;
+
+            const Node &mesh_fields = mesh[domain_path]["fields"];
+            auto field_itr = mesh_fields.children();
+            while (field_itr.has_next())
+            {
+                const Node &field = field_itr.next();
+                const std::string fieldname = field_itr.name();
+
+                const bool globally_matdep = fields_material_status[fieldname].as_string() == "yes";
```
siramok:
```suggestion
                const bool globally_matdep = "yes" == fields_material_status[fieldname].as_string()";
```
Another opportunity for inverted evaluation

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+                const bool locally_matdep  = field.has_child("matset");
+
+                if (globally_matdep && locally_matdep)
+                {
+                    // nothing to do
+                }
```
siramok, line 4316: I'm not sure if there will one day be something to do in this case, but it might be just as good to have these empty cases exist in-place as comments. I see the argument for expressing intent, but the compiler most likely optimizes these blocks out anyways, so I think having them as comments is the best of both worlds: you continue expressing intent but also avoid writing intentionally dead code.

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+                {
+                    error = 1;
+                    error_oss << "Material dependence could not be established for field "
+                                  << fieldname << " on domain " << domain_id;
+                    return false;
+                }
+                else // (! globally_matdep && ! locally_matdep)
+                {
+                    // nothing to do
+                }
```
siramok, line 4331: See above feedback.

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+            }
+
+            return changes_needed;
+        }();
```
siramok, line 4335: Jeez, I forgot we were still in a lambda.

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+
+        // break out of the domain loop if an error was encountered
+        if (error == 1)
```
siramok:
```suggestion
        if (1 == error)
```
Opportunity to invert the evaluation

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+
+        // If there are not any material dependent fields on this domain, we can skip
+        if (! any_matdep_fields_requiring_changes)
+        {
+            continue;
+        }
+
+        //
+        // If we have made it this far, we have at least one material dependent field
+        // on this domain, so we must now find its matset
```
siramok, line 4351: Man this is complicated.

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+    // material dependent have matset_values.
+
+
+    for (int domain_id = domain_start; domain_id < domain_end; domain_id ++)
+    {
+        // fetch the domain path for this domain_id
+        const std::string domain_path = conduit_fmt::format("domain_{:06d}", domain_id);
+
+        // ensure we have fields on this domain
+        if (! mesh.has_path(domain_path + "/fields"))
```
siramok, line 4293: If you really wanted to get crazy, you could cache these `domain_path + "/fields"` strings also. 👀

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+            continue;
+        }
+
+        //
+        // If we have made it this far, we have at least one material dependent field
+        // on this domain, so we must now find its matset
+        //
+
+        // check for the existence of matsets
+        if (! mesh.has_path(domain_path + "/matsets"))
```
siramok, line 4355: Seems to be the first construction of these, so no caching opportunity. 🐢

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+        {
+            error = 1;
+            error_oss << "Missing matsets for domain " << domain_path;
+            break; // break out of the domain loop if an error was encountered
+        }
+
+        // make sure there is only one matset
+        const Node &mesh_matsets = mesh[domain_path]["matsets"];
+        if (mesh_matsets.number_of_children() != 1)
```
siramok:
```suggestion
        if (1 != mesh_matsets.number_of_children())
```
Opportunity for eval inversion

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+            {
+                // We have already handled the other cases earlier:
+                //
+                // For both
+                //   (globally_matdep && locally_matdep)
+                //   (! globally_matdep && ! locally_matdep)
+                // there is nothing to do, and for
+                //   (! globally_matdep && locally_matdep)
+                // we have already errored.
+            }
```
siramok, line 4414: Same feedback above about leaving this as-is, but as a comment to avoid writing intentionally dead code.

src/libs/relay/conduit_relay_io_silo.cpp
```diff
@@ -4089,6 +4089,355 @@ read_var_attributes(DBfile *dbfile,
...
+        }
+    } // end loop over domains
+
+    // parallel error handling
+#ifdef CONDUIT_RELAY_IO_MPI_ENABLED
+    n_local.set((int)error);
```
siramok:
```suggestion
    n_local.set(std::static_cast<int>(error));
```
Up to you

src/libs/relay/conduit_relay_io_silo.hpp
```diff
@@ -254,6 +254,16 @@ void CONDUIT_RELAY_API read_mesh(const std::string &root_file_path,
                                  const conduit::Node &opts,
                                  conduit::Node &mesh);
 
+//-----------------------------------------------------------------------------
+///
+/// Added to header for testing purposes; not meant to be used publicly
```
siramok, line 259: Without looking any closer than what's presented here, perhaps there's a way to hide this under a `detail::` or something to guarantee that it can't be used publicly? Maybe that conflicts with what `CONDUIT_RELAY_API` does, in which case oh well.

src/libs/relay/conduit_relay_mpi_io_silo.hpp
```diff
@@ -152,6 +152,17 @@ void CONDUIT_RELAY_API read_mesh(const std::string &root_file_path,
                                  conduit::Node &mesh,
                                  MPI_Comm comm);
 
+//-----------------------------------------------------------------------------
+///
+/// Added to header for testing purposes; not meant to be used publicly
```
siramok, line 157: Same feedback as above.

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1254,6 +1254,279 @@ TEST(conduit_relay_io_silo, overlink_specset_rules)
...
+        while (doms_itr.has_next())
+        {
+            Node &dom = doms_itr.next();
+            const Node &matset = dom["matsets"]["matset"];
+
+            if (0 == dom_id || 2 == dom_id)
+            {
+                EXPECT_FALSE(conduit::blueprint::mesh::matset::has_mixed_elements(matset));
+            }
+            else
```
siramok: Sometimes adding a comment to explain the `else` is overly obvious, but in this case I can't figure out what it would mean if `dom_id` were something other than 1 or 2, so I think it would be genuinely helpful.

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1254,6 +1254,279 @@ TEST(conduit_relay_io_silo, overlink_specset_rules)
...
+        {
+            read_opts["matset_style"] = "sparse_by_element";
+        }
+        io::silo::load_mesh(filename, read_opts, load_mesh);
+        EXPECT_TRUE(blueprint::mesh::verify(load_mesh, info));
+
+        // make changes to part_mesh so the diff will pass
+        for (index_t child = 0; child < part_mesh.number_of_children(); child ++)
+        {
+            if (matset_style == "sparse_by_element" && (0 == child || 2 == child))
```
siramok:
```suggestion
            if ("sparse_by_element" == matset_style && (0 == child || 2 == child))
```
Opportunity to invert eval

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1254,6 +1254,279 @@ TEST(conduit_relay_io_silo, overlink_specset_rules)
...
+                Node &n_matset = part_mesh[child]["matsets"]["matset"];
+
+                // cheat a little bit - we don't have these to start
+                n_matset["sizes"].set_external(load_mesh[child]["matsets"]["mesh_matset"]["sizes"]);
+                n_matset["offsets"].set_external(load_mesh[child]["matsets"]["mesh_matset"]["offsets"]);
+            }
+
+            // TODO remove when https://github.com/llnl/conduit/issues/1600
+            // is addressed
+            if (matset_style == "sparse_by_material")
```
siramok:
```suggestion
            if ("sparse_by_material" == matset_style)
```
Opportunity to invert eval

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1254,6 +1254,279 @@ TEST(conduit_relay_io_silo, overlink_specset_rules)
...
+            {
+                // get the matset for this domain
+                Node &n_matset = part_mesh[child]["matsets"]["matset"];
+
+                Node material_map;
+                conduit::blueprint::mesh::matset::create_or_reuse_material_map(n_matset, material_map);
+                n_matset["material_map"].set(material_map);
+
+                if (child == 0)
```
siramok:
```suggestion
                if (1 == child)
```
Opportunity to invert eval

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1254,6 +1254,279 @@ TEST(conduit_relay_io_silo, overlink_specset_rules)
...
+                {
+                    n_matset["volume_fractions"].remove_child("circle_a");
+                    n_matset["volume_fractions"].remove_child("circle_b");
+                    n_matset["element_ids"].remove_child("circle_a");
+                    n_matset["element_ids"].remove_child("circle_b");
+                }
+                else if (child == 1)
```
siramok:
```suggestion
                else if (1 == child)
```
Opportunity to invert eval

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1254,6 +1254,279 @@ TEST(conduit_relay_io_silo, overlink_specset_rules)
...
+                {
+                    n_matset["volume_fractions"].remove_child("circle_b");
+                    n_matset["element_ids"].remove_child("circle_b");
+                }
+                else if (child == 2)
```
siramok:
```suggestion
                else if (2 == child)
```
Opportunity to invert eval

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1254,6 +1254,279 @@ TEST(conduit_relay_io_silo, overlink_specset_rules)
...
+                {
+                    n_matset["volume_fractions"].remove_child("circle_a");
+                    n_matset["volume_fractions"].remove_child("circle_b");
+                    n_matset["element_ids"].remove_child("circle_a");
+                    n_matset["element_ids"].remove_child("circle_b");
+                }
+                else // if (child == 3)
```
siramok:
```suggestion
                else // if (3 == child)
```
Opportunity to invert eval

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1254,6 +1254,279 @@ TEST(conduit_relay_io_silo, overlink_specset_rules)
...
+                {
+                    // nothing to do for child == 3
+                }
+            }
+
+            silo_name_changer("mesh", part_mesh[child]);
```
siramok, line 1401: Nice API name

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1254,6 +1254,279 @@ TEST(conduit_relay_io_silo, overlink_specset_rules)
...
+            const Node &s_curr = s_itr.next();
+
+            EXPECT_TRUE(l_curr.has_path("fields/mesh_importance/matset"));
+            EXPECT_TRUE(l_curr.has_path("fields/mesh_importance/matset_values"));
+            EXPECT_TRUE(l_curr.has_path("fields/mesh_area/matset"));
+            EXPECT_TRUE(l_curr.has_path("fields/mesh_area/matset_values"));
+
+            // we can only do these checks in the SBE case
+            // the other cases will need to rely on the diff
+            if (matset_style == "sparse_by_element")
```
siramok:
```suggestion
            if ("sparse_by_element" == matset_style)
```

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1254,6 +1254,279 @@ TEST(conduit_relay_io_silo, overlink_specset_rules)
...
+                const Node &imp_mset_vals = l_curr["fields"]["mesh_importance"]["matset_values"];
+                const Node &imp_field_vals = l_curr["fields"]["mesh_importance"]["values"];
+                const Node &area_mset_vals = l_curr["fields"]["mesh_area"]["matset_values"];
+                const Node &area_field_vals = l_curr["fields"]["mesh_area"]["values"];
+                if (0 == dom_id || 2 == dom_id)
+                {
+                    EXPECT_FALSE(imp_field_vals.diff(imp_mset_vals, info, CONDUIT_EPSILON, true));
+                    EXPECT_FALSE(area_field_vals.diff(area_mset_vals, info, CONDUIT_EPSILON, true));
+                }
+                else
```
siramok: Same feedback as above about having a brief comment (2-4 words) that explains what this case represents.

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1254,6 +1254,279 @@ TEST(conduit_relay_io_silo, overlink_specset_rules)
...
+
+    // remove unwanted fields
+    venn_mesh["fields"].remove_child("circle_b");
+    venn_mesh["fields"].remove_child("circle_c");
+    venn_mesh["fields"].remove_child("radius_a");
+    venn_mesh["fields"].remove_child("radius_b");
+    venn_mesh["fields"].remove_child("radius_c");
+    venn_mesh["fields"].remove_child("background");
+    venn_mesh["fields"].remove_child("overlap");
+    venn_mesh["fields"].remove_child("mat_check");
```
siramok, line 1472: I see that the prior test does the same, could be a helper if you wanted:
```c++
void
remove_field_names(Node &mesh)
{
    mesh.remove_child("circle_b");
    mesh.remove_child("circle_c");
    mesh.remove_child("radius_a");
    mesh.remove_child("radius_b");
    mesh.remove_child("radius_c");
    mesh.remove_child("background");
    mesh.remove_child("overlap");
    mesh.remove_child("mat_check");
}
```
called like:
```c++
remove_field_names(venn_mesh["fields"]);
```

src/tests/relay/t_relay_io_silo.cpp
```diff
@@ -1254,6 +1254,279 @@ TEST(conduit_relay_io_silo, overlink_specset_rules)
...
+        if (0 == dom_id || 2 == dom_id)
+        {
+            EXPECT_FALSE(conduit::blueprint::mesh::matset::has_mixed_elements(matset));
+            Node &fields = dom["fields"];
+            fields["importance"].remove_child("matset");
+            fields["importance"].remove_child("matset_values");
+            fields["area"].remove_child("matset");
+            fields["area"].remove_child("matset_values");
+        }
+        else
```
siramok: Same feedback as before.

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+
+        // remove unwanted fields
+        venn_mesh["fields"].remove_child("circle_b");
+        venn_mesh["fields"].remove_child("circle_c");
+        venn_mesh["fields"].remove_child("radius_a");
+        venn_mesh["fields"].remove_child("radius_b");
+        venn_mesh["fields"].remove_child("radius_c");
+        venn_mesh["fields"].remove_child("background");
+        venn_mesh["fields"].remove_child("overlap");
+        venn_mesh["fields"].remove_child("mat_check");
```
siramok, line 1809: Third candidate for a helper.

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        while (doms_itr.has_next())
+        {
+            Node &dom = doms_itr.next();
+            const Node &matset = dom["matsets"]["matset"];
+
+            if (0 == dom_id || 2 == dom_id)
+            {
+                EXPECT_FALSE(conduit::blueprint::mesh::matset::has_mixed_elements(matset));
+            }
+            else
```
siramok: Same feedback as before.

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+                EXPECT_TRUE(conduit::blueprint::mesh::matset::has_mixed_elements(matset));
+            }
+
+            dom["fields"].remove_child("original_vertex_ids");
+            dom["fields"].remove_child("original_element_ids");
+
+            dom_id ++;
+        }
+
+        if (par_rank == 0)
```
siramok:
```suggestion
        if (0 == par_rank)
```
Opportunity for eval inversion

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        {
+            part_mesh.remove(3);
+            part_mesh.remove(1);
+        }
+        else if (par_rank == 1)
```
siramok:
```suggestion
        else if (1 == par_rank)
```
Opportunity for eval inversion

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        {
+            part_mesh.remove(2);
+            part_mesh.remove(0);
+        }
+        else
```
siramok, line 1850: Not sure what's going on here, early return if we execute this test with more than 2 MPI ranks?

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        {
+            // cyrus was wrong about 2 mpi ranks.
+            EXPECT_TRUE(false);
```
siramok, line 1853: OOF

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        // {
+        //     std::cout << "rank 0" << std::endl;
+        //     part_mesh.print();
+        // }
+        // MPI_Barrier(comm);
+        // if (par_rank == 1)
+        // {
+        //     std::cout << "rank 1" << std::endl;
+        //     part_mesh.print();
+        // }
```
siramok: ~Expressing intent? If this is old, I'd suggest just cleaning it out of here.~ Ah, for debug. Still probably better to delete if you've already verified that the test works correctly, but dealer's choice.

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        relay::mpi::io::silo::save_mesh(part_mesh, basename, comm);
+
+        MPI_Barrier(comm);
+
+        //
+        // Part 2: Read from Silo and verify that our fields are all mixed
+        //
+
+        Node load_mesh, read_opts, info;
+        if (matset_style == "full")
```
siramok:
```suggestion
        if ("full" == matset_style)
```
Opportunity for eval inversion

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        {
+            read_opts["matset_style"] = "multi_buffer_full";
+        }
+        else if (matset_style == "sparse_by_material")
```
siramok:
```suggestion
        else if ("sparse_by_material" == matset_style)
```
Opportunity for eval inversion

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        {
+            read_opts["matset_style"] = "multi_buffer_by_material";
+        }
+        else // if (matset_style == "sparse_by_element")
```
siramok:
```suggestion
        else // if ("sparse_by_element" == matset_style)
```
Opportunity for eval inversion

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        {
+            read_opts["matset_style"] = "sparse_by_element";
+        }
+        relay::mpi::io::silo::load_mesh(filename, read_opts, load_mesh, comm);
+        EXPECT_TRUE(blueprint::mesh::verify(load_mesh, info));
+
+        if (par_rank == 0)
```
siramok:
```suggestion
        if (0 == par_rank)
```
Feel free to reject all of these btw

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        {
+            std::cout << "rank 0" << std::endl;
+            load_mesh.print();
+        }
+        MPI_Barrier(comm);
+        if (par_rank == 1)
```
siramok:
```suggestion
        if (1 == par_rank)
```
Feel free to reject all of these btw

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        {
+            std::cout << "rank 1" << std::endl;
+            load_mesh.print();
+        }
```
siramok: Like the previous commented-out block I commented on, this looks like a duplicate used for debug. No problem with printing stuff in a test, but worth considering doing to this block whatever you do with that one (commenting it out, removing it, or secret option C leaving it as-is because it's not hurting anything).

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+
+        // make changes to part_mesh so the diff will pass
+        for (index_t child = 0; child < part_mesh.number_of_children(); child ++)
+        {
+            if (matset_style == "sparse_by_element" && par_rank == 0)
```
siramok:
```suggestion
            if ("sparse_by_element" == matset_style && 0 == par_rank)
```
Feel free to reject all of these btw

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+                Node &n_matset = part_mesh[child]["matsets"]["matset"];
+
+                // cheat a little bit - we don't have these to start
+                n_matset["sizes"].set_external(load_mesh[child]["matsets"]["mesh_matset"]["sizes"]);
+                n_matset["offsets"].set_external(load_mesh[child]["matsets"]["mesh_matset"]["offsets"]);
+            }
+
+            // TODO remove when https://github.com/llnl/conduit/issues/1600
+            // is addressed
+            if (matset_style == "sparse_by_material")
```
siramok:
```suggestion
            if ("sparse_by_material" == matset_style)
```
Feel free to reject all of these btw

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+            {
+                // get the matset for this domain
+                Node &n_matset = part_mesh[child]["matsets"]["matset"];
+
+                Node material_map;
+                conduit::blueprint::mesh::matset::create_or_reuse_material_map(n_matset, material_map);
+                n_matset["material_map"].set(material_map);
+
+                if (par_rank == 0 && child == 0) // original domain 0
```
siramok:
```suggestion
                if (0 == par_rank && 0 == child) // original domain 0
```
Feel free to reject all of these btw

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+                {
+                    n_matset["volume_fractions"].remove_child("circle_a");
+                    n_matset["volume_fractions"].remove_child("circle_b");
+                    n_matset["element_ids"].remove_child("circle_a");
+                    n_matset["element_ids"].remove_child("circle_b");
+                }
+                else if (par_rank == 1 && child == 0) // original domain 1
```
siramok:
```suggestion
                else if (1 == par_rank && 0 == child) // original domain 1
```
Feel free to reject all of these btw

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+                {
+                    n_matset["volume_fractions"].remove_child("circle_b");
+                    n_matset["element_ids"].remove_child("circle_b");
+                }
+                else if (par_rank == 0 && child == 1) // original domain 2
```
siramok:
```suggestion
                else if (0 == par_rank && 1 == child) // original domain 2
```
Feel free to reject all of these btw

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+                {
+                    n_matset["volume_fractions"].remove_child("circle_a");
+                    n_matset["volume_fractions"].remove_child("circle_b");
+                    n_matset["element_ids"].remove_child("circle_a");
+                    n_matset["element_ids"].remove_child("circle_b");
+                }
+                else // if (par_rank == 1 && child == 0) // original domain 3
+                {
+                    // nothing to do for child == 3
+                }
```
siramok: Same feedback as elsewhere about keeping this as-is, just as a comment to avoid writing intentionally dead code.

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+            }
+
+            silo_name_changer("mesh", part_mesh[child]);
```
siramok, line 1951: I wonder what other things silo can change?

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        }
+        // rank 0 has original domains 0 and 2 and
+        // rank 1 has original domains 1 and 3.
+        // In the process of I/O, the domain numbers have been scrambled.
+        // 0 (rank 0) => 0 (rank 0)
+        // 2 (rank 0) => 1 (rank 0)
+        // 1 (rank 1) => 2 (rank 1)
+        // 3 (rank 1) => 3 (rank 1)
+        if (par_rank == 0)
```
siramok:
```suggestion
        if (0 == par_rank)
```
Feel free to ignore

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        {
+            // old domain 2
+            part_mesh[1]["state"]["domain_id"].set(1);
+        }
+        else // (par_rank == 1)
```
siramok:
```suggestion
        else // (1 == par_rank)
```
Feel free to ignore

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+            const Node &s_curr = s_itr.next();
+
+            EXPECT_TRUE(l_curr.has_path("fields/mesh_importance/matset"));
+            EXPECT_TRUE(l_curr.has_path("fields/mesh_importance/matset_values"));
+            EXPECT_TRUE(l_curr.has_path("fields/mesh_area/matset"));
+            EXPECT_TRUE(l_curr.has_path("fields/mesh_area/matset_values"));
+
+            // we can only do these checks in the SBE case
+            // the other cases will need to rely on the diff
+            if (matset_style == "sparse_by_element")
```
siramok:
```suggestion
            if ("sparse_by_element" == matset_style)
```
Feel free to ignore

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+            {
+                // check that mset vals are field vals copied for the clean domains
+                const Node &imp_mset_vals = l_curr["fields"]["mesh_importance"]["matset_values"];
+                const Node &imp_field_vals = l_curr["fields"]["mesh_importance"]["values"];
+                const Node &area_mset_vals = l_curr["fields"]["mesh_area"]["matset_values"];
+                const Node &area_field_vals = l_curr["fields"]["mesh_area"]["values"];
+                if (par_rank == 0)
```
siramok:
```suggestion
                if (0 == par_rank)
```
Feel free to ignore

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+                {
+                    // rank 0 has original domains 0 and 2
+                    EXPECT_FALSE(imp_field_vals.diff(imp_mset_vals, info, CONDUIT_EPSILON, true));
+                    EXPECT_FALSE(area_field_vals.diff(area_mset_vals, info, CONDUIT_EPSILON, true));
+                }
+                else
+                {
+                    // rank 1 has original domains 1 and 3
```
siramok, line 2004: I think this comment explains the other `else` statements that I didn't understand.

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+            // if (par_rank == 1)
+            // {
+            //     std::cout << "rank 1" << std::endl;
+            //     if (l_curr.diff(s_curr, info, CONDUIT_EPSILON, true))
+            //     {
+            //         l_curr.print();
+            //         s_curr.print();
+            //         info.print();
+            //     }
+            // }
```
siramok: Same feedback about removing this if you're done debugging the test. Doesn't hurt to keep it, but it's also easy to write something like this again if you really needed it in the future (and maybe you will need it again very shortly for more matset dev work, in which case feel free to ignore me).

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+
+    // remove unwanted fields
+    venn_mesh["fields"].remove_child("circle_b");
+    venn_mesh["fields"].remove_child("circle_c");
+    venn_mesh["fields"].remove_child("radius_a");
+    venn_mesh["fields"].remove_child("radius_b");
+    venn_mesh["fields"].remove_child("radius_c");
+    venn_mesh["fields"].remove_child("background");
+    venn_mesh["fields"].remove_child("overlap");
+    venn_mesh["fields"].remove_child("mat_check");
```
siramok, line 2052: Another candidate for a helper.

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        if (0 == dom_id || 2 == dom_id)
+        {
+            EXPECT_FALSE(conduit::blueprint::mesh::matset::has_mixed_elements(matset));
+            Node &fields = dom["fields"];
+            fields["importance"].remove_child("matset");
+            fields["importance"].remove_child("matset_values");
+            fields["area"].remove_child("matset");
+            fields["area"].remove_child("matset_values");
+        }
+        else
```
siramok: Same comment as before. Seems to be because `mesh::partition` will leave rank 0 with only domains 0 and 2 and rank 1 with only domains 1 and 3.

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        dom["fields"].remove_child("original_vertex_ids");
+        dom["fields"].remove_child("original_element_ids");
+
+        const std::string domain_path = conduit_fmt::format("domain_{:06d}", dom_id);
+        test_mesh[domain_path].set(dom);
+
+        dom_id ++;
+    }
+
+    if (par_rank == 0)
```
siramok:
```suggestion
    if (0 == par_rank)
```
Feel free to ignore

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+    {
+        test_mesh.remove(3);
+        test_mesh.remove(1);
+    }
+    else if (par_rank == 1)
```
siramok:
```suggestion
    else if (1 == par_rank)
```
Feel free to ignore

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+    {
+        test_mesh.remove(2);
+        test_mesh.remove(0);
+    }
+    else
+    {
+        // cyrus was wrong about 2 mpi ranks.
```
siramok, line 2109: lol :)

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+        EXPECT_TRUE(false);
+    }
+
+    MPI_Barrier(comm);
+
+    // rank 0 has original domains 0 and 2 and
+    // rank 1 has original domains 1 and 3.
+
+    // test missing matset assoc
+    if (par_rank == 1)
```
siramok:
```suggestion
    if (1 == par_rank)
```
At this point I can't help but not continue being consistent, but feel free to ignore

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+    {
+        // domain 1
+        test_mesh.child(0)["fields"]["importance"].remove_child("matset");
+    }
+    MPI_Barrier(comm);
+    EXPECT_THROW(relay::mpi::io::silo::honor_material_dependent_fields(0, 4, test_mesh, comm), conduit::Error);
+    // restore matset name
+    if (par_rank == 1)
```
siramok:
```suggestion
    if (1 == par_rank)
```
Feel free to ignore

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+    {
+        test_mesh.child(0)["fields"]["importance"]["matset"].set("matset");
+    }
+
+    // test missing matsets
+    // cache matset for use later
+    Node dom_2_matset;
+    if (par_rank == 0)
```
siramok:
```suggestion
    if (0 == par_rank)
```
Feel free to ignore

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+    {
+        // domain 2
+        dom_2_matset.set(test_mesh.child(1)["matsets"]["matset"]);
+        test_mesh.child(1).remove_child("matsets");
```
siramok, line 2139: Just an organization thing, I haven't been paying attention to if this is done elsewhere so I won't comment further on it if I see it, but this (and other instances) may be nicer as:
```c++
Node &domain_2 = test_mesh.child(1);
dom_2_matset.set(domain_2["matsets"]["matset"]);
domain_2.remove_child("matsets");
```

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+    }
+    MPI_Barrier(comm);
+    EXPECT_THROW(relay::mpi::io::silo::honor_material_dependent_fields(0, 4, test_mesh, comm), conduit::Error);
+    // now restore missing matset
+    if (par_rank == 0)
```
siramok:
```suggestion
    if (0 == par_rank)
```
Just finishing the job, feel free to disregard if you're not liking these

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+    {
+        test_mesh.child(1)["matsets"]["matset"].set(dom_2_matset);
+    }
+
+    // test ambiguous matsets
+    if (par_rank == 0)
```
siramok:
```suggestion
    if (0 == par_rank)
```
Feel free to disregard

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+    {
+        // domain 2
+        test_mesh.child(1)["matsets"]["matset2"].set(dom_2_matset);
+    }
+    MPI_Barrier(comm);
+    EXPECT_THROW(relay::mpi::io::silo::honor_material_dependent_fields(0, 4, test_mesh, comm), conduit::Error);
+    // remove extra matset
+    if (par_rank == 0)
```
siramok: Actually, I won't keep going with the change suggestions, because it occurs to me that for this many uses of the same exact evaluation, why not just do
```c++
const bool is_par_0 = 0 == par_rank;
```
and reuse that everywhere?

src/tests/relay/t_relay_mpi_io_silo.cpp
```diff
@@ -1755,6 +1755,452 @@ TEST(conduit_relay_mpi_io_silo, overlink_specset_rules3)
...
+    {
+        test_mesh.child(1)["matsets"].remove_child("matset2");
+    }
+
+    // test mixed matset error
+    if (par_rank == 0)
```
siramok: I commented on every other instance, so it feels wrong not to write something here... but see my most recent comment for a better simplification anyways.

siramok (commented):
None of these are correctness suggestions (I am not qualified for that), so take them or leave them. LGTM otherwise.

siramok (commented):
None of these are correctness suggestions (I am not qualified for that), so take them or leave them. LGTM otherwise.

src/tests/relay/t_relay_io_silo.cpp
JustinPrivitera: I added a comment.

src/tests/relay/t_relay_io_silo.cpp
JustinPrivitera: Thanks!

src/tests/relay/t_relay_io_silo.cpp
JustinPrivitera: fixed

src/tests/relay/t_relay_io_silo.cpp
JustinPrivitera: I think I will leave this in so it is explicit.

src/tests/relay/t_relay_io_silo.cpp
JustinPrivitera: added a comment

src/tests/relay/t_relay_mpi_io_silo.cpp
JustinPrivitera: hmm while I am doing this multiple times I do feel like the selection of fields is sort of arbitrary...

src/tests/relay/t_relay_mpi_io_silo.cpp
JustinPrivitera: added a comment

src/tests/relay/t_relay_mpi_io_silo.cpp
JustinPrivitera: yes; all (most?) of the MPI tests are set up so that they will bail out if this happens. The comment about cyrus being wrong about ranks is a relic that has been passed town through tests through the years.

src/tests/relay/t_relay_mpi_io_silo.cpp
JustinPrivitera: LOL

src/tests/relay/t_relay_mpi_io_silo.cpp
JustinPrivitera: 🙃

src/tests/relay/t_relay_mpi_io_silo.cpp
JustinPrivitera: I removed the other one :)

src/tests/relay/t_relay_mpi_io_silo.cpp
JustinPrivitera: done

src/tests/relay/t_relay_mpi_io_silo.cpp
JustinPrivitera: `silo_life_changer`

src/tests/relay/t_relay_mpi_io_silo.cpp
JustinPrivitera: I got rid of it

src/tests/relay/t_relay_mpi_io_silo.cpp
JustinPrivitera: ☹️

JustinPrivitera: gotta get the # of code lines up for github stats!

src/tests/relay/t_relay_mpi_io_silo.cpp
JustinPrivitera: added comment

src/libs/relay/conduit_relay_io_silo.hpp
JustinPrivitera: I think I will leave it where it is but I thought about this.

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
JustinPrivitera: I like that for all of the functions that operate on matsets, they split out the four cases cleanly. Maybe we could save on indentation but I think with this approach there are less mistakes.

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
JustinPrivitera: 💺

src/libs/blueprint/conduit_blueprint_mesh_matset_xforms.cpp
JustinPrivitera: I like to give index variables used in for loops what I think are meaningful names, but the effect of that is that for loops become very verbose like this. Glad you like my solution :)

src/libs/relay/conduit_relay_io_silo.cpp
JustinPrivitera: The point of inverted style is that the compiler will throw an error if you use `=` instead of `==`.

For example, 

```c++
// regular
if (var == 5) // compiles
if (var = 5) // compiles because it sets var to 5 and returns 5. This is bad
// inverted
if (5 == var) // compiles
if (5 = var) // does not compile. This is good
```

In this case, I believe `field_status.as_string()` returns a string literal, so either way we put the operands of the `==` statement, we will get a compilation error if we use `=`.

JustinPrivitera: but I do like consistency so I will use your suggestion :)

src/libs/relay/conduit_relay_io_silo.cpp
JustinPrivitera: I write stuff like this because it gets around some of my biggest gripes with C++. To me this makes sense as the most logical way to organize, but I can see how it would be confusing if this pattern is unfamiliar. I can tell you more about my issues with C++ but perhaps that is a topic that is only interesting to me. :)

src/tests/relay/t_relay_mpi_io_silo.cpp
JustinPrivitera: I think we live pretty fast and loose in the MPI tests. But that might be nice for future MPI tests.

src/libs/relay/conduit_relay_io_silo.cpp
JustinPrivitera: 🙃

src/libs/relay/conduit_relay_io_silo.cpp
JustinPrivitera: we have to see what is happening locally, then see what is happening globally, then determine what needs to be done locally based on the global result. It becomes a headache quickly :/

src/libs/relay/conduit_relay_io_silo.cpp
JustinPrivitera: :)

cyrush (approved):
Looks good, thanks for winning this battle.

PR #1662, opened 2026-07-31
Fix Blueprint Docs Typo

JustinPrivitera:


cyrush (approved):
thanks for finding and fixing this!

JustinPrivitera:
codex

```
• The CI failure looks like a Docker image build networking problem, not a Conduit build/test failure.

  The hard failure is here: log.txt:5018

  curl: (28) Failed to connect to launchpad.net port 443 after 268099 ms: Couldn't connect to server

  That happens in Docker build step [5/18], at src/examples/docker/ubuntu/Dockerfile:66, while downloading:

  https://launchpad.net/ubuntu/+source/mpich/4.2.0-5.1/+build/28285882/+files/libmpich12_4.2.0-5.1_amd64.deb

  Likely causes:

  - launchpad.net was unreachable from the CI runner, possibly transient.
  - CI/network policy or routing is blocking launchpad.net:443.
  - The Dockerfile has no retry/fallback around this direct Launchpad download, so one flaky connection fails the whole job.

  Less likely: a bad package URL. A missing file would usually show an HTTP error like 404, not “couldn’t connect to server”.

  The earlier SyntaxWarning messages and invoke-rc.d: policy-rc.d denied execution of start lines look non-fatal. The job dies only when curl times out during the manual
  MPICH .deb download.

```

Issue #1660, opened 2026-07-29
mesh bp verify error with shape_map that uses int64 insteaed of int32


cyrush:
Node::as_int32() const -- DataType int64 at path topologies/rays/elements/shape_map/line does not equal expected DataType int32

JustinPrivitera:
```python
>>> import conduit
>>> mesh = conduit.Node()
>>> conduit.blueprint.mesh.examples.braid("mixed_2d", 2, 2, 0, mesh)
>>> mesh

state: 
  time: 3.1415
  cycle: 100
coordsets: 
  coords: 
    type: "explicit"
    values: 
      x: [-10.0, 10.0, -10.0, 10.0]
      y: [-10.0, -10.0, 10.0, 10.0]
topologies: 
  mesh: 
    type: "unstructured"
    coordset: "coords"
    elements: 
      shape: "mixed"
      shape_map: 
        quad: 9
        tri: 5
      shapes: [5, 5]
      sizes: [3, 3]
      offsets: [0, 3]
      connectivity: [0, 1, 3, ..., 3, 2]
fields: 
  braid: 
    association: "vertex"
    topology: "mesh"
    values: [2.00961335163232, -0.814560002066095, 2.00961402595266, -0.814562706228405]
  vel: 
    association: "vertex"
    topology: "mesh"
    values: 
      u: [-10.0, 10.0, -10.0, 10.0]
      v: [-10.0, -10.0, 10.0, 10.0]

>>> info = conduit.Node()
>>> conduit.blueprint.mesh.verify(mesh, info)
True
>>> mesh["topologies/mesh/elements/shape_map"].remove_child("quad")
>>> mesh["topologies/mesh/elements/shape_map"].remove_child("tri")
>>> mesh["topologies/mesh/elements/shape_map/quad"].set(conduit.DataType.int64(1))
>>> mesh["topologies/mesh/elements/shape_map/quad"] = 9
>>> mesh["topologies/mesh/elements/shape_map/quad"].dtype
dtype('int64')
>>> mesh["topologies/mesh/elements/shape_map/tri"].set(conduit.DataType.int64(1))
>>> mesh["topologies/mesh/elements/shape_map/tri"] = 5
>>> mesh["topologies/mesh/elements/shape_map/tri"].dtype
dtype('int64')
>>> mesh["topologies/mesh/elements/shape_map"]

quad: 9
tri: 5

>>> conduit.blueprint.mesh.verify(mesh, info)
[...]

PR #1663, opened 2026-07-31
Shape map verify tests for int64 and int32 case

JustinPrivitera:
I did these while investigating #1660

It is unlikely to be broken again but it doesn't hurt to test just in case something is changed in the future.


JustinPrivitera:
codex resume 019fd804-4ff5-7820-a039-4bf83762d339

cyrush (approved):
Looks good, thanks!

PR #1665, opened 2026-08-01
Add execution::dispatch for automatically upgrading DataViews to typed views

siramok:
The PR has evolved a lot, but I'll leave this as-is:

After collecting benchmark data for the poster, I was a bit disappointed to see that the execution model didn't always win when comparing serial to serial execution times. GPU execution times were all pretty good, but in particular, serial execution on uniform->rectilinear was ~70% and ~30% worse with the execution model vs without for 64 and 128 dim braid meshes. This prompted me to investigate where so much performance was being lost.

AI noted that every array access (read/write) on a `DataAccessor` has to execute a switch statement and then cast a memory pointer to the correct dtype (e.g. [`element`](`element`), [`set_value_helper`](https://github.com/llnl/conduit/blob/9391258fff93510207310292bbf761dcfab0c6d0/src/libs/conduit/conduit_data_accessor.hpp#L376)). This seems reasonable in the generic case where our `forall` functors don't necessarily know what dtype or stride the underlying buffers being operated on will have at runtime. However, there are common cases  (e.g., our benchmarks) that involve iterating sequentially over buffers of a fixed dtype and no stride. So we end up incurring per-element overhead that pre-execution model doesn't incur.

This PR adds new tools that enable us to avoid incurring that per-element overhead. Although in retrospect, it looks like this limitation was already known about: #1588.

Essentially: `DataAccessor<T>::packed_ptr()` returns a typed pointer when the buffer's dtype is a packed `T` (not strided), and then the `dispatch_array_read/write/read_write/read_and_write` helpers leverage this to invoke a `forall` kernel with either a raw-pointer wrapper (fast) or the accessor itself if the dtype is not a packed `T` (existing behavior). In other words, this does not improve all data access cases, but lets us have a fast path when applicable.

The results speak for themselves (all measured on 1 node of tuolumne, all plotted with log scale):

Pre- and Post- execution model, without this PR, and `options["copy"] = 1` for `mesh::convert` (the result that prompted me to look for optimizations)
mesh_speedup_summary

Pre- and Post- execution model, with this PR, and `options["copy"] = 1` for `mesh::convert` (much better, but includes the time spent copying the converted meshes into the output nodes, which can be 1/2 or more of the measured conversion time)
mesh_speedup_summary

Pre- and Post- execution model, with this PR, and `options["copy"] = 0` for `mesh::convert` (same "work" performed as before,  but excludes the time spent copying the converted meshes into the output nodes)
mesh_speedup_summary

Same configuration as above, but plotted as individual conversion speedups instead of the overall average. We now see serial outperforming GPU for uniform->rectilinear, which makes much more sense.
mesh_speedup_vs_serial_pre

Note: it's not for this PR, but it may be worth investigating why `mesh::convert` with `options["copy"] = 1` appears to be slower with the execution model vs without.


siramok:
Since the initial description above was written, this PR has evolved quite a bit. Summary of what changed:
- Accessors no longer need to match the underlying dtype to get the performance benefit. Previously, a `float64_accessor` over compact `float32` data still had slow accesses because the dtypes didn't match exactly. Now any accessor can get fast accesses as long as the underlying data is compact and of a supported dtype. Strided and non-numeric data falls back to the accessor, which preserves the existing behavior.
- The API moved around and things were renamed. While getting generic dtype matchups working, I significantly reworked the design to be as clear, sensible, and painless to integrate with our existing work as possible.
- Tests moved into their own file, since this PR provides a new abstraction on top of `execution::` rather than changing it. The tests prove that users can pass any type of array into existing `forall`s without issues: fast accesses are made available automatically when the right conditions are met, and existing behavior is unchanged when they aren't.
- I found additional places throughout the code that benefit from this abstraction and converted them.

Comparing to the initial commit (measured on tuo, exact same test methodology):
mesh_speedup_summary
mesh_speedup_vs_serial_pre

siramok:
Just writing some thoughts down:
- ~I currently call this thing an "array view" but don't necessarily like that name very much.~
    - I decided to go with "typed accessor".
- ~There are currently separate APIs that wrap the `forall` kernels depending on whether we want an array view for direct reading, writing, or both read/write. It occurred to me that we don't gain much value from having the extra granularity, and that the direct read/write wrapper should be able to cover the other use cases. Assuming that's true, I'm envisioning a very nice simplification that becomes possible after removing the other wrappers (i.e., having exactly one external wrapper for getting an array view).~
    - This worked out great, and the PR is now ~500 LoC less.
- ~I'd like to measure how much worse compile times get with vs without this PR (with the existing amount of instances) to evaluate if it makes sense to gate this functionality behind a cmake option.~
    - I measured ~2% slower compile time overall on my local PC, so I don't think we need to be overly concerned.
- ~Justin wondered if an array view could be used internally to improve `DataAccessor`'s `set/fill` performance (the necessary conditions would have to be met of course). My answer was "maybe yes", but it needs investigating. I've already experimented with replacing `.set(*)` with a parallel `forall` and found it to perform worse, but I don't think I tested it with an array view (which certainly would be faster if it works). Thinking about using parallel `forall`s to replace `.set(*)` in general, my intuition says that this would only be beneficial for especially large amounts of values (>100K). In that case, it would be easy to use array size as a heuristic for deciding whether to `.set(*)` normally or do a parallel `.set(*)`.~
    - Addressed in #1671.

- ~I'd like to play around to see if it's possible to support certain classes of strided arrays.~
    - Supporting compact strided arrays (e.g., contiguous groups of [(x0, y0, z0),(x1,y1,z1),...]) should be possible and might be worth it, but I think the PR is in a good place as-is. I added a TODO to investigate in a future PR.

siramok:
This PR has become much more elegant, and it's unlikely that I'll think of another radical way to improve it (I tried making it built-in to our `forall` implementation, but nvcc said not allowed).

I'm pleased with what's here and think it's in a good spot for initial review/feedback. We can schedule a time to go through it together offline.

siramok:
From offline discussion, it was asked whether this would work with `DataArray`s. Furthermore, if the answer turns out to be yes, the natural next question is what happens if a `DataAccessor` and a `DataArray` are both passed to a `with_typed_accessor` lambda? Ideally they could both reduce down to a `RawDataAccessor` (assuming that their access interface is identical) and then not much about the current implementation would need to change. If it does require changes, and if those changes end up increasing compile times by an unbearable amount, we would obviously want to consider alternatives.

siramok:
~We should verify what's happening if `T == U` and see if there is a further optimization to be made to `RawDataAccessor` that avoids casting (note, it's possible that the compiler is already doing this for us, we'd need to measure the difference too to verify).~ See latest comment, Godbolt seems to confirm that we're already getting this optimization.

siramok:
~We want to call it `execution::dispatch`.~ This has been done.

JustinPrivitera:
> We should verify what's happening if `T == U` and see if there is a further optimization to be made to `RawDataAccessor` that avoids casting (note, it's possible that the compiler is already doing this for us, we'd need to measure the difference too to verify).

potentially useful: https://godbolt.org/

siramok:
It seems that we can indeed make `dispatch` work with both `DataAccessor` and `DataArray` without much extra effort. I can even mix and match them in a single `dispatch` like we wondered about, which is neat.

On performance: I expect the benefit for `DataArray` to be smaller than for `DataAccessor`, since `DataArray` doesn't have a per-access dtype switch (we do still save a bit of per-access overhead though, particularly by avoiding stride math). I still haven't benchmarked the difference yet though.

On compile time: since `DataArray` doesn't do type conversion, its type is its type, so we can get away with only doing 2 kernel instantiations. One for the typed accessor and one for the DataArray itself. 

But one thing I discovered while creating the tests: `DataArray<T>` reinterprets the underlying data as `T`. In other words, it doesn't do a `U->T` type conversion like `DataAccessor` does. Nothing prevents you from constructing a `DataArray<float64>` over `float32` data, but iterating over it just returns garbage. I'm not sure how, but maybe it would be worth investigating ways to disallow this?

~(I just need to take a pass over the comments before committing, since they were all geared towards `DataAccessor`.)~ Done.

siramok:
Godbolt seems to confirm that both gcc and clang optimize the casts away for us when `T == U`. The following can be pasted there to verify that both variants compile to identical assembly:
```c++
#include <cstdint>
using index_t = std::int64_t;

// What we have in the PR
template <typename T, typename U>
struct RawDataAccessor
{
    U *ptr;

    T operator[](index_t idx) const {
        return static_cast<T>(ptr[idx]);
    }

    void set(index_t idx, T value) const {
        ptr[idx] = static_cast<U>(value);
    }
};

// A variant without any casting
template <typename T>
struct NoCastAccessor
{
    T *ptr;

    T operator[](index_t idx) const {
        return ptr[idx];
    }

    void set(index_t idx, T value) const {
        ptr[idx] = value;
    }
};

// Scaling kernel with casting
void scale_with_cast(RawDataAccessor<double, double> src,
                     RawDataAccessor<double, double> dst,
                     index_t n)
{
    for (index_t i = 0; i < n; i++)
    {
        dst.set(i, 2.0 * src[i]);
    }
}

// Scaling kernel without casting
void scale_without_cast(NoCastAccessor<double> src,
                        NoCastAccessor<double> dst,
                        index_t n)
{
    for (index_t i = 0; i < n; i++)
    {
        dst.set(i, 2.0 * src[i]);
    }
}
```

siramok:
All feedback has been addressed, changes are pushed.

siramok:
Polished up the implementation by getting rid of the two big switch statements (if we have to add more cases in the future, we won't have to add more big switch statements). Added tests for all dtypes, and made the host/device tests more comprehensive.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -8484,52 +8573,24 @@ void polyhedral_face_centers_normals(conduit::execution::ExecutionPolicy exec_po
...
-        }
-    }
-    if(!handled)
+
+    // Read the coordinates through typed raw pointers to avoid the accessors'
+    // per-element dtype dispatch.
+    conduit::execution::dispatch(n_x.as_double_accessor(),
+                                 n_y.as_double_accessor(),
+                                 n_z.as_double_accessor(),
+                                 [&](auto x, auto y, auto z)
```
JustinPrivitera, line 8582: this is awesome

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
```diff
@@ -2162,32 +2163,34 @@ TopologyMetadata::Implementation::build_edge_key_to_id(
...
-        // Make a key for this edge.
-        index_t edge[2];
-        edge[0] = conn1D[edge_index * 2];
-        edge[1] = conn1D[edge_index * 2 + 1];
-        if(edge[1] > edge[0])
-            std::swap(edge[0], edge[1]);
-        uint64 key = conduit::utils::hash(edge, 2);
-        // Store the edge in the map.
-        edge_key_to_id[edge_index] = std::make_pair(key, edge_index);
+        conduit::execution::forall(policy, 0, nedges, [&](index_t edge_index)
```
JustinPrivitera, line 2169: should this lambda be device-decorated?

src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
```diff
@@ -4695,72 +4696,81 @@ PointQuery::normalSearch(int ndims,
...
+                                     coords[2]->as_double_accessor(),
+                                     [&](auto x, auto y, auto z)
         {
-            const double *searchPt = &input_ptr[i * 3];
-            int found = NotFound;
-            const auto x = coords[0]->as_double_accessor();
-            const auto y = coords[1]->as_double_accessor();
-            const auto z = coords[2]->as_double_accessor();
-            for(conduit::index_t ptid = 0; ptid < numCoordsetPts; ptid++)
+            conduit::execution::forall(policy, 0, numInputPts, [&](conduit::index_t i)
```
JustinPrivitera, line 4704: should this lambda be device-decorated?

src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
```diff
@@ -4695,72 +4696,81 @@ PointQuery::normalSearch(int ndims,
...
+        conduit::execution::dispatch(coords[0]->as_double_accessor(),
+                                     coords[1]->as_double_accessor(),
+                                     [&](auto x, auto y)
         {
-            const double *searchPt = &input_ptr[i * 3];
-            int found = NotFound;
-            const auto x = coords[0]->as_double_accessor();
-            const auto y = coords[1]->as_double_accessor();
-            for(conduit::index_t ptid = 0; ptid < numCoordsetPts; ptid++)
+            conduit::execution::forall(policy, 0, numInputPts, [&](conduit::index_t i)
```
JustinPrivitera, line 4732: should this lambda be device decorated?

src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
```diff
@@ -4695,72 +4696,81 @@ PointQuery::normalSearch(int ndims,
...
         conduit::execution::ExecutionPolicy policy = conduit::execution::ExecutionPolicy::host();
-        conduit::execution::forall(policy, 0, numInputPts, [&](conduit::index_t i)
+        conduit::execution::dispatch(coords[0]->as_double_accessor(),
+                                     [&](auto x)
         {
-            const double *searchPt = &input_ptr[i * 3];
-            int found = NotFound;
-            const auto x = coords[0]->as_double_accessor();
-            for(conduit::index_t ptid = 0; ptid < numCoordsetPts; ptid++)
+            conduit::execution::forall(policy, 0, numInputPts, [&](conduit::index_t i)
```
JustinPrivitera, line 4758: should this lambda be device decorated?

src/libs/conduit/conduit_execution_dispatch.hpp
```diff
@@ -0,0 +1,416 @@
...
+//-----------------------------------------------------------------------------
+// A wrapper around a raw pointer that implements the same access interface as
+// DataAccessor<T> and DataArray<T>. This allows kernels to be templated over
+// any of the 3 accessor types. T is the dtype that the kernel consumes while U
+// is the dtype of the underlying data, since Conduit allows the dtype of the
+// underlying data to differ from the dtype of the accessor. Although in the
+// DataArray case, T and U are always the same because DataArrays don't perform
+// type conversion.
+//
+// NOTE: We verified via Godbolt that the compiler optimizes the static_casts
```
JustinPrivitera:
```suggestion
// NOTE: We verified via Godbolt (https://godbolt.org/) that the compiler optimizes the static_casts
```

src/libs/conduit/conduit_execution_dispatch.hpp
```diff
@@ -0,0 +1,416 @@
...
+// away when T and U are the same, so there is no further performance to be
+// gained by specializing the RawDataAccessor for that case.
```
JustinPrivitera: Compilers are finicky creatures, and turning on or off optimizations can change things. Are you sure that the casts are _always_ optimized away?

src/libs/conduit/conduit_execution_dispatch.hpp
JustinPrivitera, line 1: @cyrush do you want to take a pass over this file? Decide if you like the names of things and so forth.

JustinPrivitera, line 1: And this works with data arrays as well? Most of the verbiage here is about accessors including in the implementation.

src/tests/conduit/t_conduit_execution_dispatch.cpp
JustinPrivitera, line 1: I wonder if any of these helpers should live in the execution test utils file?

src/tests/conduit/t_conduit_execution_dispatch.cpp
```diff
@@ -0,0 +1,820 @@
...
+TEST(conduit_execution_dispatch, policies)
+{
+    // Because nvcc doesn't support extended generic lambdas
+    run_test_single_policies();
+    run_test_pair_policies();
+    run_test_group_policies();
+}
+
+//-----------------------------------------------------------------------------
+int main(int argc, char *argv[])
```
JustinPrivitera: is the main needed at all? There are no arguments to this file

JustinPrivitera (changes requested):
This is really awesome.

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
siramok: Definitely yes, at some point. These lambdas were already host-only (I believe written by Brad, using the early execution model), so adding dispatch to them nets us some free performance without having to do any real porting work. I agree that the porting work should happen, but perhaps that's out of scope for this PR?

src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
siramok: Same comment as above.

src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
siramok: Same comment as above.

src/libs/blueprint/conduit_blueprint_mesh_utils.cpp
siramok: Same comment as above.

src/libs/conduit/conduit_execution_dispatch.hpp
siramok: I am not sure that they are always optimized away. I can play with optimization flags on the above example (starting by figuring out what their default flags are) and get back to you about this.

src/libs/conduit/conduit_execution_dispatch.hpp
siramok: Yes, it works with both interchangeably or in a mixed configuration (as exercised by the tests and also #1671).

I figured that "accessor" was a decent way to generically refer to both, and then specified `DataArray`/`DataAccessor` when wanting to describe something specific to one or the other. If there's a better name/generic term that encompasses both, I'm happy to migrate everything to that (perhaps a good candidate for @cyrush's consideration).

src/tests/conduit/t_conduit_execution_dispatch.cpp
siramok: I wondered the same. I originally chose not to put the helpers there since they aren't used by the execution test itself, but dispatch does ultimately live in the same execution namespace, so maybe it does make sense. I can do that.

src/tests/conduit/t_conduit_execution_dispatch.cpp
siramok: Probably not needed, probably just copy/pasted from an existing test. I can work on removing it.

src/libs/conduit/conduit_execution_dispatch.hpp
JustinPrivitera: We should brainstorm a new name for these template types.

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
JustinPrivitera: The history here is that Brad had his own forall and then I blew it away when we started the new execution model and ported what he had to use our forall, _before_ we started doing anything on the device.

I would say that if device-decorating the lambdas doesn't break anything then let's just do it in this PR.

siramok: Sounds good, will work on that. :)

src/libs/conduit/conduit_execution_dispatch.hpp
siramok: Cyrus suggested "view", although he didn't find it satisfying.

I just now asked AI and it suggested using `DataInterface` instead of `Accessor` for the template names. For the written comments: "A typed accessor" would become "A typed Data interface" and "kernel dispatches with generic accessors" would become "kernel dispatches with generic Data interfaces". Not bad honestly.

Doing the same for "view":  "A typed accessor" would become "A typed view" and "kernel dispatches with generic accessors" would become "kernel dispatches with generic views". Also not bad.

Another option is to use `DataInterface` for the template names, but refer to them as views in all of the comments/documentation.

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
siramok: Looking more closely, I think each of these instances would take some porting effort beyond adding `CONDUIT_EXEC` to the lambdas. Some of the inputs being operated on are plain arrays allocated in host memory, so we'd need to switch them to use data accessors/arrays and make sure that we allocate them in the desired execution space. That shouldn't be super bad (I can still do it if you'd like), but for this specific instance that I'm commenting under, `edge_key_to_id` is actually a `std::vector<std::pair<uint64, index_t>>` so it would involve a heavier refactor.

src/libs/conduit/conduit_execution_dispatch.hpp
JustinPrivitera: What about `ArrayInterface`?

src/libs/blueprint/conduit_blueprint_mesh_topology_metadata.cpp
JustinPrivitera: Ok, fair enough. We can look to hit these parts of the code later as we continue porting the functions in this file.

src/libs/conduit/conduit_execution_dispatch.hpp
siramok: "A typed array interface" and "kernel dispatches with generic array interfaces"

Also sounds fine, I'll defer to whatever you/Cyrus want

JustinPrivitera: Your choice :)

JustinPrivitera: View is maybe too general... ArrayView? Sadly overloaded term :(

src/tests/conduit/t_conduit_execution_dispatch.cpp
siramok: The helpers have all been moved to the execution test utils file.

src/tests/conduit/t_conduit_execution_dispatch.cpp
siramok: I removed it and verified that everything still compiles and that all tests still pass (on tuo).

src/tests/conduit/t_conduit_execution_dispatch.cpp
JustinPrivitera: Thanks!

src/tests/conduit/t_conduit_execution_dispatch.cpp
JustinPrivitera: Thanks!

src/libs/conduit/conduit_execution_dispatch.hpp
siramok: I played around some more with the code example I shared above. When asking for no compiler optimizations via `-O0`, it looks like gcc/clang/rocmcc still generates identical assembly for both the cast vs no cast accessors. I made sure to also test the same compiler versions as what tuo provides (gcc 13.3 and clang 22) and got the same results there too.

I also found a [stack overflow post](https://stackoverflow.com/questions/5814406/does-a-static-cast-to-the-same-primitive-type-produce-any-code) about this where the consensus was that any decent compiler (circa 2011) will treat `static_cast<T>(U)` as a no-op if `T == U` and not generate any extra code to handle this (appears to be confirmed by my experimenting).

src/libs/conduit/conduit_execution_dispatch.hpp
siramok: I think I called this an array view in one of the early iterations of the PR lol.

The only issue I have with `ArrayView` or `ArrayInterface` is that it may be similarly confusing as saying "accessor" to generically refer to both. Perhaps not as bad?

There's also `DataView`... I think it's a decent compromise that 1) references to the common part in `DataArray`/`DataAccessor` and 2) still somewhat describes what the underlying thing is (a view).

src/libs/conduit/conduit_execution_dispatch.hpp
JustinPrivitera: great, thanks for tracking this down all the way.

src/libs/conduit/conduit_execution_dispatch.hpp
JustinPrivitera: Yeah, naming is hard :/ Sorry to give you the run around.

I think `DataView` is good.

siramok: I am glad that you agree, because I had already spent the last few hours making the change! :) I took a pass over all of the naming/verbiage to make things clearer. `Accessor` -> `DataView` and `acc` -> `view` and `RawDataAccessor` -> `TypedDataView` etc.

src/libs/conduit/conduit_execution_dispatch.hpp
```diff
@@ -0,0 +1,450 @@
...
+    if (dt.stride() != width)
+    {
+        return false;
+    }
+
+    // Aligned means that the memory address of the first element is a multiple
+    // of the dtype's width.
+    uintptr_t address = reinterpret_cast<uintptr_t>(view.element_ptr(0));
+    return address % static_cast<uintptr_t>(width) == 0;
+}
```
siramok: The most recent commits mainly renamed stuff. This was the only change of substance, so I felt that highlighting it separately would be the correct thing to do.

I asked claude to take a final look at the implementation within this file and point out any potential flaws or issues. It crafted two different examples (one via `to_compact` and another via `set_external`) of inputs with misaligned memory addresses that would previously be upgraded to `TypedDataViews`.

Conduit is flexible enough to let someone misalign their data, but the issue is that trying to dereference misaligned memory is undefined behavior. Looking into it more, it seems that x86 will let you do it anyways (usually with a performance penalty), while GPUs throw an exception if you try. Being "upgradeable" means that it's safe to treat a `DataView` as a regular raw pointer, so that doesn't seem like it should pass in this case (and I would hope it's rare in practice anyways).

The change is that this helper now checks for both native striding and aligned memory (i.e., is the address a multiple of the dtype width) before returning true.

JustinPrivitera: Nice attention to detail.

JustinPrivitera (approved):

siramok:
We discussed this in person: given that the next PR is adding some new dispatch overloads, I went ahead and reordered functions/trimmed a few comments in this base PR so that adding new overloads to dispatch.hpp remains graceful into the future. No new functionality was added.

siramok:
It occurred to me that this could probably use at least a paragraph in the existing execution model docs.

JustinPrivitera (changes requested):
Let's merge after you add the aforementioned paragraph to the docs.

JustinPrivitera (approved):

PR #1670, opened 2026-08-08
Update `venn` example

### LINKED ISSUE #1600: venn example produces empty background for large radii
see https://github.com/visit-dav/visit/issues/20950

venn background should have an array, even if the radius is strange

JustinPrivitera:
Resolves #1600

See changelog for updates
I also updated the docs to add another couple tips for using matsets

### DIFF
--- a/src/docs/sphinx/blueprint_mesh.rst
+++ b/src/docs/sphinx/blueprint_mesh.rst
@@ -887,6 +887,7 @@ Several important details about Conduit Blueprint material sets are collected he
    There are no restrictions on them other than that they are unique.
  * In the case of **multi-buffer** material sets, where the ``volume_fractions`` node is an *Object* for which material names are the names of the children, if a ``material_map`` is included, the order of material names in the ``volume_fractions``  and in the ``material_map`` need not be the same.
    The ``volume_fractions`` may have less children than the ``material_map``, but the ``material_map`` may not have less children than the ``volume_fractions``.
+ * In the case of **multi-buffer** material sets, where the ``volume_fractions`` node is an *Object* for which material names are the names of the children, if a material is part of a material set but is not present on a particular domain, then that material must **not** appear as a child of the ``volume_fractions`` or ``element_ids``, and instead must appear under the ``material_map``. Therefore, it is recommended to include a ``material_map`` when working with multiple domains and using **multi-buffer** material sets, even though they are optional for that case, because not all materials may appear on all domains.
  * For **uni-buffer** material sets, the **o2mrelation** rules apply.
    Therefore, a valid **uni-buffer** material set may have
 
--- a/src/docs/sphinx/blueprint_mesh.rst
+++ b/src/docs/sphinx/blueprint_mesh.rst
@@ -2230,9 +2230,24 @@ The ``venn()`` function creates meshes that use three overlapping circle regions
 .. code:: cpp
 
     conduit::blueprint::mesh::examples::venn(const std::string &matset_type,
-                                             index_t nx,
-                                             index_t ny,
-                                             float64 radius,
+                                             const index_t nx,
+                                             const index_t ny,
+                                             const float64 radius,
+                                             Node &res);
+
+    conduit::blueprint::mesh::examples::venn(const std::string &matset_type,
+                                             const index_t nx,
+                                             const index_t ny,
+                                             const float64 radius,
+                                             const std::string &generate_material_map,
+                                             Node &res);
+
+    conduit::blueprint::mesh::examples::venn(const std::string &matset_type,
+                                             const index_t nx,
+                                             const index_t ny,
+                                             const float64 radius,
+                                             const std::string &generate_material_map,
+                                             const std::string &generate_specset,
                                              Node &res);
 
 
@@ -2259,6 +2274,46 @@ Here is a list of valid strings for the ``matset_type`` argument:
 ``nx``, ``ny`` specify the number of elements in the x and y directions.
 
 ``radius`` specifies the radius of the three circles.
+
+``generate_material_map`` allows users to choose if they want a material map to be included.
+
+Here is a list of valid strings for the ``generate_material_map`` argument:
+
+.. list-table::
+   :widths: 10 15
+   :header-rows: 1
+
+   * - **Generate Material Map Option**
+     - **Description**
+
+   * - yes
+     - Include a material map.
+
+   * - no
+     - Do not include a material map. An error may be thrown if a material map is required.
+
+   * - default
+     - Include a material map where required or in the case that not including one would lead to ambiguity.
+
+``generate_specset`` allows users to choose if they want a species set to be included.
+
+Here is a list of valid strings for the ``generate_specset`` argument:
+
+.. list-table::
+   :widths: 10 15
+   :header-rows: 1
+
+   * - **Generate Species Set Option**
+     - **Description**
+
+   * - yes
+     - Include a species set.
+
+   * - no
+     - Do not include a species set.
+
+   * - default
+     - Do not include a species set.
 
 The resulting data is placed the Node ``res``, which is passed in via reference.
 
--- a/src/libs/blueprint/conduit_blueprint_mesh_examples_generate.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh_examples_generate.cpp
@@ -455,23 +455,22 @@ generate(const std::string &example_name,
     }
     else if(example_name == "venn")
     {
-        std::string      matset_type           = detail::fetch_string(opts,example_name,"matset_type");
-        conduit::index_t nx                    = detail::fetch_index_t(opts,example_name,"nx");
-        conduit::index_t ny                    = detail::fetch_index_t(opts,example_name,"ny");
-        conduit::float64 radius                = detail::fetch_float64(opts,example_name,"radius");
-        std::string prefer_no_material_map_str = detail::fetch_string(opts,example_name,"prefer_no_material_map","yes");
-        bool prefer_no_material_map            = "yes" == prefer_no_material_map_str;
-        std::string specsets_on_str            = detail::fetch_string(opts,example_name,"specsets_on","no");
-        bool specsets_on                       = "yes" == specsets_on_str;
+        std::string      matset_type      = detail::fetch_string(opts,example_name,"matset_type");
+        conduit::index_t nx               = detail::fetch_index_t(opts,example_name,"nx");
+        conduit::index_t ny               = detail::fetch_index_t(opts,example_name,"ny");
+        conduit::float64 radius           = detail::fetch_float64(opts,example_name,"radius");
+        std::string generate_material_map = detail::fetch_string(opts,example_name,"generate_material_map","default");
+        std::string generate_specset      = detail::fetch_string(opts,example_name,"generate_specset","default");
 
         // wrap
         // void venn(const std::string &matset_type,
         //           const index_t nx,
         //           const index_t ny,
         //           const float64 radius,
-        //           const bool prefer_no_material_map,
+        //           const std::string &generate_material_map,
+        //           const std::string &generate_specset,
         //           Node &res);
-        venn(matset_type,nx,ny,radius,prefer_no_material_map,specsets_on,res);
+        venn(matset_type,nx,ny,radius,generate_material_map,generate_specset,res);
     }
     else if (example_name == "bent_multi_grid")
     {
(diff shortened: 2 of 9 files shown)


cyrush:
This error means a memory access crime was committed:
```
 unknown file : error : SEH exception with code 0xc0000005 thrown in the test body. [D:\a\conduit\conduit\build\RUN_TESTS.vcxproj]
```

Windows judges these harshly.

JustinPrivitera:
> This error means a memory access crime was committed:
> 
> ```
>  unknown file : error : SEH exception with code 0xc0000005 thrown in the test body. [D:\a\conduit\conduit\build\RUN_TESTS.vcxproj]
> ```
> 
> Windows judges these harshly.

This is very odd. The test that is failing is unrelated to my changes. Perhaps a transient CI failure? I am investigating.

JustinPrivitera:
Windows looks to be passing now. All I did was rerun the job.

src/docs/sphinx/blueprint_mesh.rst
```diff
@@ -887,6 +887,7 @@ Several important details about Conduit Blueprint material sets are collected he
    There are no restrictions on them other than that they are unique.
  * In the case of **multi-buffer** material sets, where the ``volume_fractions`` node is an *Object* for which material names are the names of the children, if a ``material_map`` is included, the order of material names in the ``volume_fractions``  and in the ``material_map`` need not be the same.
    The ``volume_fractions`` may have less children than the ``material_map``, but the ``material_map`` may not have less children than the ``volume_fractions``.
+ * In the case of **multi-buffer** material sets, where the ``volume_fractions`` node is an *Object* for which material names are the names of the children, if a material is part of a material set but is not present on a particular domain, then that material must **not** appear as a child of the ``volume_fractions`` or ``element_ids``, and instead must appear under the ``material_map``. Therefore, it is recommended to include a ``material_map`` when working with multiple domains and using **multi-buffer** material sets, even though they are optional for that case, because not all materials may appear on all domains.
```
siramok, line 890: This explanation made sense to me. The details obviously have to be precise, which makes this complicated.

Just for my curiosity, suppose I were a user and I defined a multi-buffer material set in the incorrect way (with different/unique material names per domain and without a `material_map` to keep it all straight). I know that the docs here say not to do that, but will conduit save me from myself and throw an error explaining what I did wrong/how to correct it?

src/tests/blueprint/t_blueprint_mesh_examples_generate.cpp
```diff
@@ -93,6 +93,69 @@ TEST(conduit_blueprint_mesh_examples_generate, gen_driver_all_examples)
...
+    opts["ny"]     = 100;
+    opts["radius"] = 0.25;
+
+    // test full
+    opts["matset_type"]            = "full";
+    opts["prefer_no_material_map"] = "yes";
+    conduit::blueprint::mesh::examples::generate("venn", opts, res);
+    EXPECT_FALSE(res.has_path("matsets/matset/material_map"));
+
+    opts["prefer_no_material_map"] = "no";
```
siramok: Just a variable naming thing, but having the possible inputs be "yes"/"no" creates an awkward double negative: "Do I want to prefer no material map? No." Makes it hard to know what I am asking for and what behavior I am getting.

And now that I'm dwelling on it, "prefer" implies that regardless of what I ask for, I might not get what I asked for anyways?

src/docs/sphinx/blueprint_mesh.rst
JustinPrivitera: Conduit currently does not perform verification checks like that across multiple domains. So you will see strange bugs and behavior. I don't know that we have a good way to check this case. Maybe we can add something in VisIt when trying to work with material sets.

src/tests/blueprint/t_blueprint_mesh_examples_generate.cpp
JustinPrivitera: This is what I was looking for feedback on. I'm not sure what good names for these choices should be.

The reason it is "prefer" is because under some conditions, you might get a material map whether you want one or not. You can ask for a material map or you can ask to not give a material map if it is not needed, so to me that feels like asking if you prefer to not have one. But it is confusing and it is worth thinking if there is a better way to describe this option.

src/libs/blueprint/conduit_blueprint_mesh_examples_venn.cpp
```diff
@@ -314,46 +319,79 @@ void venn_sparse_by_material_matset(Node &res, index_t nx, index_t ny)
...
+        Node &matset_importance_bg = res["fields/importance/matset_values/background"];
+        matset_importance_bg.set(DataType::float64(bgcount));
+        float64_array matset_importance_bg_value = matset_importance_bg.value();
+
+        index_t nidx = 0;
+        for (index_t idx = 0; idx < elements; ++idx)
         {
-            bg_idx[nidx] = (int32)idx;
+            const float64 x_pos = static_cast<float64>(idx % nx) / nx;
+            const float64 y_pos = static_cast<float64>(idx / nx) / ny;
```
siramok, line 381: Minor optimization: since you only need `x_pos` and y_pos` when `if (fgvf < one_minus_epsilon)` is true, why not move the construction of these variables into that if statement?

src/libs/blueprint/conduit_blueprint_mesh_examples_venn.cpp
```diff
@@ -369,38 +407,30 @@ void venn_sparse_by_material_matset(Node &res, index_t nx, index_t ny)
...
+    {
+        res["fields/mat_check/matset_values/background"].set(DataType::int64(bgcount));
+    }
     res["fields/mat_check/matset_values/circle_a"].set(DataType::int64(cir_a_count));
     res["fields/mat_check/matset_values/circle_b"].set(DataType::int64(cir_b_count));
     res["fields/mat_check/matset_values/circle_c"].set(DataType::int64(cir_c_count));
 
     // set with unique values
 
-    // NOTE: this is good use case for adding DataArray.fill ...
```
siramok, line 379: Hey, I know what that is...

src/libs/blueprint/conduit_blueprint_mesh_examples_venn.cpp
```diff
@@ -529,146 +559,409 @@ void venn_sparse_by_element_matset(Node &res, index_t nx, index_t ny)
...
+        {
+            const index_t elem_id = x + nx * y;
+            const float64 xf = static_cast<float64>(x);
 
-    // per element how many circles overlap
-    res["fields/overlap/association"] = "element";
-    res["fields/overlap/topology"] = "topo";
-    res["fields/overlap/values"] = DataType::float64(nx * ny);
+            // circle_a species vary horizontally
+            float64 specvalue = xf / nxf;
```
siramok, line 622: Since `nxf` and `nyf` never change and are only used as divisors within this loop, you could instead define:
```c++
// i for inverse, but call it whatever you want
const float64 inxf = 1. / static_cast<float64>(nx);
const float64 inyf = 1./ static_cast<float64>(ny);
```

Then all of these repeat divisions can become repeat multiplies (faster):
```c++
float64 specvalue = xf * inxf;
...
float64 specvalue = yf * inyf;
...
specvalue = (yf * inyf + xf * inxf) * 0.5;
```

Although perhaps the compiler is already smart enough to make this optimization on your behalf.

src/libs/blueprint/conduit_blueprint_mesh_examples_venn.cpp
```diff
@@ -529,146 +559,409 @@ void venn_sparse_by_element_matset(Node &res, index_t nx, index_t ny)
...
+                // circle_b species vary vertically
+                const float64 specvalue = yf / nyf;
+                cb_spec1.push_back(specvalue);
+                cb_spec2.push_back(1.0 - specvalue);
+            }
 
+            if (contains(element_ids["circle_c"], elem_id))
+            {
+                // circle_c species vary diagonally
+                const float64 specvalue = (yf / nyf + xf / nxf) / 2.0;
```
siramok, line 691: Same comment/feedback as above.

src/libs/blueprint/conduit_blueprint_mesh_examples_venn.cpp
```diff
@@ -529,146 +559,409 @@ void venn_sparse_by_element_matset(Node &res, index_t nx, index_t ny)
...
+                    // circle_b species vary vertically
+                    const float64 specvalue = yf / nyf;
+                    matset_values.push_back(specvalue);
+                    matset_values.push_back(1.0 - specvalue);
+                    num_specs_in_zone += 2;
+                }
+                else if ("circle_c" == matname)
+                {
+                    // circle_c species vary diagonally
+                    const float64 specvalue = (yf / nyf + xf / nxf) / 2.0;
```
siramok, line 788: Same comment/feedback as above.

src/libs/blueprint/conduit_blueprint_mesh_examples_venn.cpp
```diff
@@ -529,146 +559,409 @@ void venn_sparse_by_element_matset(Node &res, index_t nx, index_t ny)
...
+        x_coords[i] = vx;
+        vx+=dx;
+    }
+    
+    float64 vy = 0;
+    for(index_t i =0; i< ny+1; i++)
+    {
+        y_coords[i] = vy;
+        vy+=dy;
+    }
```
siramok, line 878: I'm not sure how common it is for `nx == ny` in practice, but I can't help but notice that this work could be done using a single loop if `nx == ny`. Although I'm not sure if these numbers ever get large enough for that optimization to matter.

src/libs/blueprint/conduit_blueprint_mesh_examples_venn.cpp
```diff
@@ -529,146 +559,409 @@ void venn_sparse_by_element_matset(Node &res, index_t nx, index_t ny)
...
+    
+    res["topologies/topo/type"] = "rectilinear";
+    res["topologies/topo/coordset"] = "coords";
+
+    // create the fields
+
+    // circle a distance field
+    res["fields/radius_a/association"] = "element";
+    res["fields/radius_a/topology"] = "topo";
+    res["fields/radius_a/values"] = DataType::float64(nx * ny);
```
siramok, line 890: The compiler most likely does optimize this one away for you, but all of these instances of `nx * ny` could be computed once up above as `const index_t num_values = nx * ny;` and then reused.

siramok (commented):
The diff made it challenging to grasp the full context, so I only commented on localized things I noticed.

src/tests/blueprint/t_blueprint_mesh_examples_generate.cpp
siramok: Naming is hard. `prefer_material_map` is the first thing that comes to mind:
```c++
opts["prefer_material_map"] = "yes";
...
if (! prefer_material_map)
```
Keeps what you have while getting rid of the double negatives, but is maybe still confusing. These don't really seem better, but in case it helps strike inspiration:
`try_generate_material_map`
`try_material_map`
`generate_material_map_if_possible`

JustinPrivitera:
As discussed offline, the diff is broken and some of the feedback given here is aimed at code that was not edited. That code could maybe use some updates, but those updates are outside of the scope of this work.

siramok:
Yes, feel free to resolve/close comments that not applicable to this PR.

cyrush:
Now I don't recall which test was failing in windows, but you are sure it wasn't related?
Usually fire when we see smoke like that.

JustinPrivitera:
> Now I don't recall which test was failing in windows, but you are sure it wasn't related?
> Usually fire when we see smoke like that.

My analysis indicated that the place where things failed was a Silo test that did not leverage the `venn` example. The only thing that could have been touched my it that I changed is my slight tweaks to `verify`, but you can see for yourself that my changes are very minor and are unlikely to cause a problem like what we saw.

If re-running CI caused the failure to disappear, then what do you suggest? I can keep my eye out for this as we continue opening PRs.

cyrush:
@JustinPrivitera thanks for the info, given this I think we are ok to move forward.

src/libs/blueprint/conduit_blueprint_mesh_examples_generate.cpp
```diff
@@ -434,18 +455,23 @@ generate(const std::string &example_name,
...
-        conduit::index_t ny          = detail::fetch_index_t(opts,example_name,"ny");
-        conduit::float64 radius      = detail::fetch_float64(opts,example_name,"radius");
+        std::string      matset_type           = detail::fetch_string(opts,example_name,"matset_type");
+        conduit::index_t nx                    = detail::fetch_index_t(opts,example_name,"nx");
+        conduit::index_t ny                    = detail::fetch_index_t(opts,example_name,"ny");
+        conduit::float64 radius                = detail::fetch_float64(opts,example_name,"radius");
+        std::string prefer_no_material_map_str = detail::fetch_string(opts,example_name,"prefer_no_material_map","yes");
+        bool prefer_no_material_map            = "yes" == prefer_no_material_map_str;
+        std::string specsets_on_str            = detail::fetch_string(opts,example_name,"specsets_on","no");
+        bool specsets_on                       = "yes" == specsets_on_str;
```
JustinPrivitera: curious for feedback here @cyrush

cyrush: What about `generate_material_map = {yes, no, auto}`

if yes and it can't be done - error
if auto, generated if needed
if no and it will be done (is this even a case?) -  error

JustinPrivitera:
@cyrush this should be ready to go now, assuming it passes CI.

PR #1671, opened 2026-08-08
Port DataAccessor and DataArray APIs to the execution model

siramok:
Taking inspiration from the stuff learned in #1665 (but not dependent on it), this PR optimizes the bulk set and fill operations in `DataAccessor` and `DataArray` by evaluating the dtype exactly once per operation as opposed to once per access.


siramok:
This PR optimizes the existing implementation of bulk set/fill, but we had an offline discussion about replacing these implementations outright with `execution::forall` versions instead.

What we'd hope to gain from going that route:

- The bulk set/fill operations would work with device-resident data (they currently don't).
- While we're in there, we could give count, min, max, and mean the same `forall` treatment.
- We may see better performance via parallelization.

Things to consider:

- We'd have to enable TU compilation for the entire `DataAccessor` and `DataArray` source files.
    - Past experience says that this may blow up compilation times beyond what we are willing to bear.
    - It may require fixes/changes to how we currently implement things to satisfy CUDA/HIP, although Cyrus made a good point that we successfully pulled it off for `conduit_blueprint_mesh.cpp` which is considerably more complicated.
- We would want to leverage #1665 to ensure that we're still getting fast accesses whenever possible.
- We may see worse performance via parallelization (at least in some cases, like in serial execution or small data sizes). I can think of at least 2 ways to mitigate this:
    - We could simply keep the current implementation around for serial execution only.
    - If we find that other backends become slower below certain data sizes, the current implementation could be the fallback if we do a runtime `.size()` check and decide that it would be better to avoid parallelism if the number of elements being operated on are less than an arbitrary threshold (we could do some empirical testing to find the sweet spot).  
- It's currently unclear how we would handle/guard against the mixing of host vs device memory for bulk set. For self-contained operations like bulk fill, we agreed that we could use the `active_space()` API to select an execution policy based on where the data already lives. This assumes that the user has already called `.use_with()` to move the data where they want it to be before calling `.fill()`, which seems reasonable to me. With bulk set, `accessor_a`'s data may live on device but the input `accessor_b`'s data could live on the host, which creates a problem for how we would handle `accessor_a.set(accessor_b)`. We probably don't want to modify `accessor_b`'s data to fulfill the bulk set (i.e., silently move it to the correct memory space via `.use_with()`). We could just choose to throw an error if each accessor returns something different for `active_space()`, but that doesn't seem satisfying. Maybe we could use `#if/#else/#endif` to implement these cases using `cudaMemcpy` or `hipMemcpy` depending on which runtime conduit was compiled with, or maybe Umpire exposes something like this already? For the sake of not letting this piece block the whole idea, we could make it an error for the time being and leave a TODO to implement it better later on.

siramok:
For the sake of communication, this PR is going to become about porting the underlying DataArray and DataAccessor APIs to the execution model (dispatch + forall), which does still happen to optimize bulk fill and set.

The initial version is largely finished, I just want to do more polishing and testing. It depends on the dispatch PR anyways, so that one should come first. Compile times did initially blow up to >20 extra minutes when using the full power of dispatch, but I've gotten it down closer to 3 extra minutes by being selectively conservative with it.

It includes all of the other APIs that could be ported as well, i.e. count, mean, min, max, etc.

Performance is generally the same or slightly better in the benchmarks on average, but `uniform_to_structured` in particular improved by a large amount.

src/libs/conduit/conduit_data_accessor.hpp
```diff
@@ -40,6 +40,9 @@ namespace execution
 {
     class ExecutionPolicy;
 }
+// Gives the kernels in conduit_data_kernels.hpp access to the private
+// set_value_helper()
+struct DataAccessorKernelAccess;
```
JustinPrivitera: we don't need this in data array hpp?

src/libs/conduit/conduit_data_accessor.hpp
```diff
@@ -365,6 +368,14 @@ class CONDUIT_API DataAccessor
...
 private:
 
+//-----------------------------------------------------------------------------
+// Kernel friends
+//-----------------------------------------------------------------------------
+    // The kernels access set_value_helper() through this because the set()
+    // overloads are ambiguous for c-native types that don't alias a bitwidth
+    // type, e.g. long long.
+    friend struct conduit::DataAccessorKernelAccess;
+
```
JustinPrivitera: we don't need this in data array hpp?

src/libs/conduit/conduit_data_kernels.hpp
```diff
@@ -0,0 +1,442 @@
...
+
+//-----------------------------------------------------------------------------
+// These kernels use functor structs rather than lambdas for two reasons:
+// First, the structs seem to compile a bit faster under nvcc. Second, this
+// header is included by conduit_data_array and conduit_data_accessor, so that
+// these kernel instantiations can be reused in both. That reuse is unsafe
+// with lambdas because nvcc registers extended lambdas per translation unit,
+// so a deduplicated lambda instantiation can end up referencing a discarded
+// translation unit's registration table and crash at runtime.
+//-----------------------------------------------------------------------------
```
JustinPrivitera: :(

src/libs/conduit/conduit_data_kernels.hpp
```diff
@@ -0,0 +1,442 @@
...
+
```
JustinPrivitera: In general I wish the functors lived next to their use to make this section easier to understand.

src/libs/conduit/conduit_data_kernels.hpp
```diff
@@ -0,0 +1,442 @@
...
+//-----------------------------------------------------------------------------
+template <typename T, typename U>
+struct FillOp
+{
```
JustinPrivitera: What is FillOp for? It looks like you have cases for FillArrayOp and FillAccessorOp already

src/libs/conduit/conduit_data_kernels.hpp
```diff
@@ -0,0 +1,442 @@
...
+    CONDUIT_EXEC void operator()(index_t i) const
+    {
+        DataAccessorKernelAccess::set(vals, i, value);
+    }
+};
+
+//-----------------------------------------------------------------------------
+template <typename T, typename U>
+struct CopyFromPtrOp
+{
```
JustinPrivitera: Again, what is the third case used for? We have array and accessor cases as well.

src/libs/conduit/conduit_data_kernels.hpp
```diff
@@ -0,0 +1,442 @@
...
+
+    CONDUIT_EXEC void operator()(index_t i) const
+    {
+        vals.element(i) = static_cast<T>(src[i]);
+    }
+};
+
+//-----------------------------------------------------------------------------
+template <typename T, typename U>
+struct CopyFromAccAccessorOp
```
JustinPrivitera: is this name meant to indicate that we are copying from a data accessor to a data accessor? Or does `Acc` mean something more general?

src/libs/conduit/conduit_data_kernels.hpp
```diff
@@ -0,0 +1,442 @@
...
+// These kernels use functor structs rather than lambdas for two reasons:
+// First, the structs seem to compile a bit faster under nvcc. Second, this
+// header is included by conduit_data_array and conduit_data_accessor, so that
+// these kernel instantiations can be reused in both. That reuse is unsafe
+// with lambdas because nvcc registers extended lambdas per translation unit,
+// so a deduplicated lambda instantiation can end up referencing a discarded
+// translation unit's registration table and crash at runtime.
+//-----------------------------------------------------------------------------
+
+//-----------------------------------------------------------------------------
```
JustinPrivitera: I get that you wanted data array and data accessor to share implementation. I have felt that way in the past as well.

However, I am not sure that putting the implementation in this file in this manner is better than having separate implementations that are easier to understand/more localized.

My question is, how much of this infrastructure would need to be duplicated? Do we even need to add `src/libs/conduit/conduit_data_kernels.hpp` at all? Do we need the functors and helpers here at all? What would change if we no longer intended to reuse the same implementation?

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -15,6 +15,7 @@
 #include <algorithm>
 #include <limits>
 #include <type_traits>
+#include <vector>
```
JustinPrivitera: why do we need vector?

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -38,88 +41,122 @@ namespace detail
...
+    {
+        // This could be implemented, but forcing the caller to decide where
+        // the data should live first keeps the cost of allocating and copying
+        // from becoming an unexpected side-effect of set().
+        CONDUIT_ERROR("DataAccessor::set() cannot copy data to and from "
+                      "different memory spaces. Use use_with() and sync() "
+                      "to ensure that the source and destination data live in "
+                      "the same memory space first.");
+    }
+}
```
JustinPrivitera, line 424: @cyrush should we fail in this case? I think failure is better than what we have right now, but do we want this case to magically work? I think this should be possible using the strided memcpy approach used in `use_with`. @siramok can you investigate that approach? Let me know if you have questions.

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -38,88 +41,122 @@ namespace detail
...
+                  const U *values,
+                  index_t num_elements)
+{
+    if (execution::DeviceMemory::is_device_ptr(values))
+    {
+        CONDUIT_ERROR("DataAccessor::set() cannot convert from a "
+                      "device-resident source of a different type. Use "
+                      "use_with() and sync() to ensure that the source and "
+                      "destination data live in the same memory space first.");
+    }
```
JustinPrivitera: This error is confusing; the if-statement checks if we are dealing with incoming device values and throws, but the error message says the user should make sure their data is in the same memory space. If their source and destination data are both on device, in the same memory space, this error will still trigger because they are doing device.

Why do we need to restrict this to host anyhow? Maybe I haven't read far enough along to see if there is another path for device? If that is the case, please add a comment to this function so that it is clear. If not, then I understand that we can't run this on device because of the `std::vector`, but maybe that means that that is not the right approach? Why shouldn't this be able to work on device?

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -38,88 +41,122 @@ namespace detail
...
+    std::vector<T> converted(num_elements);
+    for (index_t i = 0; i < num_elements; i++)
+    {
+        converted[i] = static_cast<T>(values[i]);
+    }
+    set_values_helper(accessor, converted.data(), num_elements);
+}
+
+//-----------------------------------------------------------------------------
+template <typename U, template <typename> class Accessor, typename T>
```
JustinPrivitera: what is happening here?

why do we need this?

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -38,88 +41,122 @@ namespace detail
...
+void
+set_values_acc_helper(const DataAccessor<T> &accessor,
+                      const Accessor<U> &values,
```
JustinPrivitera: what does an `Accessor` represent here?

Maybe some comments would help.

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -38,88 +41,122 @@ namespace detail
...
+    else // dst and src are in different memory spaces
+    {
+        // This could be implemented, but forcing the caller to decide where
+        // the data should live first keeps the cost of allocating and copying
+        // from becoming an unexpected side-effect of set().
+        CONDUIT_ERROR("DataAccessor::set() cannot copy data to and from "
+                      "different memory spaces. Use use_with() and sync() "
+                      "to ensure that the source and destination data live in "
+                      "the same memory space first.");
+    }
```
JustinPrivitera: same comment here about how I think we can support copying from different memory spaces if we are inspired by the implementation of `use_with`

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -242,35 +279,21 @@ template <typename T>
...
-        {
-            res = val;
-        }
-    }
-
-    return res;
+    const index_t num_elements = number_of_elements();
+    execution::ExecutionPolicy policy = detail::select_policy(active_space(),
+                                                              num_elements);
+    return detail::min_kernel<T>(policy, num_elements, *this);
```
JustinPrivitera: I think I feel a little uncomfortable about grabbing stuff from a `detail` namespace in another TU

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -279,153 +302,47 @@ template <typename T>
...
-        res += val;
-    }
-
-    res = res / float64(number_of_elements());
-    return res;
+    const index_t num_elements = number_of_elements();
+    execution::ExecutionPolicy policy = detail::select_policy(active_space(),
+                                                              num_elements);
+    return detail::mean_kernel<T>(policy, num_elements, *this) /
+           static_cast<float64>(num_elements);
```
JustinPrivitera: does this mean that `mean_kernel` does not actually calculate the mean, but instead calculates the sum?

JustinPrivitera: As an aside, to discover the answer to my question, I had to move through multiple levels of indirection. Not sure what order these comments will appear in but I am wondering what all that indirection is buying us, or if we could simplify things by having separate implementations for data accessor and data array.

JustinPrivitera: Why do we need `mean_kernel` if it only calculates the sum? Why can't we use the sum kernel?

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -473,7 +390,7 @@ DataAccessor<T>::use_with(conduit::execution::ExecutionPolicy policy)
                                                        number_of_elements(),
                                                        dtype().element_bytes(),
                                                        m_other_dtype.stride(),
-                                                       m_data,
+                                                       element_ptr(0),
                                                        dtype().stride());
```
JustinPrivitera, line 843: Why make the change from `m_data` to `element_ptr(0)`?

JustinPrivitera: Also, this is the function I am suggesting you use for the non-shared memory space cases for the other parts of the API where you threw an error.

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -591,12 +508,15 @@ DataAccessor<T>::sync()
...
-                                               m_stride);
+        // data_ptr() is the node's base pointer, so we add the node dtype's
+        // offset to write back to the correct elements
+        utils::conduit_memcpy_strided_elements(
+            static_cast<char*>(m_node_ptr->data_ptr()) + m_node_ptr->dtype().offset(),
+            number_of_elements(),
+            m_node_ptr->dtype().element_bytes(),
+            m_node_ptr->dtype().stride(),
+            element_ptr(0),
+            m_stride);
```
JustinPrivitera: What issues did you run into that suggested that this needed to change?

src/libs/conduit/conduit_data_array.cpp
```diff
@@ -25,18 +25,140 @@
 #include "conduit_log.hpp"
 #include "conduit_data_accessor.hpp"
 #include "conduit_execution.hpp"
+#include "conduit_execution_dispatch.hpp"
+#include "conduit_data_kernels.hpp"
 #include "conduit_annotations.hpp"
 
-// Easier access to the Conduit logging functions
-using namespace conduit::utils;
-
```
JustinPrivitera, line 32: ?

src/libs/conduit/conduit_data_array.cpp
JustinPrivitera, line 1: I have similar questions about data array as I do about data accessor but I have not commented here to avoid repeating myself.

src/tests/conduit/t_conduit_data_accessor.cpp
```diff
@@ -901,3 +906,369 @@ TEST(conduit_data_accessor, set_using_data_accessor)
...
+
+        // An int64 accessor over the strided int8 data
+        int64_accessor acc = n.value();
+
+        acc.fill(42);
+        for (index_t i = 0; i < num_elements; i++)
+        {
+            EXPECT_EQ(buff[2 * i], (int8)-7);
+            EXPECT_EQ(buff[2 * i + 1], (int8)42);
+            EXPECT_EQ(acc[i], (int64)42);
```
JustinPrivitera, line 935: This is cool, but could you maybe add comments above each line to explain what they are checking? I understand after staring but more comments are always better :)

JustinPrivitera (changes requested):
Let's discuss :)

siramok:
@JustinPrivitera Just got to your comments here. I have to apologize, I had intended to iterate on this PR further before asking for your eyes on it. :(

The PR that existed here previously was ready for review (the bulk set/fill without any porting to execution model), but after we decided to go all the way and port things to the execution model, I repurposed this PR instead of making a new one. To avoid further confusion, I'll convert this back to a draft PR for now.

What's here was a result of me pushing the progress from my laptop while traveling so I could resume work on my regular PCs. Some of these changes are suggestions from claude that I haven't finished validating yet, including some of the things that you commented on, so I'll have to get back to you on those.

Your comments are still very helpful and appreciated. :)

JustinPrivitera:
@siramok no worries; it was good to wrap my head around what you've been up to anyhow.

src/libs/conduit/conduit_data_kernels.hpp
siramok: I think this would be a good and reasonable change.

src/libs/conduit/conduit_data_array.cpp
siramok: This is unavoidable to make TU compilation work, we had to do the same thing when porting `conduit_blueprint_mesh.cpp`. The issue is that plain `log()` becomes ambiguous when compiling a file as a TU, since CUDA has the log math function in its default namespace.

src/libs/conduit/conduit_data_accessor.cpp
siramok: This was one of the changes that I didn't get around to deciding is actually valid or not. If I'm already making heavy changes, I tend to ask claude if it sees any mistakes/errors/bugs within an entire file. If they seem like reasonable errors, I ask to see what the fix looks like.

In this instance, it pointed out the previous version of the call would be incorrect if either `m_node_ptr` or `m_data` had an offset. In other words, the previous version does not respect offsets, and copies data from the beginning of `m_node_ptr` to the beginning of the `m_data`.

The change basically fixes that to consider the offset in both cases: the first param becomes `static_cast<char*>(m_node_ptr->data_ptr()) + m_node_ptr->dtype().offset()` and `element_ptr(0)` returns a pointer to `m_data` that factors in the offset.

src/libs/conduit/conduit_data_accessor.cpp
siramok: Same answer as below, this is an unverified change. `element_ptr(0)` returns a pointer to `m_data` that factors in the offset if it has one. Although in this case, I don't know why `m_other_ptr` wouldn't also need to have its offset considered too (maybe context matters, or maybe it's currently just wrong).

src/libs/conduit/conduit_data_accessor.cpp
siramok: I agree, no sense in having an extra kernel for this when the operation is just a parallel sum and a single division.

src/libs/conduit/conduit_data_array.cpp
JustinPrivitera: Ahh that's right, thanks for reminding me.

src/libs/conduit/conduit_data_accessor.cpp
JustinPrivitera: The first line can be cleaned up from
```c++
static_cast<char*>(m_node_ptr->data_ptr()) + m_node_ptr->dtype().offset()
```
to 
```c++
m_node_ptr->element_ptr(0)
```

JustinPrivitera: I think your changes here are correct. It's just good to think deeply about them :)

src/libs/conduit/conduit_data_accessor.hpp
siramok: This shouldn't be necessary in the final version, but the reason this wasn't needed in the other direction is because setting on a `DataArray` is less complicated than setting on a `DataAccessor` due to it not performing any type conversion. `DataAccessor` has to evaluate the switch statement of possibilities (with a private helper, hence the friendship) while `DataArray` does not.

JustinPrivitera: Could you add a comment to this effect in the code? Although if it won't be necessary in the final version then I guess it doesn't matter.

src/libs/conduit/conduit_data_kernels.hpp
siramok: Yes, `Acc` was a relic from when I was generically referring to `DataAccessor` and `DataArray` as just an `Accessor`. This PR will need to pull in the latest changes we made to verbiage there and do similar updates here.

src/libs/conduit/conduit_data_kernels.hpp
siramok: I've looked into it, I think we can get away with duplication without blowing up compile times. I do think it would make the implementation simpler.

src/libs/conduit/conduit_data_accessor.cpp
siramok: To be clear, this definitely does work, so we can make this behave like magic if desired. My concern is a scenario where maybe we (or the user) forgets to call `use_with` themselves but finds that `.set()` still works anyways. Maybe copying across memory spaces was the desired behavior, but maybe it was not, and there wouldn't be a way to tell that something different happened under the hood. If `.set()` refuses to work, then it's an indicator that we should be more intentional about where memory lives.

src/libs/conduit/conduit_data_accessor.cpp
siramok: I believe that this was the single worst kernel that was making compile times unacceptably long, so making it host-only for now is a compromise (but I agree that the error message is bad)

src/libs/conduit/conduit_data_kernels.hpp
siramok: I've switched to duplication.

src/libs/conduit/conduit_data_kernels.hpp
siramok: This has been done.

JustinPrivitera: Thanks!

src/libs/conduit/conduit_data_kernels.hpp
JustinPrivitera: Gotcha. Does that mean that conduit_data_kernels.hpp will no longer exist?

siramok:
This PR is now in a state that I am happy with (minus the fact that we haven't decided about whether to support magic host<->device sets).

Compile times are noticeably worse, but not unbearable (+2 min or so, maybe this is cope and I have gotten used to how much worse it is). If we find it annoying, we can go down the route of adding a flag to `build_conduit.sh` that prevents dispatch from compiling so many extra kernel instantiations (on by default, but we can turn it off for ourselves to make development nicer).

The final conduit.so file also went from ~5 MB to ~30 MB. It's not 1999 so maybe we don't care, but just something to be aware of. Goes to show that we're compiling _a lot_ of extra kernels to let things be fast.

Data kernels .hpp went away and all of the kernels are now duplicated across `DataArray/Accessor`.

Given that all of these internal APIs are now querying for policy/active space, I went ahead and added the active space caching that we discussed previously. It's a minor optimization (each time we call `is_device_ptr()` ultimately has to query CUDA/HIP, which we don't want to have to do more than once) but the difference was measurable.

I did new performance tests which showed that this PR matches or slightly improves upon the base dispatch PR. So it feels like we didn't gain too terribly much by going this route, but 1) it's not worse and 2) maybe we just haven't found a good benchmark that highlights the changes yet.

src/libs/conduit/conduit_data_accessor.hpp
siramok: This didn't make it into the final version.

src/libs/conduit/conduit_data_accessor.cpp
JustinPrivitera: @cyrush is on team magic 🪄

siramok:
We want to look into caching the execution policy itself and not just the execution space (we do still want to save the cost of `to_device_ptr`). We can figure it out perhaps within the constructor, or add an unknown policy. We decided to explore this more.

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -242,15 +575,13 @@ template <typename T>
...
-    for(index_t i = 0; i < number_of_elements(); i++)
+    execution::dispatch(*this, [&](auto vals)
     {
-        const T &val = element(i);
-        if(val < res)
-        {
-            res = val;
-        }
-    }
+        res = detail::accessor_min_kernel<T>(m_policy, num_elements, vals);
```
JustinPrivitera: we should call active_policy() instead for whenever we use m_policy

siramok: Fixed.

src/libs/conduit/conduit_data_accessor.cpp
siramok: Support for magic sets has been added.

src/tests/conduit/t_conduit_data_accessor.cpp
siramok: I have added these comments.

siramok:
This PR has become quite large and unwieldy, but I think it now addresses everything we hoped to address in porting these data structures to the execution model.

JustinPrivitera:
> I did new performance tests which showed that this PR matches or slightly improves upon the base dispatch PR. So it feels like we didn't gain too terribly much by going this route, but 1) it's not worse and 2) maybe we just haven't found a good benchmark that highlights the changes yet.

What do you mean by this? I think the point of this effort was to add new functionality (being able to use more of the API with device memory), not necessarily to make things faster.

src/docs/sphinx/data_parallel_execution_model.rst
```diff
@@ -269,15 +269,15 @@ Moving Data Between Memory Spaces
...
 * ``sync()``: copy the contents of the working buffer back into the buffer the Node owns, reallocating the Node's data if the types or element counts no longer match. The Node keeps its original memory space, and the wrapper keeps its working buffer. A no-op if the data never moved.
 * ``assume()``: hand the working buffer to the Node. The Node releases its original buffer and takes ownership of the working buffer where it currently lives, which means the Node's data may change memory spaces. This avoids the copy that ``sync()`` performs. A no-op if the data never moved.
-* ``active_space()``: returns an ``ExecutionPolicy`` describing the memory space where the wrapper's data currently lives. This lets you execute wherever the input already resides:
+* ``active_policy()``: returns an ``ExecutionPolicy`` corresponding to the memory space where the wrapper's data currently lives. This lets you execute wherever the input already resides:
```
JustinPrivitera: It also is the cached execution policy. So if someone says use_with(openMP) then it will be openMP not serial. I think it has evolved now to encode more than just the memory space. Maybe we could make this description more informative.

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -37,90 +38,492 @@ namespace conduit
...
+                                               num_elements,
+                                               type_size,
+                                               type_size,
+                                               values,
+                                               type_size);
+        const T *temp_vals = static_cast<const T*>(temp_ptr);
+        execution::dispatch(accessor, [&](auto vals)
+        {
+            accessor_copy_from_view_kernel(policy, num_elements, temp_vals, vals);
+        });
```
JustinPrivitera: Do we need a temporary here? Could the accessor free its pointer and point to this new pointer?

src/libs/conduit/conduit_data_accessor.hpp
```diff
@@ -460,7 +467,9 @@ class CONDUIT_API DataAccessor
 
     index_t         m_offset;
     index_t         m_stride;
-    
+
+    /// Caches the execution policy, starts with an empty policy.
+    mutable conduit::execution::ExecutionPolicy m_policy;
```
JustinPrivitera, line 472: why does this need to be `mutable`? We could use a getter and setter instead, no?

src/libs/conduit/conduit_data_array.hpp
```diff
@@ -544,7 +546,9 @@ class CONDUIT_API DataArray
 
     index_t         m_offset;
     index_t         m_stride;
-    
+
+    /// Caches the execution policy, starts with an empty policy.
+    mutable conduit::execution::ExecutionPolicy m_policy;
```
JustinPrivitera, line 551: same q here

JustinPrivitera (changes requested):
This looks great, especially the thorough tests. I just had a few final questions.

siramok:
> > I did new performance tests which showed that this PR matches or slightly improves upon the base dispatch PR. So it feels like we didn't gain too terribly much by going this route, but 1) it's not worse and 2) maybe we just haven't found a good benchmark that highlights the changes yet.
> 
> What do you mean by this?

Similar to why I mention that this PR makes compile times significantly worse, I think all of the trade-offs and benefits (or lack thereof) of a big change should be made clear up front. For as many places that `DataAccessor/Array` get used, I was expecting more of an improvement across the board. In retrospect, I'm pleased that this PR doesn't make performance worse. That was not true initially; it took a decent amount of effort for that not to be the case in this current version.

>  I think the point of this effort was to add new functionality ... not necessarily to make things faster.

I agree that performance is not the only thing that matters. But porting to the execution model has (so far) always demonstrated that existing APIs become faster. The benchmark exists to help us quantify/validate the performance difference after porting new things to the execution model. Given that `DataAccessor/Array` are used everywhere that the execution model goes, why would we treat their port differently and not quantify the difference?

src/docs/sphinx/data_parallel_execution_model.rst
siramok: I agree, will work on that.

src/libs/conduit/conduit_data_accessor.cpp
siramok: Interesting question, I will investigate.

src/libs/conduit/conduit_data_accessor.hpp
siramok: I believe it had to be mutable to get around the fact that we have `const` functions modifying it. I'll investigate using a getter/setter.

src/libs/conduit/conduit_data_array.hpp
siramok: I believe it had to be mutable to get around the fact that we have `const` functions modifying it. I'll investigate using a getter/setter.

src/libs/conduit/conduit_data_accessor.hpp
JustinPrivitera: maybe we need it when DataAccessors are const?

JustinPrivitera:
> Given that `DataAccessor/Array` are used everywhere that the execution model goes, why would we treat their port differently and not quantify the difference?

You're right to look at old serial versus new serial, but I feel like our choices were more constrained in this case.

> So it feels like we didn't gain too terribly much by going this route, but 1) it's not worse and 2) maybe we just haven't found a good benchmark that highlights the changes yet.

I think we did gain a lot whether or not the performance is better, since we wanted data accessor/array to work for all of these device cases. Adding that missing functionality is more important than speeding up serial execution before and after.

siramok:
Good points. :)

src/libs/conduit/conduit_data_accessor.hpp
siramok: It's needed in this case because `active_policy`'s signature is const: `conduit::execution::ExecutionPolicy active_policy() const;`. Since `m_policy` is lazy evaluated, `active_policy` does have to modify `m_policy` the first time it gets called, and the mutable is what lets us get away with that.

src/libs/conduit/conduit_data_array.hpp
siramok: It's needed in this case because `active_policy`'s signature is const: `conduit::execution::ExecutionPolicy active_policy() const;`. Since `m_policy` is lazy evaluated, `active_policy` does have to modify `m_policy` the first time it gets called, and the mutable is what lets us get away with that.

src/libs/conduit/conduit_data_accessor.cpp
siramok: I have investigated. We do need a temp buffer. A bulk copy can move raw bytes into a different memory space, but there is no way to bulk copy data that is arbitrarily strided, offset, or needs a type conversion. So we first copy into the correct memory space, which is what the temp exists for, and then run a second pass that moves the temp data into the dst buffer with the correct striding, offset, and dtype.

However, if we aren't strided and if we don't have to do a conversion, then you can avoid temp and do a direct memcpy. I have added a fast path for that.

src/docs/sphinx/data_parallel_execution_model.rst
siramok: This has been done.

siramok:
All feedback has been addressed, this PR should be good to go.

src/libs/conduit/conduit_data_accessor.cpp
JustinPrivitera: Thanks for looking closely at this. Would it be possible to add a comment to this effect, here in the code? Then no future developer will scratch their head like me. 😅

src/libs/conduit/conduit_data_accessor.hpp
JustinPrivitera: Maybe the further question then is why does `conduit::execution::ExecutionPolicy active_policy()` need to be `const`? Is it because we will need it for a `const` declared `DataAccessor`?

src/libs/conduit/conduit_data_accessor.cpp
```diff
@@ -364,8 +364,30 @@ set_values_helper(const DataAccessor<T> &accessor,
...
+        // When the source type matches the destination's underlying dtype and
+        // the destination has a compact layout, no conversion is necessary,
+        // so we can just memcpy the values directly to the destination.
+        const bool same_type =
+            (std::is_floating_point<T>::value && dtype.is_floating_point()) ||
+            (std::is_integral<T>::value && std::is_signed<T>::value &&
+             dtype.is_signed_integer()) ||
+            (std::is_integral<T>::value && std::is_unsigned<T>::value &&
+             dtype.is_unsigned_integer());
+        const bool same_layout = same_type &&
```
JustinPrivitera, line 379: 🫨

JustinPrivitera (changes requested):
Looks good I think there was just one follow-up question to something from before.

src/libs/conduit/conduit_data_accessor.hpp
siramok: I have investigated, it is for multiple reasons.

You are correct, if we have a const reference to a `DataAccessor/Array` then `active_policy` needs to be const if we want the ability to ask for its policy:
```c++
const DataArray<T> &values = foo();
values.active_policy();
```

Furthermore, `active_policy` gets called from const functions in both classes: min, max, sum, mean, count, and set. Removing the constness from `active_policy` would require removing the constness of the functions that call it too.

src/libs/conduit/conduit_data_accessor.cpp
siramok: I have added comments that describe why temp buffers are needed in the relevant locations.

src/libs/conduit/conduit_data_accessor.hpp
JustinPrivitera: Gotcha, thanks for explaining it to me!

src/libs/conduit/conduit_data_accessor.cpp
JustinPrivitera: Thanks!

JustinPrivitera (approved):

PR #1672, opened 2026-08-10
adds optional support for reading and writting hdf5 attributes to conduit relay

cyrush:


src/libs/relay/conduit_relay_io.cpp
```diff
@@ -71,6 +71,47 @@ check_basic_protocol(const std::string &protocol)
...
+        }
+
+       ~HDF5OptionsContext()
+        {
+            if(!prev_options.dtype().is_empty())
+            {
+                hdf5_set_options(prev_options);
+                prev_options.reset();
+            }
+        }
```
siramok: Just nitpicking, but for code readability, I enjoy when the destructor immediately follows the constructor.

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -218,6 +223,34 @@ class HDF5Options
             }
         }
 
+        if(opts.has_child("attributes"))
+        {
+            const Node &atts = opts["attributes"];
+
+            if(atts.has_child("enabled"))
+            {
+                std::string enabled = atts["enabled"].as_string();
```
siramok, line 232: The user would have to be going out of their way to do the wrong thing here, but technically all we know at this point is that a child called `enabled` exists, not that it contains a string dtype.

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -218,6 +223,34 @@ class HDF5Options
...
+                if(enabled == "false")
+                {
+                    attributes_enabled = false;
+                }
+                else
+                {
+                    attributes_enabled = true;
+                }
```
siramok, line 240: Do you want attributes to be enabled by default or disabled by default?

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -218,6 +223,34 @@ class HDF5Options
...
+            }
+
+            if(atts.has_child("attributes_key"))
+            {
+                attributes_key = atts["attributes_key"].as_string();
```
siramok, line 245: Same feedback (might not matter, feel free to ignore).

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -218,6 +223,34 @@ class HDF5Options
...
+            }
+
+            if(atts.has_child("value_key"))
+            {
+                attributes_value_key = atts["value_key"].as_string();
```
siramok, line 250: Same feedback (might not matter, feel free to ignore).

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -803,9 +895,33 @@ bool check_if_conduit_list_is_compatible_with_hdf5_tree(const Node &node,
                                                         const Node &opts,
                                                 std::string &incompat_details);
 
+//-----------------------------------------------------------------------------
+bool check_if_attributes_are_compatible_with_hdf5_obj(const Node &node,
+                                                   const std::string &ref_path,
+                                                        hid_t hdf5_id,
+                                                        const Node &opts,
+                                                std::string &incompat_details);
```
siramok:
```suggestion
bool check_if_attributes_are_compatible_with_hdf5_obj(const Node &node,
                                                      const std::string &ref_path,
                                                      hid_t hdf5_id,
                                                      const Node &opts,
                                                      std::string &incompat_details);
```
I suspect this is what you were going for.

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -1609,6 +1792,17 @@ check_if_conduit_list_is_compatible_with_hdf5_tree(const Node &node,
 }
 
 
+//-----------------------------------------------------------------------------
+bool check_if_attributes_are_compatible_with_hdf5_obj(const Node &node,
+                                                   const std::string &ref_path,
+                                                        hid_t hdf5_id,
+                                                        const Node &opts,
+                                                std::string &incompat_details)
```
siramok:
```suggestion
bool check_if_attributes_are_compatible_with_hdf5_obj(const Node &node,
                                                      const std::string &ref_path,
                                                      hid_t hdf5_id,
                                                      const Node &opts,
                                                      std::string &incompat_details)
```

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -1609,6 +1792,17 @@ check_if_conduit_list_is_compatible_with_hdf5_tree(const Node &node,
...
+{
+   bool res = true;
+   // TODO
```
siramok: 👀

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -1621,15 +1815,20 @@ check_if_conduit_node_is_compatible_with_hdf5_tree(const Node &node,
     bool res = true;
 
     DataType dt = node.dtype();
+
     // check for leaf or group
-    if(dt.is_number() || dt.is_string() || dt.is_empty())
+    if(dt.is_number() || dt.is_string() || dt.is_empty() )
```
siramok:
```suggestion
    if(dt.is_number() || dt.is_string() || dt.is_empty())
```

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -1684,6 +1883,46 @@ check_if_hdf5_group_has_conduit_list_attribute(hid_t hdf5_group_id,
...
+
+#if H5_VERSION_GE(1, 12, 0) && !defined(H5_USE_18_API)
+    herr_t h5_status = H5Oget_info(hdf5_id, &h5_obj_info, H5O_INFO_ALL);
+#else
+    herr_t h5_status = H5Oget_info(hdf5_id, &h5_obj_info);
+#endif
+
+    hsize_t num_attrs = 0;
+
+    if( CONDUIT_HDF5_STATUS_OK(h5_status))
```
siramok:
```suggestion
    if( CONDUIT_HDF5_STATUS_OK(h5_status) )
```
Consistency with line 1668

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -2418,7 +2708,7 @@ write_conduit_node_children_to_hdf5_group(const Node &node,
                                               ref_path,
                                               child_name.c_str());
         }
-        else if( dt.is_object() || dt.is_list() )
+        else if( check_if_conduit_node_is_hdf5_group(child) )
```
siramok:
```suggestion
        else if(check_if_conduit_node_is_hdf5_group(child))
```
Consistency with the previous else if in the chain

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -2480,6 +2770,41 @@ write_conduit_node_children_to_hdf5_group(const Node &node,
...
+bool
+check_if_conduit_node_is_hdf5_dataset(const Node &node)
+{
+    DataType dt = node.dtype();
+    return ( dt.is_number() || dt.is_string() ) // basic case
+           ||
+           ( HDF5Options::attributes_enabled &&
+             dt.is_object() &&
+             check_if_conduit_object_is_hdf5_dataset_with_attributes(node)
+           );
```
siramok: I've not seen something like this before (I had no idea you could have newlines after a return). This isn't much more readable, but I present it for your consideration anyways (feel free to ignore):
```c++
const bool is_node_hdf5 = HDF5Options::attributes_enabled &&
                          dt.is_object() &&
                          check_if_conduit_object_is_hdf5_dataset_with_attributes(node);
return dt.is_number() || dt.is_string() || is_node_hdf5;
```

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -2480,6 +2770,41 @@ write_conduit_node_children_to_hdf5_group(const Node &node,
...
+}
+
+
+//---------------------------------------------------------------------------//
+bool
+check_if_conduit_node_is_hdf5_group(const Node &node)
+{
+    DataType dt = node.dtype();
+    return ( dt.is_object() &&
```
siramok: Similar feedback as above, but keep it if you like it.

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -2593,6 +2916,78 @@ remove_conduit_hdf5_list_attribute(hid_t hdf5_group_id,
...
+                                        ref_path);
+            h5_dtype_hnd.check_created();
+
+            hsize_t num_eles = (hsize_t) dt.number_of_elements();
+
+            RelayH5SHandle h5_dspace_hnd( H5Screate_simple(1,
+                                                           &num_eles,
+                                                           NULL),
+                                          hdf5_id,
+                                          ref_path);
```
siramok:
```suggestion
            RelayH5SHandle h5_dspace_hnd(H5Screate_simple(1,
                                                          &num_eles,
                                                          NULL),
                                         hdf5_id,
                                         ref_path);
```
Consistency

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -3024,6 +3419,20 @@ read_hdf5_group_into_conduit_node(hid_t hdf5_group_id,
...
     }
 
+    if(HDF5Options::attributes_enabled)
+    {
+        if(check_if_hdf5_object_has_attributes(hdf5_group_id,
+                                               ref_path))
+        {
+            const std::string &atts_key = HDF5Options::attributes_key;
+
+            std::string chld_ref_path = join_ref_paths(ref_path, atts_key);
```
siramok, line 3665: Perhaps can be const also?

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -3517,6 +3968,167 @@ read_hdf5_tree_into_conduit_node(hid_t hdf5_id,
...
+        // if they don't match, modify the dt
+        // and get the proper hdf5 data type handle
+        dt.set_endianness(Endianness::machine_default());
+
+        // clean up our old handle
+        h5_dtype_hnd.close();
+
+        // get ref to standard variant of this dtype
+        h5_dtype_hnd.set_id(conduit_dtype_to_hdf5_dtype(dt,
+                                                   ref_path));
```
siramok:
```suggestion
        h5_dtype_hnd.set_id(conduit_dtype_to_hdf5_dtype(dt,
                                                        ref_path));
```

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -3517,6 +3968,167 @@ read_hdf5_tree_into_conduit_node(hid_t hdf5_id,
...
+        // copy out to our dest
+        dest.set(n_tmp);
+    }
+
+    CONDUIT_CHECK_HDF5_ERROR_WITH_FILE_AND_REF_PATH(h5_status,
+                                                        hdf5_attr_id,
+                                                        ref_path,
+                                                        "Error reading HDF5 Attribue: "
+                                                         << hdf5_attr_id);
+
```
siramok:
```suggestion
    CONDUIT_CHECK_HDF5_ERROR_WITH_FILE_AND_REF_PATH(h5_status,
                                                    hdf5_attr_id,
                                                    ref_path,
                                                    "Error reading HDF5 Attribue: "
                                                     << hdf5_attr_id);
```
Alignment

src/tests/relay/t_relay_io_hdf5.cpp
```diff
@@ -2411,6 +2411,39 @@ TEST(conduit_relay_io_hdf5, conduit_hdf5_write_read_zfp_1d)
 
 }
 
+//-----------------------------------------------------------------------------
+TEST(conduit_relay_io_hdf5, empty_root_group_tests)
+{
+    // get objects in flight already
+    int DO_NO_HARM = check_h5_open_ids();
```
siramok, line 2418: A nice ideal to aspire to.

src/tests/relay/t_relay_io_hdf5_atts.cpp
```diff
@@ -0,0 +1,279 @@
...
+    nload.reset();
+    io::hdf5_read("tout_hdf5_attrs_custom_keys.hdf5",nload);
+    EXPECT_FALSE(nsansattr.diff(nload,info));
+
+}
+
+//-----------------------------------------------------------------------------
+TEST(conduit_relay_io_hdf5, hdf5_read_existing_atts)
+{
+    io::hdf5_reset_options();
```
siramok, line 105: Is it worth having a test which verifies that `io:hdf5_reset_options()` actually restores the options to the expected defaults?

siramok (commented):
LGTM, mainly just commented on nitpicks and style things (since I am not qualified to comment on the correctness of what's going on here). I think there were only 1-2 comments of substance, so feel free to reject/resolve the rest if you aren't a fan.

src/libs/relay/conduit_relay_io_hdf5.cpp
cyrush: thanks for the suggestion, yours is easier to read. 

I will change, but use

```
is_node_leaf_with_attributes
```

as the intermediate bool.

src/tests/relay/t_relay_io_hdf5_atts.cpp
cyrush: - [x] add a separate test for this

cyrush (commented):
Thanks for the suggestions, I accepted / committed most of them and I will work on the others.

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -426,6 +471,43 @@ class HDF5Options
...
+        HDF5Options::chunking_enabled   = true;
+        HDF5Options::chunk_size         = 1000000; // 1 mb
+        HDF5Options::chunk_threshold    = 2000000; // 2 mb
+
+        HDF5Options::attributes_enabled   = false;
+        HDF5Options::attributes_key       = "attributes";
+        HDF5Options::attributes_value_key = "value";
+
+        // TODO: ndarray chunk heuristics
```
JustinPrivitera, line 492: do you have a ticket for this? Maybe we don't need one?

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -1630,6 +1825,10 @@ check_if_conduit_node_is_compatible_with_hdf5_tree(const Node &node,
                                                                 opts,
                                                                 incompat_details);
     }
+    else if(check_if_conduit_node_is_hdf5_dataset(node)) // attributes
+    {
+        // TODO ATTS SUPPORT
```
JustinPrivitera: 👀

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -2480,6 +2770,41 @@ write_conduit_node_children_to_hdf5_group(const Node &node,
     }
 }
 
+//---------------------------------------------------------------------------//
+bool
+check_if_conduit_object_is_hdf5_dataset_with_attributes(const Node &node)
+{
+    return ( node.number_of_children() == 2 &&
+             node.has_child(HDF5Options::attributes_value_key) && // TODO also check for leaf
+             node.has_child(HDF5Options::attributes_key) // TODO also check for obj?
```
JustinPrivitera: TODOs?

JustinPrivitera (approved):

src/libs/relay/conduit_relay_io_hdf5.cpp
cyrush: Added: https://github.com/llnl/conduit/issues/1680

src/libs/relay/conduit_relay_io_hdf5.cpp
cyrush: They are disabled by default in the constructor and in the reset method of the HDF5Options class

src/libs/relay/conduit_relay_io_hdf5.cpp
cyrush: I agree this logic should be improved: Added a ticket to track this in the future:
https://github.com/llnl/conduit/issues/1681

src/libs/relay/conduit_relay_io_hdf5.cpp
cyrush: Added this logic

src/libs/relay/conduit_relay_io_hdf5.cpp
cyrush: Added this logic

src/libs/relay/conduit_relay_io_hdf5.cpp
cyrush: refactored this and added the extra checks

siramok (approved):

src/libs/relay/conduit_relay_io_hdf5.cpp
```diff
@@ -1799,9 +1814,145 @@ bool check_if_attributes_are_compatible_with_hdf5_obj(const Node &node,
...
+//-----------------------------------------------------------------------------
+bool
+check_if_attribute_leaf_is_compatible_with_hdf5_attribute(const DataType &dtype,
+                                                    const std::string &ref_path,
+                                                    hid_t hdf5_id,
+                                                    const Node & /*opts*/,
+                                                    std::string &incompat_details)
+{
+
+    bool res = true;
```
JustinPrivitera: If we're being really picky, I think it's safer to have return statements in the bottom levels instead of a bool that gets written to and then returned, since if you miss a case (you didn't) you can have unwanted behavior, while if you have a return each time and no return at the end then the compiler will warn you if you missed a case.

Please ignore this comment :)

JustinPrivitera (approved):

src/libs/relay/conduit_relay_io_hdf5.cpp
cyrush: Agreed, part of this style was due to old logic related to hdf5 handle cleanup.
But now that we have the handle classes that auto cleanup, we could switch to direct return statements.

Even then we could still miss a case (default return is going to be true, if we missed logic to flip to false, we true it is)

cyrush: refactored

JustinPrivitera: > Even then we could still miss a case (default return is going to be true, if we missed logic to flip to false, we true it is)

Doesn't the compiler give you at least a warning?

siramok: I believe it says something to the effect of "not all control paths return a value" if you're missing a return somewhere.

siramok: (I like direct returning unless it's a case that can't be directly returned; unique_mask in generate_centroids is that way. Another place it makes sense is if you're returning success/fail of a bunch of chained function calls.)

PR #1673, opened 2026-08-11
Port generate_centroids to the execution model

siramok:
Ports `mesh::topology::unstructured::generate_centroids` to the execution model. As with `to_unstructured`, I looked for opportunities to simplify and optimize where possible. I think it's in a good place, but I want to take another pass before I call it ready to be reviewed.

The diff is unfortunately atrocious, I think it might be easier to compare old vs new side-by-side.

Something I wanted to ask about: both `unstructured_centroid` and `unstructured_centroid_polyhedral` are optimized around an arbitrary upper bound for tracking the number of unique vertices per element. As long as each element consists of fewer unique vertices than the arbitrary upper bound, then computing centroids from those elements is relatively fast.

Previously this limit was 12 unique vertices per element, but in my testing I found that some of our example meshes went up to 16 unique vertices per element (starts to explain why pyramid-based meshes were particularly slow to compute). I upped the limit to 32 and kept the slow fallback paths, but this is all arbitrary. Maybe we would prefer to set an even higher limit that surely would never be exceeded in the real world (say, 64 or 128) and remove the slow paths?


src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1612,7 +1612,7 @@ unique_mask(const Container &values, index_t offset, index_t n, int *mask)
         {
             for(int row = 0; row < n; row++)
             {
-                for(int col = row + 1; col <= n; col++)
+                for(int col = row + 1; col < n; col++)
```
JustinPrivitera: this makes me scared 😨

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1626,7 +1626,7 @@ unique_mask(const Container &values, index_t offset, index_t n, int *mask)
...
```
JustinPrivitera: same comment here 😨

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: Then I will dress up as an off by 1 error for halloween 👻

`n` is the length of `values` (technically, `values` is much bigger than `n`, but `n` is the number of points associated with the current element in the larger connectivity array). `row` iterates `n` times, and `col` is always `row + 1`. Considering `if(values[row] == values[col])` , when `row` is in its final iteration it is equal to `n - 1` and so `col` is equal to `n`. If `n` is the length of `values`, `values[n]` is out of bounds (except in this case, `values` holds all of the element data, so you'd only read out of bounds for the very last element).

But this code was never getting compiled anyways because the LUT stuff is strangely compile-time guarded. This loop exists for creating masks of arbitrary sizes without a LUT, however the LUT only goes up to 8. We know that we can get up to 16 unique points per element in practice, so in that case, we simply don't create a LUT and force ourselves to go down the slow path. I've decided that I'm just going to make this work better, this clearly demonstrates that it should always be possible for us to pre-generate a mask so there's no reason not to.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: See above comment.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1469,30 +1469,68 @@ convert_topology_to_unstructured(const std::string &base_type,
...
+        of each unique value. Subsequent occurrences are marked 0.
 
  @param values The sequence of values to be searched.
  @param offset An offset from the start of values.
  @param n The number of values in the sequence.
- @param mask A buffer in which to store the mask. It must have at least n
-             elements.
- 
- @return True if there were dups; False otherwise.
+ @param mask A buffer in which to store the mask. It must have at least n elements.
```
JustinPrivitera: We technically have a line width style rule, although it isn't often followed.

Why was this change made though? It seems it was only done to make the line width longer than the rest?

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1469,30 +1469,68 @@ convert_topology_to_unstructured(const std::string &base_type,
...
-#define LUT
-#ifdef LUT
-    // Look up tables for the comparisons we make for n<=8.
-    const int ncaseslut[] = {0,0,1,3,6,10,15,21,28};
-    const int offsets[]   = {0,0,0,1,4,10,20,35,56};
-    const int leftlut[] = {
+    // Look up tables (LUTs) for elements of size n <= 16
+
+    // Number of comparisons for each element size
+    constexpr int count_lut[] = {
```
JustinPrivitera: where did the old numbers come from and where do the new numbers come from?

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1527,9 +1565,111 @@ unique_mask(const Container &values, index_t offset, index_t n, int *mask)
...
+        // 10 values
+        0,0,0,0,0,0,0,0,0,
+        1,1,1,1,1,1,1,1,
+        2,2,2,2,2,2,2,
+        3,3,3,3,3,3,
+        4,4,4,4,4,
+        5,5,5,5,
+        6,6,6,
+        7,7,
+        8,
```
JustinPrivitera: this all looks crazy - do we really need all of this written out like this? What is this stuff?

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1564,82 +1704,151 @@ unique_mask(const Container &values, index_t offset, index_t n, int *mask)
...
             {
-                for(int col = row + 1; col < n; col++)
+                if (values[row + offset] == values[col + offset])
                 {
-                    if(values[row] == values[col])
-                    {
-                        mask[col]--;
-                        needmask = true;
-                    }
+                    mask[col] = 0;
```
JustinPrivitera: I am confused by all of these changes... did the diff get confused?

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: I was sprucing up the entire function (it is only called by the non-polyhedral generate centroids kernel), tweaked the header as part of the sprucing, and while I was there seemingly decided that I didn't like that this was the only variable with a line break.

I have condensed the comment so that this variable description doesn't exceed the line width style rule.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: I couldn't find a named algorithm for deriving these numbers, but they are explicitly written out indices (pre-computed as opposed to using a double for loop) intended to accelerate finding the duplicate vertices in an element. The purpose of this function is to creating a mask of the unique vertices within an element, and I guess at one point in time these LUTs were faster than just plain looping (see next comment).

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: I wanted to pump up my GH stats :(

In seriousness, your comment inspired me to benchmark the performance difference with and without these LUTs:

- CPU performance was effectively the same either way.
- GPU performance improved by ~4% without LUTs.

I have therefore removed the LUTs entirely. They may have been beneficial when first introduced, but they are measurably detrimental as of today.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: ~This PR will probably be easier to review with the split diff rather than unified~ (this summer has taught me that GH diffs of non-trivial changes are very bad).

Nevermind, I checked for myself, the split diff is horribly designed and does not make reviewing this PR any easier.

TBH, these functions were effectively rewritten in place, and the number of lines changed is now almost the same as the number of lines removed. So the actual easiest way might just be to pull up the old file vs the new file to compare and contrast them side-by-side.

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: Wow that's interesting. Great, that makes all this simpler then. Thanks for investigating!

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: No worries, I trust you.

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: but now these are gone?

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: No worries. I was just wondering if this was a claude code thing or if it was intentional.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -322,7 +322,7 @@ TEST(blueprint_mesh_transform_benchmark, generate_transforms)
         configs.push_back({"generate_points_" + shape,    shape, "generate_points",    true});
         configs.push_back({"generate_lines_" + shape,     shape, "generate_lines",     true});
         configs.push_back({"generate_faces_" + shape,     shape, "generate_faces",     true});
-        configs.push_back({"generate_centroids_" + shape, shape, "generate_centroids", true});
+        configs.push_back({"generate_centroids_" + shape, shape, "generate_centroids", false});
```
JustinPrivitera, line 325: One thing that might be useful would be to examine the current tests for generate_centroids and add tests for polytess if there are not already tests for it... maybe we can look together and see what makes sense at our meeting.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -9285,6 +9274,7 @@ void mesh::convert(const conduit::Node &n_mesh,
             }
             else if(target == "generate_sides")
             {
+                // Justin called dibs on porting this
```
JustinPrivitera: I better do it now... or this comment will live forever
```suggestion
                // TODO Justin called dibs on porting this
```

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -5371,21 +5692,28 @@ mesh::topology::unstructured::generate_centroids(const Node &topo,
...
+    conduit::execution::dispatch(map_acc, [&](auto map_vals)
+    {
+        centroid_map_fill_kernel(policy, n, map_vals);
+    });
+    map_acc.data_movement(sync_strategy);
+
     d2smap.reset();
-    map_node.to_data_type(int_dtype.id(), s2dmap);
-    map_node.to_data_type(int_dtype.id(), d2smap);
+    d2smap.set_allocator(output_allocator_id);
```
JustinPrivitera: I wonder about the implications of having these special maps live on device... maybe there are none. We can discuss at 11.

JustinPrivitera (changes requested):
Just a few things for us to discuss; otherwise this looks good.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: Yes, they are now gone.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: Good question. I didn't look too closely before (assuming that things existed as they did for a reason), but now that you ask, I'm tracing what these maps are for and it seems like (generally speaking) they're only used for internal `mesh::convert` purposes. I.e., they are never handed back to the user.

However, I'm realizing that `generate_centroids` is a particularly bad instance because I don't think anything even consumes these maps... They appear to be created and then immediately torn down. 😖

My thoughts are:

1. If it's leftover code from a past refactor that is effectively dead, I see no reason to keep it around (it's equivalent to leaving a `sleep()` call in the code for no reason).
2. Even if I'm missing something and it's not actually dead, we know it's not meant to go back to the user, so at the very least I see no reason to bother syncing these back to the output memory space when we're done.
3. All the other generate APIs seem to call `topo_data.get_dim_map(*)` fill out these maps, so things are fishy with this particular instance. I suspect it just got missed in a past refactor.

JustinPrivitera: as we discussed we can remove these entirely

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: I extended the existing test so that different mesh and shape types (including polytess) could be exercised. Although I'm not really sure how to test their correctness, so instead I just have it checking for basic things, such as all of the centroids existing within the bounds of the coordset. Let me know what you think.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: I realized that removing these entirely would require changing the API and updating several places that call generate_centroids. I didn't do that for now, but if you want me to I can.

siramok: We decided on thorough removal, so I will work on that.

siramok: It turns out that these maps will be semi-complicated to remove from generate_centroids thoroughly. Removing them from the function parameters will require changes across several files, including a workaround in one case that unconditionally expects all of the generate API signatures to take these maps as input (even though it never gets used).

I've removed the trivial bits that I could get away with for now, and suggest that we make an issue to track thorough removal and then follow-up with a new PR to avoid bloating this one.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2029,24 +2034,22 @@ calculate_unstructured_centroids(const conduit::Node &topo,
...
-    const DataType subconn_dtype(topo_subconn.dtype().id(), 1);
-    const DataType suboffset_dtype(topo_suboffsets.dtype().id(), 1);
-    const DataType subsize_dtype(topo_subsizes.dtype().id(), 1);
+    // Following the convention used in other APIs
+    const Node &first_axis_values = coordset["values"][csys_axes[0]];
+
+    conduit::execution::ExecutionPolicy policy =
+        conduit::execution::get_execution_policy(first_axis_values);
+    const index_t output_allocator_id =
+        conduit::execution::get_output_allocator_id(first_axis_values);
```
siramok, line 2050: > > The execution policy is now queried from an existing node and is not queried from the global settings.
> 
> I'm not sure that's what we decided. We wanted the global settings to take an accessor instead of a node.
> 
> We could start that work on a new branch if you're not already too far.

To clarify, I was referring to our discussion about making sure that we're following this pattern (we don't need a fallback in this case because `generate_centroids` by design assumes that it was passed a valid coordset).

Making `get_execution_policy` take accessors will be its own big change that will need a separate PR.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: We decided that this is acceptable plan. I made an issue to track it here: #1689.

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
JustinPrivitera: Looks good. If we wanted to be more thorough we could verify that each centroid is contained in the bounding box of its related element, instead of the bounding box for the whole mesh.

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: What do you mean by valid coordset? There are valid coordsets that do not have any array data (uniform coordsets). Do we know for sure we will never see a uniform coordset? Will we always get an explicit one?

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: That is a reasonable change, I can work on that.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: Valid as in it should always be an explicit coordset.

Before any `generate_*` API gets called, we unconditionally convert the input mesh to an unstructured mesh if it's not already an unstructured mesh: https://github.com/llnl/conduit/blob/d0c957f961c119e22aabe750f7555a677efe88b9/src/libs/blueprint/conduit_blueprint_mesh.cpp#L9420-L9434

Based on how `to_unstructured` is implemented, my understanding is that unstructured meshes are always going to have an explicit coordset: https://github.com/llnl/conduit/blob/d0c957f961c119e22aabe750f7555a677efe88b9/src/libs/blueprint/conduit_blueprint_mesh.cpp#L1554-L1570

Someone could technically call `generate_centroids` directly (bypassing `mesh::convert`) and pass in a uniform coordset, but bad things would happen if they tried. The API would either break in the worst case or return a empty/useless result in the best case, because both of the centroid kernels unconditionally expect to take a non-empty coordset as input: https://github.com/llnl/conduit/blob/d0c957f961c119e22aabe750f7555a677efe88b9/src/libs/blueprint/conduit_blueprint_mesh.cpp#L1775-L1779

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
JustinPrivitera: might be a little tricky to find the element coordinates... will involve some array indirection. But that could be a good exercise!

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: Ok great, thanks for tracking that down!

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: This has been addressed.

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -2029,24 +2034,22 @@ calculate_unstructured_centroids(const conduit::Node &topo,
...
-    const DataType size_dtype(topo_sizes.dtype().id(), 1);
 
-    const DataType subconn_dtype(topo_subconn.dtype().id(), 1);
-    const DataType suboffset_dtype(topo_suboffsets.dtype().id(), 1);
-    const DataType subsize_dtype(topo_subsizes.dtype().id(), 1);
+    // Following the convention used in other APIs
+    const Node &first_axis_values = coordset["values"][csys_axes[0]];
+
+    conduit::execution::ExecutionPolicy policy =
+        conduit::execution::get_execution_policy(first_axis_values);
```
JustinPrivitera, line 2048: I don't want to muddy the water too much. The intention with this approach (passing a node to the global settings) is to pass a node that is representative of the data that is going to be read in order to compute the resulting data. "Run where your data already is". I see that the `first_axis_values` are used later in the function, but it isn't clear to me from looking at the GitHub diff that they are used when creating the connectivity elements, which is what the execution policy is first used for. I wonder if it makes sense to fetch a new execution policy by passing in representative source data for each forall that occurs in a given function? Since data could in theory live in different places. Coordset data could be in one place and topology data another. This case will be common if people run their meshes through these transforms and they start on host and want to run on device. Conduit will only move the pieces it touches to the device, so at any point there could be source data that is still on host and other source data already on device, depending how far through a set of transforms we are.

I guess all this is to say that we should be certain we are making the right choices each time we pass a node (and later, an accessor) to `conduit::execution::get_execution_policy()`. I'm not convinced from looking at the diff that this particular choice makes sense, and I wonder if we should request a new policy for every forall that we run, or if that is too much of a bother. This should be something we discuss. I will put a link to this comment somewhere that we will remember so we can look at this in the future. I think this is fine for now.

JustinPrivitera (approved):
This looks good. Thanks for addressing everything.

PR #1678, opened 2026-08-21
python: Export variable giving install python modules path

scottwittenburg:
Export a new variable, `CONDUIT_PYTHON_DIR`, giving the absolute installed location of Conduit's Python modules, regardless of whether a custom prefix was used, or the build was relocated.

Update install test to validate the module can be found in the advertised location, even if the install is relocated.


cyrush (approved):
Thanks for iterating and adding this!

PR #1679, opened 2026-08-22
Add python build directories to .gitignore

siramok:
`build_conduit.sh` enables python by default now, which generates its build directories under `src/libs/conduit/python/`.

I've been extra careful to avoid committing them, but it's only a matter of time...


cyrush (approved):
Thanks for adding this!

PR #1682, opened 2026-08-25
Optimize the uniform and explicit fill kernels for device execution

siramok:
Optimizes uniform and explicit fill kernels for device execution (I measured a ~1.4-30x speedup across 5 of the 6 mesh conversions on my local PC). This is achieved by 1) consolidating the 3 separate kernel launches for uniform fill into a single kernel launch (the act of launching a GPU kernel is somewhat expensive) and 2) expressing the explicit fill as a device-friendly problem.

Host execution is the exactly same for explicit fill and is slightly faster for uniform fill (since it benefits from the kernel consolidation). This is an interesting case where for the first time (that I have come across this summer), there is a good performance reason to maintain separate kernels for host/device execution for the same operation.

I did save the converted meshes before vs after this PR and verified that their diff passes/they are identical.

### DIFF
--- a/src/libs/blueprint/conduit_blueprint_mesh.cpp
+++ b/src/libs/blueprint/conduit_blueprint_mesh.cpp
@@ -1065,29 +1065,41 @@ coordset_explicit_fill_host_kernel(conduit::execution::ExecutionPolicy &policy,
                                    const index_t dim_block_size,
                                    const index_t dim_block_count)
 {
-    conduit::execution::forall(policy, 0, dim_len, [=] CONDUIT_EXEC(index_t d)
+    // We duplicate these foralls to avoid having to branch on
+    // rectilinear/uniform. The only difference is how we ultimately set
+    // the values.
+    if (is_base_rectilinear)
     {
-        const index_t doffset = d * dim_block_size;
-        for (index_t b = 0; b < dim_block_count; b++)
+        conduit::execution::forall(policy, 0, dim_len, [=] CONDUIT_EXEC(index_t d)
         {
-            const index_t boffset = b * dim_block_size * dim_len;
-            for (index_t bi = 0; bi < dim_block_size; bi++)
+            const index_t doffset = d * dim_block_size;
+            for (index_t b = 0; b < dim_block_count; b++)
             {
-                const index_t ioffset = bi + doffset + boffset;
-                // TODO: Investigate whether it's faster to have 2 separate
-                // kernels for rectilinear and uniform fills to avoid this
-                // comparison every iteration.
-                if (is_base_rectilinear)
+                const index_t boffset = b * dim_block_size * dim_len;
+                for (index_t bi = 0; bi < dim_block_size; bi++)
                 {
+                    const index_t ioffset = bi + doffset + boffset;
                     dst_cvals.set(ioffset, src_cvals[d]);
                 }
-                else if (is_base_uniform)
+            }
+        });
+    }
+    else if (is_base_uniform)
+    {
+        conduit::execution::forall(policy, 0, dim_len, [=] CONDUIT_EXEC(index_t d)
+        {
+            const index_t doffset = d * dim_block_size;
+            for (index_t b = 0; b < dim_block_count; b++)
+            {
+                const index_t boffset = b * dim_block_size * dim_len;
+                for (index_t bi = 0; bi < dim_block_size; bi++)
                 {
+                    const index_t ioffset = bi + doffset + boffset;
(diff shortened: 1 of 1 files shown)


src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -954,68 +985,103 @@ convert_coordset_to_rectilinear(const std::string &/*base_type*/,
...
+            const index_t boffset = b * dim_block_size * dim_len;
+            for (index_t bi = 0; bi < dim_block_size; bi++)
             {
-                index_t ioffset = doffset + boffset + bi;
-                if(is_base_rectilinear)
+                const index_t ioffset = bi + doffset + boffset;
+                // TODO: Investigate whether it's faster to have 2 separate
+                // kernels for rectilinear and uniform fills to avoid this
+                // comparison every iteration.
+                if (is_base_rectilinear)
```
JustinPrivitera: can we just bump the if statement outside of the forall and just have two foralls?

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1025,6 +1091,56 @@ coordset_explicit_fill_kernel(conduit::execution::ExecutionPolicy &policy,
...
+    const index_t coords_len = dim_len * dim_block_size * dim_block_count;
+    conduit::execution::forall(policy, 0, coords_len, [=] CONDUIT_EXEC(index_t i)
+    {
+        // Invert the layout formula to recover d. With bi < dim_block_size,
+        //   i = bi + d * dim_block_size + b * dim_block_size * dim_len
+        //   i / dim_block_size = d + b * dim_len
+        // and taking that modulo dim_len leaves d.
+        const index_t d = (i / dim_block_size) % dim_len;
+
+        // TODO: Investigate whether it's faster to have 2 separate
```
JustinPrivitera: same q here

src/libs/blueprint/conduit_blueprint_mesh.cpp
```diff
@@ -1098,21 +1214,42 @@ convert_coordset_to_explicit(const std::string &base_type,
...
-                                          src_vals,
-                                          dst_vals);
-        });
+        if (policy.is_device_policy())
+        {
+            conduit::execution::dispatch(src_cvals_acc,
+                                         dst_cvals_acc,
+                                         [&](auto src_vals, auto dst_vals)
+            {
+                coordset_explicit_fill_device_kernel(policy,
```
JustinPrivitera, line 1247: it would be good to have a comment in the code explaining why we need the two different kernels

JustinPrivitera (changes requested):
see my minor comments

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: Yes, probably.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: Yes, probably.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: I think the headers contain this, would you prefer that it be moved into the function body itself?

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: ok let's make the change

src/libs/blueprint/conduit_blueprint_mesh.cpp
JustinPrivitera: I think it is good to have explanations live in all of the places where people reading the code might have confusion. It doesn't hurt to have the same comment in several places.

siramok: I have added more comments.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: I have made this change per your suggestion and verified that everything still compiles and runs as expected.

src/libs/blueprint/conduit_blueprint_mesh.cpp
siramok: I have made this change per your suggestion and verified that everything still compiles and runs as expected.

JustinPrivitera (approved):

PR #1684, opened 2026-08-26
cmake: register python tests with Python3_EXECUTABLE

scottwittenburg:
`SetupPython.cmake` sets `Python3_EXECUTABLE` from `PYTHON_EXECUTABLE`, but not the other way around, so configuring with `-DPython3_EXECUTABLE` leaves `PYTHON_EXECUTABLE` empty. At the same time, `add_python_test()` and `add_python_mpi_test()` still used the old name, which registered every python test with an empty program name, i.e.:

```
add_test(NAME t_python_conduit_smoke COMMAND "" -B -m unittest ...)
```

ctest then treated "-B" as the executable, reporting "Could not find executable -B" for each one, and the tests could not run.

Make the tests use `Python3_EXECUTABLE`, so that regardless of which cache variable is chosen by users, the tests can be run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 
 ### Fixed
 
+#### General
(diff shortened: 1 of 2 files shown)


cyrush (approved):
This looks - good, thanks for this fix. It should fully bridge new and old way of providing python exe. 

Our CI is still using the old way, so it wouldn't have exercised the new way.

PR #1686, opened 2026-08-26
Guard against quiet install failures with old setuptools

scottwittenburg:
### Summary

This is a follow-up to #1675, hardening the Python-module install step so it can no longer fail silently. #1675 made the install air-gap-safe by adding `--no-index --no-deps --no-build-isolation` to the install-time `pip install`. A consequence of that is `--no-build-isolation` makes pip build with the setuptools already in the selected Python (instead of fetching the version specified in `pyproject.toml`), and this exposed a second, quieter failure mode.

The staging step can fail silently in two distinct ways:

1. pip fails (its exit status is ignored at install time)
2. pip succeeds but produces nothing (an old setuptools ignores the PEP 621 metadata and builds an empty package)

This PR adds a guard for each, plus an early configure-time check for the most common trigger of the second failure.

### Background

The modern day project description, `src/libs/conduit/python/pyproject.toml`, requests `setuptools>=68`. Under build isolation, pip would fetch that, but with `--no-build-isolation` (needed for offline/no-ssl builds) pip uses the existing setuptools and the `requires` minimum version is ignored. If that existing setuptools is too old to understand the metadata (PEP 621 support was added with setuptools `61.0.0`), it doesn't error. Instead it builds a package with no python sources (name reported as UNKNOWN, version 0.0.0) and exits 0. In that case, `make install` succeeds, but the installed Conduit module is empty of any `.py` files (only the compiled modules).

### Changes

This PR adds a configure-time check of the setuptools version (failing if less than 61.0.0), and two post-install checks that catch the two failure modes mentioned above. The configure-time check is enforcement of a hard bare-minimum, and is not related to the setuptools version pinned in `pyproject.toml`.

1. Add a configure-time check of the setuptools version (when python is enabled), and fail if we don't have `setuptools<=61.0.0`.
2. Add a post-install exit-status guard. The install-time `EXECUTE_PROCESS` now captures `RESULT_VARIABLE` and fails on a non-zero pip exit. The build-time pip is an `add_custom_command`, which already fails the build on non-zero exit, this change just make the install-time path behave the same.
3. Add a post-install check ensuring the install didn't fail to place the top-level `__init__.py`.


cyrush (approved):
Perfect -- this looks great, thanks again!
I plan to merge post CI.

PR #1688, opened 2026-08-27
Simplify Windows CI fix

siramok:
The latest fix to CI can be simplified. However, I couldn't make it more simple than this without creating new defines (which is an option).

### DIFF
--- a/src/libs/conduit/conduit_data_accessor.hpp
+++ b/src/libs/conduit/conduit_data_accessor.hpp
@@ -471,7 +471,7 @@ private:
 // -- conduit::DataAccessor explicit instantiation declarations --
 //
 //-----------------------------------------------------------------------------
-#if defined(_WIN32) && defined(CONDUIT_WINDOWS_DLL_EXPORTS) && \
+#if defined(CONDUIT_WINDOWS_DLL_EXPORTS) && \
     !defined(CONDUIT_EXPORTS_DEFINED) && \
     !defined(CONDUIT_TU_IS_CUDA) && !defined(CONDUIT_TU_IS_HIP)
 
--- a/src/libs/conduit/conduit_data_array.hpp
+++ b/src/libs/conduit/conduit_data_array.hpp
@@ -555,7 +555,7 @@ private:
 // -- conduit::DataArray explicit instantiation declarations --
 //
 //-----------------------------------------------------------------------------
-#if defined(_WIN32) && defined(CONDUIT_WINDOWS_DLL_EXPORTS) && \
+#if defined(CONDUIT_WINDOWS_DLL_EXPORTS) && \
     !defined(CONDUIT_EXPORTS_DEFINED) && \
     !defined(CONDUIT_TU_IS_CUDA) && !defined(CONDUIT_TU_IS_HIP)
 


src/libs/conduit/conduit_data_accessor.hpp
```diff
@@ -471,7 +471,7 @@ class CONDUIT_API DataAccessor
 // -- conduit::DataAccessor explicit instantiation declarations --
 //
 //-----------------------------------------------------------------------------
-#if defined(_WIN32) && defined(CONDUIT_WINDOWS_DLL_EXPORTS) && \
+#if defined(CONDUIT_WINDOWS_DLL_EXPORTS) && \
     !defined(CONDUIT_EXPORTS_DEFINED) && \
     !defined(CONDUIT_TU_IS_CUDA) && !defined(CONDUIT_TU_IS_HIP)
```
JustinPrivitera, line 476: is there not a macro that contains both of these?

src/libs/conduit/conduit_data_accessor.hpp
```diff
@@ -471,7 +471,7 @@ class CONDUIT_API DataAccessor
 // -- conduit::DataAccessor explicit instantiation declarations --
 //
 //-----------------------------------------------------------------------------
-#if defined(_WIN32) && defined(CONDUIT_WINDOWS_DLL_EXPORTS) && \
+#if defined(CONDUIT_WINDOWS_DLL_EXPORTS) && \
     !defined(CONDUIT_EXPORTS_DEFINED) && \
```
JustinPrivitera, line 475: and did we need CONDUIT_EXPORTS_DEFINED?

siramok: `defined(CONDUIT_WINDOWS_DLL_EXPORTS)` implies `defined(_WIN32)`, which is why we can remove that macro.

`CONDUIT_EXPORTS_DEFINED` is a macro we need, but it being defined does not imply that `CONDUIT_WINDOWS_DLL_EXPORTS` is also defined, hence why we do need both.

Similar to the above comment, we could define a new macro that implies the exact case we're interested in (which would let these 2 defines collapse into one), but I assumed that you would not want new macros.

This is based on the macro definition logic here: https://github.com/llnl/conduit/blob/2f1fe8ed52d4ecd5cebcbc28974e2bc6821d7d30/src/libs/conduit/conduit_exports.h.in

src/libs/conduit/conduit_data_accessor.hpp
siramok: Not 1-1.

`CONDUIT_TU_IS_CUDA` is defined if `defined(CONDUIT_USE_CUDA) && defined(__CUDACC__)`.  Similarly, `CONDUIT_TU_IS_HIP` is defined if `defined(CONDUIT_USE_HIP) && defined(__HIPCC__)`.

We have `CONDUIT_USE_DEVICE`, but it's defined when `defined(CONDUIT_USE_CUDA) || defined(CONDUIT_USE_HIP)) && defined(CONDUIT_USE_UMPIRE)`, which is not exactly the same.

There's also `CONDUIT_DEVICE_COMPILE`, which is defined when `defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__)`, which is again not exactly the same.

We could make a new macro that implies `defined(CONDUIT_TU_IS_CUDA) || defined(CONDUIT_TU_IS_HIP)` to help us simplify, but I figured you would prefer not to add new macros.

We could try substituting the other two macros to see if Windows CI continues to pass anyway (despite not being precisely equivalent to `!defined(CONDUIT_TU_IS_CUDA) && !defined(CONDUIT_TU_IS_HIP)`), but it would be trial and error at that point, given that we know what's in there now already works.

Based on the logic here: https://github.com/llnl/conduit/blob/5091fae6083fda5b37c3d7ec224aab80ff3aa89b/src/libs/conduit/conduit_execution_macros.hpp

JustinPrivitera (approved):
Thanks for being thorough!

PR #1692, opened 2026-09-04
Add a build option to control compilation of typed dispatch kernels

### LINKED ISSUE #1691: Explore making a build_conduit.sh option to toggle on/off additional kernel instantiations within execution::dispatch
We're concerned that the extra kernel instantiations produced by `execution::dispatch` are causing us CI pain. The extra compile time seems tolerable, but we theorize that the increased `conduit.so` sizes (~5 MB -> ~30 MB) are potentially causing the static CI builds to fail (maybe due to OOM?).

siramok:
This PR makes typed dispatches a build option, meaning that the extra typed kernel instantiations it generates will not be compiled if the option is turned off. The fallback kernels are always generated, so this option will never break functionality. For static builds, I saw local compile times improve by ~25% and the total build directory size reduce by ~33%.

I've defaulted it to off within cmake to fix our CI woes, but have left it defaulted to on within `build_conduit.sh` as I see that there is precedent for splitting the decision like this (python off in cmake but on in build_conduit, tests on in cmake but off in build_conduit, etc).

Closes #1691.

### DIFF
--- a/scripts/build_conduit/build_conduit.sh
+++ b/scripts/build_conduit/build_conduit.sh
@@ -51,6 +51,7 @@ build_zfp="${build_zfp:=false}"
 
 # conduit options
 build_conduit="${build_conduit:=true}"
+enable_typed_dispatch="${enable_typed_dispatch:=ON}"
 
 # see if we are building on windows
 build_windows="${build_windows:=OFF}"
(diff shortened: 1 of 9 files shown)


cyrush (approved):
this looks great to me!

PR #1693, opened 2026-09-04
Optimize the execution model for MI300A-style unified memory APUs

siramok:
This PR optimizes the execution model for MI300A-style unified memory APUs (tuolumne, el cap) based on insights extracted from [here](https://hpc.llnl.gov/documentation/user-guides/using-el-capitan-systems/using-el-capitan-systems-memory-management), [here](https://arxiv.org/html/2508.12743v1#abstract1), offline discussions, and results from the `mesh::convert` benchmark.

### What does it mean to have unified memory?
By default, the MI300A's CPU can always read memory allocated by the device runtime (e.g., `hipMalloc`) but its GPU cannot read host-allocated memory (e.g., `malloc`). Setting `HSA_XNACK=1` allows the GPU to read host-allocated memory in exchange for up to  a ~5% performance penalty in GPU execution.

We know that we're operating on a unified memory architecture when host- and device- allocated memory are accessible from both directions. There are still benefits to maintaining separate host/device pools (described later), as they will use different allocators under the hood, so that's still true in this PR (I did not go with my plan to use `hipMalloc` for everything).

From a code-perspective, it just means that we don't need to treat host and device pointers as different things (at least when it comes to execution).

### The performance implications
This will mainly affect any scenario where the execution and output locations would otherwise differ:
- We don't have to explicitly copy data to and from discrete memory spaces before we operate on it.
- Execution policies are not tied to a particular memory space, meaning:
    - `use_with()`, `sync()`, and `assume()` are always no-ops.
    - The default `output_location` of "input" can be selected based on the fastest allocator type.
    - We can select an initial active policy based on what will be fastest for the input size.

Unified memory lets us automatically make fast choices by default, without removing the ability for users to explicitly ask for the execution policy/output location that they want. 

### What about synchronization?
Unified memory complicates things slightly, because we cannot assume that device-allocated memory will be ready in time for the CPU to execute on it. One nice thing is that we don't have to worry about the reverse i.e., host-allocated memory being ready in time for the GPU to execute on it. I've added `hipSteamSynchronize` where we're at risk of this happening, and currently believe that all of our memcpy/memset instances are safe/getting synchronized.

Another thing that would normally be unsafe are the `RAJA::forall` calls. Except that we currently only use the synchronous RAJA policies, which makes our `RAJA::forall` calls safe. If we do plan to explore the async RAJA polices in the future, then yes, just be aware that extra synchronization will be necessary in that case.

### What about mixing MPI with device-allocated memory?
This was a concern raised during offline discussion. Basically, MPI doesn't work with device-allocated data, even if the CPU itself working with it. It sounds like the best solution for this is to just set `MPICH_GPU_SUPPORT_ENABLED=1`, which makes MPI to work directly with device-allocated memory.

If for some reason that's not desired or possible, the user can still manually specify "host" as their `output_location`. This will unconditionally use a host allocator for the output `Node`. Meaning that we can still pick a device execution policy, have it operate on host-allocated memory, and then subsequently have no issue moving that data around using MPI. This would have to look like: `host->device->host->MPI` or `device->device->host->MPI`.

### Plotted results
All results collected on tuolumne.

Comparison of how this PR runs with and without `HSA_XNACK=1`  set (black line is best case scenario):
unified_vs_discrete

Raw plot for `HSA_XNACK=0` (overhead of managing discrete memory spaces is apparant):
mesh_scaling

Raw plot for `HSA_XNACK=1` (the results significantly improve/tighten up):
mesh_scaling


src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -122,8 +122,10 @@ TEST(conduit_execution, execution_settings)
         EXPECT_EQ(device_alloc_id, DEVICE_ALLOC_ID);
         EXPECT_EQ(host_alloc_id, HOST_ALLOC_ID);
 
-        // prove that host data is on host
-        EXPECT_FALSE(execution::DeviceMemory::is_device_ptr(host_data.data_ptr()));
+        // prove that host data is on host (which is device accessible on
+        // unified memory systems)
+        EXPECT_EQ(execution::DeviceMemory::is_unified(), execution::DeviceMemory::is_device_ptr(host_data.data_ptr()));
```
JustinPrivitera: I think these should be separated out into separate test statements. Typically with EXPECT_EQ one argument is the expected result and the other is the actual. We are testing two different things so we should separate them out. If they are always in lockstep then we should have an if statement that breaks them out. Does that make sense?

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -133,7 +135,7 @@ TEST(conduit_execution, execution_settings)
 #else // !defined(CONDUIT_USE_DEVICE)
         // without device support, device_data stays on host
         device_data.set(src_vals);
-        EXPECT_FALSE(execution::DeviceMemory::is_device_ptr(device_data.data_ptr()));
+        EXPECT_EQ(execution::DeviceMemory::is_unified(), execution::DeviceMemory::is_device_ptr(device_data.data_ptr()));
```
JustinPrivitera: same thought here

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -324,7 +326,7 @@ TEST(conduit_execution, execution_settings)
         execution::ExecutionPolicy host_supplied_policy = execution::get_execution_policy(host_data);
         execution::ExecutionPolicy device_supplied_policy = execution::get_execution_policy(device_data);
         EXPECT_TRUE(fallback_policy.is_host_policy());
-        EXPECT_TRUE(host_supplied_policy.is_host_policy());
+        EXPECT_EQ(host_supplied_policy.is_device_policy(), execution::DeviceMemory::is_unified());
```
JustinPrivitera: and here

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -350,7 +352,7 @@ TEST(conduit_execution, execution_settings)
         execution::ExecutionPolicy host_supplied_policy = execution::get_execution_policy(host_data);
         execution::ExecutionPolicy device_supplied_policy = execution::get_execution_policy(device_data);
         EXPECT_TRUE(fallback_policy.is_serial());
-        EXPECT_TRUE(host_supplied_policy.is_host_policy());
+        EXPECT_EQ(host_supplied_policy.is_device_policy(), execution::DeviceMemory::is_unified());
```
JustinPrivitera: here

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -373,7 +375,7 @@ TEST(conduit_execution, execution_settings)
         execution::ExecutionPolicy host_supplied_policy = execution::get_execution_policy(host_data);
         execution::ExecutionPolicy device_supplied_policy = execution::get_execution_policy(device_data);
         EXPECT_TRUE(fallback_policy.is_device_policy());
-        EXPECT_TRUE(host_supplied_policy.is_host_policy());
+        EXPECT_EQ(host_supplied_policy.is_device_policy(), execution::DeviceMemory::is_unified());
```
JustinPrivitera: ^

src/tests/conduit/t_conduit_execution.cpp
```diff
@@ -667,7 +669,7 @@ TEST(conduit_execution, execution_settings)
 #else
         EXPECT_EQ(fallback_alloc_id, HOST_ALLOC_ID);
 #endif
-        EXPECT_EQ(host_supplied_alloc_id, HOST_ALLOC_ID);
+        EXPECT_EQ(host_supplied_alloc_id, execution::DeviceMemory::is_unified() ? DEVICE_ALLOC_ID : HOST_ALLOC_ID);
```
JustinPrivitera: there are so many more I stopped commenting

JustinPrivitera (changes requested):
I am just wondering if we should change how we are doing the tests.

src/tests/conduit/t_conduit_execution.cpp
siramok: Yes, that makes sense. So to be clear, do you favor adding:
```c++
if (execution::DeviceMemory::is_unified())
{
// new check
}
else
{
// old check
}
```
To all of these locations?

Or do you prefer duplicating these tests, and having the duplicates be specialized for unified memory?

src/tests/conduit/t_conduit_execution.cpp
siramok: No worries!

src/tests/conduit/t_conduit_execution.cpp
JustinPrivitera: The first option. And then we can use EXPECT_TRUE or EXPECT_FALSE instead of EXPECT_EQ

siramok: It shall be done.

siramok: This change has been made everywhere. I remembered why I didn't do it in the first place: it doubles the size of the PR. :)

JustinPrivitera: More code lines == more contributions == higher github stats!

siramok: == longer time to get reviewed due to fear of big PR xD

JustinPrivitera (approved):
Thanks for doing these final changes; sorry for the pedantic suggestion.

src/tests/conduit/t_conduit_execution.cpp
JustinPrivitera: We'll sit down in an upcoming hackathon and grind through all of it :)

PR #1695, opened 2026-09-09
Updates to the benchmark plotting script

siramok:
I've made various changes to the plotting script over the summer, and have so far not bothered to merge those changes back in. Now that I am leaving, it seems important to make these changes available to assist in future benchmarking efforts. This PR intentionally tracks the develop branch, because I don't think it makes sense to try to maintain two versions of this script that accounts for the differences. It should be fine if the device support branch syncs with develop and replaces its version of the script outright with this one.

These changes do several things:
- The develop branch had `mesh_` prefixed on all of its benchmark names, but the device support branch dropped that. This PR unifies the names between the branches to make it simpler to compare them.
- Adds the ability to do `--compare base.cali new.cali` for comparing changes across 2 benchmark runs. It works with both a mix of .cali files between the develop branch and the device support branch, or two .cali files from the same branch. I find it particularly useful for comparing stuff vs the develop and device support branches.
- I've also found it useful for comparing the new APU results of discrete memory vs unified memory, so there's some support for that as well.
- The individual plots that pre-existed had various issues that annoyed me (spacing, readability, not useful) so this should create now better, more useful plots all around.

Examples:
comparison_slide
average_speedup
mesh_scaling
generate_scaling

### DIFF
--- a/src/tests/blueprint/plot_benchmark_output.py
+++ b/src/tests/blueprint/plot_benchmark_output.py
@@ -49,9 +49,8 @@
 #         each containing:
 #           - mesh_scaling.png, mesh_scaling_per_element.png
 #           - generate_scaling.png, generate_scaling_per_element.png
-#           - generate_scaling_by_config.png and
-#             generate_scaling_per_element_by_config.png (only when more
-#             than one src/exec/out/sync configuration is present)
+#             (one panel per benchmark, one line per src/exec/out/sync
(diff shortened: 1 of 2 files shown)


JustinPrivitera (approved):
Is the intention for the version on develop to be the definitive version and the version on the device support branch to be discarded? We should consider unifying them in that case.

PR #1696, opened 2026-09-09
Extend the benchmark to work with MPI

siramok:
Based on #1695 for simplicity (lets review and merge that one first).

This PR extends the benchmark to support an arbitrary number of MPI ranks and an adjustable number of mesh domains per rank (develop branch only, device support version will have to come in another PR). It does so by building a pencil mesh (as seen elsewhere). It assumes that Caliper was also built with MPI, which conveniently handles the complexities of merging all of the timing data into exactly one .cali file.

There's also some basic support for plotting, although it treats the entire MPI run as one resulting in one set of timings (it doesn't do plot the individual rank timings). This looks at the slowest rank (max) by default, but there's an option called `--rank-reduce` to plot by the fastest rank (min) or to take the average timings of all ranks (mean).


src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
```diff
@@ -222,7 +292,13 @@ main(int argc, char *argv[])
 {
     ::testing::InitGoogleTest(&argc, argv);
 
-    if (!annotations::supported())
+#if defined(CONDUIT_RELAY_MPI_ENABLED)
```
JustinPrivitera: I don't think we want to use the relay MPI macro... is there one for Blueprint that we can use?

JustinPrivitera (changes requested):
One question

src/tests/blueprint/t_blueprint_mesh_transform_benchmark.cpp
siramok: Oddly enough, I didn't find any preexisting MPI macros that are available to the whole project. I guess Conduit mostly has files that entirely use or don't use MPI, and those files are conditionally built depending on whether MPI is available.

We have various options, it's just a matter of what you think:
1. We could define a new global macro like `CONDUIT_MPI_ENABLED` (surely something in the future will care about whether MPI is enabled?).
2. We could define a bespoke macro just for this file, like `CONDUIT_BENCHMARK_MPI_ENABLED`.
3. Depending on how it's defined (I would need to go look at it again), maybe `CONDUIT_RELAY_MPI_ENABLED` could just be renamed to `CONDUIT_MPI_ENABLED` as long as it doesn't imply anything about relay specifically.
4. We could duplicate the benchmark file and just have one be serial and the other be MPI.
5. Surely there's some other solution that I haven't thought of.

siramok: I went with option # 2. It added only like 1 or 2 new lines, so I think it's okay.

JustinPrivitera: ok, sounds good.

JustinPrivitera (approved):

PR #1700, opened 2026-09-19
Preserve caller CMake policy state when importing Conduit

### LINKED ISSUE #1697: Add "known version" to `cmake_minimum_required` in `ConduitConfig.cmake`
`ConduitConfig.cmake` has a `cmake_minimum_required(3.8)` in it. With CMake 3.31+, this warns. The solution is to add a `...<max-known>` part to the version. This declares that the file is *aware of* (it is *not* a "maximum works with" version) and works with `NEW` policies up to this version. I don't see any testing of minimum CMake versions here, so I can't really say what version it should be or whether the minimum should be bumped.

Also, it is better as this pattern to avoid affecting code outside of the file:

```
cmake_policy(PUSH)
cmake_policy(VERSION 3.8...<max-known>)

cmake_policy(POP)
```

sylvesterkaczmarek:
Fixes #1697.

Use a scoped `cmake_policy(VERSION 3.8...3.26)` in the exported package configuration. The policy version matches the project's existing CMake 3.26 build minimum, while the package's 3.8 consumption floor is unchanged. This removes the compatibility warning and avoids changing the consumer's `CMAKE_MINIMUM_REQUIRED_VERSION` or policy settings, including with `NO_POLICY_SCOPE`.

Add a deterministic CMake regression for first and repeated imports in both policy-scope modes. Its dependency files are stubbed to isolate configuration behavior; the installed-package example was also built and run separately against a real Conduit installation.

Validation on macOS arm64:
- Reproduced the original installed-package configuration failure with `-Werror=deprecated`, and the caller-minimum mutation in the new regression.
- Built and installed Conduit; all 55 tests selected by `ctest -R '^t_conduit_'` passed.
- The new regression and installed C++ example configure/build/run passed with CMake 3.26.4, 3.31.10, and 4.3.2, treating deprecation warnings as errors.
- `git diff --check` passed.

MPI, GPU, Python, Fortran, and optional external-library configurations were not exercised locally.


cyrush:
@sylvesterkaczmarek thanks for your suggested fix, but we are going to use a simpler approach to resolve #1697.

sylvesterkaczmarek:
Thanks for letting me know. Sounds good, and happy to help again if another CMake/package-config edge case comes up.

